PR #64, opened 2017-05-04
[FIX] Refactor CartLineItem to be generic

mikkoh:
This PR refactors cart line items to be a new generic type called `CartLineItem` from `CheckoutLineItemInput`.

At one time for checkouts `CheckoutLineItemInput` was the only input type for creating line items but now that we have `CheckoutLineItemUpdateInput` also I felt it's important to have a generic line item type type for the Cart from which we can create `CheckoutLineItemInput's` and `CheckoutLineItemUpdateInput's`.

The saved boolean will be used to determine when updates are done instead of creations.

### DIFF
--- a/Assets/Editor/TestCart.cs
+++ b/Assets/Editor/TestCart.cs
@@ -42,9 +42,10 @@ namespace Shopify.Tests
             cart.LineItems.AddOrUpdate("Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0LzczMzY1NjgxMzE=");
 
             Assert.AreEqual(3, cart.LineItems.All().Count, "has 3 items in cart");
-            Assert.AreEqual(33, cart.LineItems.Get("Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0VmFyaWFudC8yMDc1NjEyOTE1NQ==").quantity, "variant 20756129155 quantity is 33");
-            Assert.AreEqual(2, cart.LineItems.Get("Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0VmFyaWFudC8yMDc1NjEyOTM0Nw==").quantity, "variant 20756129347 quantity is 2");
-            Assert.AreEqual(1, cart.LineItems.Get("Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0LzczMzY1NjgxMzE=").quantity, "variant 7336568131 quantity is 1");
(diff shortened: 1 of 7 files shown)


scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
```diff
@@ -0,0 +1,354 @@
...
+        private string VariantIdFromSelectedOptions(Product product, Dictionary<string, string> selectedOptions) {
+            List<ProductVariant> variants = (List<ProductVariant>) product.variants();
+            string variantId = null;
+
+            foreach(ProductVariant variant in variants) {
+                List<SelectedOption> variantSelectedOptions = variant.selectedOptions();
+                variantId = variant.id();
+
+                if (variantSelectedOptions.Count != selectedOptions.Keys.Count) {
+                    variantId = null;
```
dengjeffrey: Would it be more clear if we removed this line, and used affirmative conditions? 
```C++
foreach(ProductVariant variant in variants) {
    List<SelectedOption> variantSelectedOptions = variant.selectedOptions();

    if (variantSelectedOptions.Count == selectedOptions.Keys.Count) {

        variantId = variant.id()

	foreach(SelectedOption variantOption in variantSelectedOptions) {
	    string optionName = variantOption.name();
            string optionValue = variantOption.value();

     	    if (!selectedOptions.ContainsKey(optionName) || selectedOptions[optionName] != optionValue) {
                variantId = null;
                break;
     	    }
	}

	if (variantId != null) {
		break;
	}
    }
}
```

Assets/Editor/TestCart.cs
```diff
@@ -54,6 +55,29 @@ public class TestCart {
...
+            });
+
+            CheckoutLineItemInput input = (CheckoutLineItemInput) cart.LineItems.Get(variandId);
+            CheckoutLineItemUpdateInput updateInput = (CheckoutLineItemUpdateInput) cart.LineItems.Get(variandId);
+
+            Assert.IsNotNull(input);
+            Assert.IsNotNull(updateInput);
+            Assert.IsNotNull(input.customAttributes);
+            Assert.IsNotNull(updateInput.customAttributes);
+            Assert.AreEqual("fun", input.customAttributes[0].key);
```
dengjeffrey: Can this be better if `fun` and other re-used values were in a constant?

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
mikkoh: 👌 I'll cleanup.

Assets/Editor/TestCart.cs
mikkoh: Yeah probably. But I think I'd rather cleanup when it becomes an issue.

PR #67, opened 2017-05-12
[FIX] Split up default queries

mikkoh:
This PR simply just splits up `DefaultQueries.cs` into two separate files which contain queries for products and collections.

Later we'll have default queries for checkouts.


dengjeffrey:
🚢

Issue #65, opened 2017-05-05
Add Native Payments


dengjeffrey:
## What
Allow the SDK to support Apple Pay and Android Pay as part of the checkout flow.

## How
Integrating native in app payments require the use of [Native Plugins](https://docs.unity3d.com/Manual/Plugins.html)

+ **TLDR:** 
	+ Accessing Plugin functions from Unity: 
		+ On iOS we expose functions to Unity by wrapping methods in C++, on Android we will access the JVM and query for class and method definitions.
	+ Accessing Unity functions from Plugins:
		+ `UnitySendMessage("GameObjectName1", "MethodName1", "Message to send")`
		+ The message must be a UTF8 String 

## Checklist

- [X] Serialize `PassKit` Objects going to Unity
- [x] Deserialize `PassKit` Objects coming from Unity
- [x] `PaymentSession` that manages the Authorization Controller
- [x] `ApplePayEventDispatcher` that forwards events from Authorization Controller to Unity
- [x] `MessageCenter` that handles responses from Unity
- [x]   `Cart+ApplePay` a native C++ wrapper that exposes methods to create a `PaymentSession`
- [x] Shipping mutations for checkout query on Unity
- [x] Hooks on `Cart` that allow the user to create a `PaymentSession`
- [x] Handle messages from the `ApplePayEventDispatcher` in Unity
- [x] Apple Pay
- [ ] Android Pay

dengjeffrey:
Below is my thought process and possible solution to implementing this. I have not experimented on Android thoroughly so any input will be helpful

## How
### iOS
+ On iOS I have considered implementing Apple Pay in Objective C, but since Objective C is not likely to be maintained in the future and we do not lose any functionality by implementing it in Swift, I went with the latter.
+ By using a bridging header we will expose Swift functions to an ObjC++ wrapper.
+ Using a ObjC++ wrapper we expose the functions to Unity scripts.

### Android
+ There seems to be few number of ways to implement Android plugins each with there own caveats. TBD 

## Proposed Native Payment Flow

1. On a checkout screen we will make a call `Cart.canCheckoutWithNativePay() -> Bool` which determines if Apple Pay or Android Pay is available on the device. Based on the result the we will show a Native Pay button.
[...]

PR #68, opened 2017-05-16
[Feature] apple pay serializable

dengjeffrey:
### What
- Adds JSON serialization to PassKit objects 
    + Only serializes properties that will be needed by the Storefront API, properties such as address2 in PKContact are omitted 
- Adds Mock PassKit objects from [Mobile-Buy-SDK](https://github.com/Shopify/mobile-buy-sdk-ios/tree/sdk-3.0/master/PayTests) for testing. 
    + Since PKPayment is not a mutable object
- Adds Tests for serialization

### Extra info
+ Part of #65
    - Serialization will be used to aid the passing of data from Apple Pay delegate calls to Unity
+ If you would like to run the tests for yourself, and have issues with linking and building the project, let me know, I have most likely encountered all of them.

### Review
@mikkoh @dbart01


Assets/Plugins/iOS/Shopify/PKContact+BuySerializable.swift
```diff
@@ -0,0 +1,62 @@
...
+    case zip
+    case email
+}
+
+extension PKContact: BuySerializable {
+    public func asDictionary() -> Dictionary<String, Any> {
+        var dictionary = [String: Any]()
+        add(.firstName, withValue: self.name?.givenName,          to: &dictionary)
+        add(.lastName, withValue: self.name?.familyName,          to: &dictionary)
+        add(.address1, withValue: self.postalAddress?.street,     to: &dictionary)
```
mikkoh: Will `self.postalAddress?.street` have apartment number?

If it does bake in the apartment number is Shopify able to handle having `address1` and `address2` bakes into `address1`?

Assets/Plugins/iOS/Shopify/BuyTests/BuySerializableTests.swift
```diff
@@ -0,0 +1,160 @@
...
+import ProductName
+
+class BuySerializableTests: XCTestCase {
+    
+    let emailAddress = "test_email@shopify.com"
+    let firstName    = "first name"
+    let lastName     = "last name"
+    let phone        = "123-456-7890"
+    let address1     = "80 Spadina Ave"
+    let address2     = "420 Wellington"
```
mikkoh: 🔥 `address2` it's never used.

Assets/Plugins/iOS/Shopify/BuyTests/BuySerializableTests.swift
```diff
@@ -0,0 +1,160 @@
...
+    let city         = "Toronto"
+    let company      = "Shopify"
```
mikkoh: `company` is never used. Might not be a bad thing to use it

Assets/Plugins/iOS/Shopify/BuyTests/BuySerializableTests.swift
```diff
@@ -0,0 +1,160 @@
...
+import PassKit
+import ProductName
+
+class BuySerializableTests: XCTestCase {
+    
+    let emailAddress = "test_email@shopify.com"
+    let firstName    = "first name"
+    let lastName     = "last name"
+    let phone        = "123-456-7890"
+    let address1     = "80 Spadina Ave"
```
mikkoh: Rename `address1` to be `address` otherwise it could be confused with Shopify's `address1` and `address2` field naming.

Assets/Plugins/iOS/Shopify/BuyTests/BuySerializableTests.swift
```diff
@@ -0,0 +1,160 @@
...
+        let shippingContactDict = paymentDict[PKPaymentSerializedField.shippingContact.rawValue]    as! [String: Any]
+        let tokenDict           = paymentDict[PKPaymentSerializedField.tokenData.rawValue]          as! [String: Any]
+        let shippingIdentifier  = paymentDict[PKPaymentSerializedField.shippingIdentifier.rawValue] as! String
+        
+        let tokenDataDict              = tokenDict[PKPaymentTokenSerializedField.paymentData.rawValue] as Any
+        let tokenData                  = try! JSONSerialization.data(withJSONObject: tokenDataDict)
+        let tokenTransactionIdentifier = tokenDict[PKPaymentTokenSerializedField.transactionIdentifier.rawValue] as! String
+        
+        assertEqualContact(billingContactDict,  contact: billingContact)
+        assertEqualContact(shippingContactDict, contact: shippingContact)
```
mikkoh: Since `shippingContact` is also the `billingContact` in this case do we need to assert `shippingContactDict` and `shippingContact` are equal here since we're essentially doing the same thing in:
`assertEqualContact(billingContactDict,  contact: billingContact)`

Assets/Plugins/iOS/Shopify/BuyTests/BuySerializableTests.swift
```diff
@@ -0,0 +1,160 @@
...
+        
+        XCTAssertEqual(shippingIdentifier,         shippingMethod.identifier)
+        XCTAssertEqual(tokenData,                  token.paymentData)
+        XCTAssertEqual(tokenTransactionIdentifier, token.transactionIdentifier)
+        
+        XCTAssertNotNil(payment.asJSONString())
+        XCTAssertNoThrow(try JSONSerialization.jsonObject(with: payment.asJSONString().data(using: .utf8)!))
```
mikkoh: Currently there are no tests to check that `asJSONString` returns correct values. Could you create a second test that ensures that `asJSONString` is working correctly.

Ideally in all tests we would check that the deserialized json string's values are correct but this might be overkill.

Assets/Plugins/iOS/Shopify/BuyTests/BuySerializableTests.swift
dengjeffrey: I decided to remove it, since no Pass Kit objects return any details on the user's company

Assets/Plugins/iOS/Shopify/PKContact+BuySerializable.swift
dengjeffrey: `self.postalAddress?.street` will have apartment number. Thanks for raising an issue with the `address1` and `address2` seems like `postalAddress.street` is a concatenation of both addresses separated by `\n`

Assets/Plugins/iOS/Shopify/BuyTests/BuySerializableTests.swift
dengjeffrey: I believe it would be better for completeness, since billingContact and the shippingContact are separately serialized.

mikkoh: Oh yes you're right. I walked through the code again and I see it now.

Assets/Plugins/iOS/Shopify/BuyTests/BuySerializableTests.swift
mikkoh: So just for documentations sake I wanted to document our IRL convo.

Since Apple encodes `Address1` and `Address2` into `street` we're gonna attempt to parse `Address1` and `Address2`from `street`

Assets/Plugins/iOS/Shopify/BuyTests/BuySerializableTests.swift
mikkoh: 👍

Assets/Plugins/iOS/Shopify/BuyTests/BuySerializableTests.swift
```diff
@@ -0,0 +1,207 @@
...
+        
+        if let firstAddress = contactDict[PKContactSerializedField.address1.rawValue] as? String {
+            if let secondAddress = contactDict[PKContactSerializedField.address2.rawValue] as? String {
+                XCTAssertEqual(firstAddress + "\n" + secondAddress, contact.postalAddress?.street);
+            }
+            else
+            {
+                XCTAssertEqual(firstAddress, contact.postalAddress?.street);
+            }
+        }
```
mikkoh: Please make this assertion of address components to be explicit.

So if you're expecting 2 address components then assert those two components.

If you're expecting 1 address component ensure that there aren't two address components.

Assets/Plugins/iOS/Shopify/BuySerializable.swift
```diff
@@ -0,0 +1,47 @@
...
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+public protocol BuySerializable {
```
dbart01: I don't think the `Buy` prefix is necessary.

Assets/Plugins/iOS/Shopify/BuySerializable.swift
```diff
@@ -0,0 +1,47 @@
...
+    func asJSONString() -> String
+    func asDictionary() -> Dictionary<String, Any>
```
dbart01: It seems more intuitive for a `Serializable` type to vend `Data` rather than a string, in my opinion. We can also make this interface more Swifty:
```swift
typealias JSON = [String: Any]

public protocol Serializable {
    func serializedJson() -> JSON
}

extension Serializable {
    
    func serializedData() throws -> Data {
        return try JSONSerialization.data(withJSONObject: self.serializedJson(), options: .prettyPrinted)
    }

    func serializedString() -> String {
        return String(data: self.serializedData(), encoding: .utf8)
    }
}
```

The two methods in the `extension` should not be declared in the protocol. They are derived from `JSON` and are implemented automatically for any types conforming that are able to provide a `JSON` representation of themselves.

Assets/Plugins/iOS/Shopify/BuySerializable.swift
```diff
@@ -0,0 +1,47 @@
...
+
+extension BuySerializable {
+    public func asJSONString() -> String {
+        let dictionary = asDictionary()
+        let jsonData = try! JSONSerialization.data(withJSONObject: dictionary, options: .prettyPrinted)
+        return String(data: jsonData, encoding: .utf8)!
+    }
+}
+
+public func insert<T: RawRepresentable>(_ key: T, withValue value: Any?, to dictionary: inout Dictionary<String, Any>)
```
dbart01: This would probably be better as an `extension` to `Dictionary`.

Assets/Plugins/iOS/Shopify/BuySerializable.swift
dengjeffrey: Yes this looks a lot better, thanks 😄

Assets/Plugins/iOS/Shopify/Dictionary+RawRepresentable.swift
```diff
@@ -0,0 +1,35 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+extension Dictionary  {
+    public mutating func insertIfNotNull<T: RawRepresentable>(_ value: Dictionary.Value?, forKey key: T) where T.RawValue == Dictionary.Key {
```
dbart01: This seems very Objective-C: `insertIfNotNull`. How about `insert(nullable value: Dictionary.Value?, for key: T)` instead?

Assets/Plugins/iOS/Shopify/PKContact+Serializable.swift
```diff
@@ -0,0 +1,77 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+import PassKit
+
+public enum PKContactSerializedField: String {
```
dbart01: We shouldn't adopt Apple framework' naming conventions here. Just `ContactField` is sufficient.

Assets/Plugins/iOS/Shopify/PKContact+Serializable.swift
```diff
@@ -0,0 +1,77 @@
...
+            
+            if addresses.count > 1 {
+                add(.address2, withValue: addresses[1], to: &json)
+            }
+        }
+        
+        return json
+    }
+    
+    private func add(_ key: PKContactSerializedField, withValue value: Any?, to json: inout JSON) {
```
dbart01: What's the point of this abstraction?

Assets/Plugins/iOS/Shopify/PKContact+Serializable.swift
```diff
@@ -0,0 +1,77 @@
...
+extension PKContact: Serializable {
+    func serializedJSON() -> JSON {
+        var json = [String: Any]()
+        add(.firstName, withValue: self.name?.givenName,           to: &json)
+        add(.lastName,  withValue: self.name?.familyName,          to: &json)
+        add(.city,      withValue: self.postalAddress?.city,       to: &json)
+        add(.country,   withValue: self.postalAddress?.country,    to: &json)
+        add(.province,  withValue: self.postalAddress?.state,      to: &json)
+        add(.zip,       withValue: self.postalAddress?.postalCode, to: &json)
+        add(.email,     withValue: self.emailAddress,              to: &json)
```
dbart01: Building on my last comment, we can modify the `Dictionary` extension to have an interface similar to this:
```swift
json.insert(nullable: self.name?.givenName, for: ContactField.firstName)
json.insert(nullable: self.name?.familyName, for: ContactField.lastName)
json.insert(nullable: self.postalAddress?.city, for: ContactField.city)
...
```

Assets/Plugins/iOS/Shopify/Serializable.swift
```diff
@@ -0,0 +1,45 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+
+protocol Serializable {
+    typealias JSON = [String: Any]
```
dbart01: Is there any particular reason `JSON` is nested inside `Serializable`?

Assets/Plugins/iOS/Shopify/PKContact+Serializable.swift
```diff
@@ -0,0 +1,77 @@
...
+        }
+        
+        return json
+    }
+    
+    private func add(_ key: PKContactSerializedField, withValue value: Any?, to json: inout JSON) {
+        json.insertIfNotNull(value, forKey: key)
+    }
+    
+    private func addressComponents(inStreet street: String) -> [String] {
```
dbart01: I don't see a lot of benefit in this abstraction. It would almost be worthwhile as an extension on `String`:
```swift
extension String {
    var lines: [String] {
        return self.components(separatedBy: "\n")
    }
}
```
which would make the call site more expressive:
```swift
if let street = self.postalAddress?.street,
    let lines = street.lines {
    // ...
}

File: Assets/Plugins/iOS/Shopify/Serializable.swift (hunk 15 above)
Comment (dengjeffrey) [inline Assets/Plugins/iOS/Shopify/Serializable.swift]: Nope, nice catch

File: Assets/Plugins/iOS/Shopify/PKContact+Serializable.swift (hunk 13 above)
Comment (dengjeffrey) [inline Assets/Plugins/iOS/Shopify/PKContact+Serializable.swift]: It was to give context so that we could just use dot notation instead of explicitly writing the Enum name. Now that the enum is shorter, the suggestion below is just as clear, thanks!

**Comment by dengjeffrey** [comment]
Thanks for all the feedback 😄 , please take another look when possible

File: Assets/Plugins/iOS/Shopify/Dictionary+RawRepresentable.swift [hunk 17]
```diff
@@ -0,0 +1,36 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+extension Dictionary  {
+    
+    public mutating func insert<T: RawRepresentable>(nullable value: Dictionary.Value?, forKey key: T) where T.RawValue == Dictionary.Key {
```
dbart01: I think the Swift naming conventions prefer `for key: T` instead of `forKey key: T`.

dbart01: I realize now that they use `forKey` in the `Dictionary` interface... but I still prefer the former approach 😝

mikkoh:
🚢

Assets/Plugins/iOS/Shopify/Dictionary+RawRepresentable.swift
dengjeffrey: Haha I think it's because Dictionary.Key is considered weakly typed, hence the `forKey`. In the autofill all you would see is `updateValue(String, for: String)`

PR #70, opened 2017-05-17
[Fix] Test default query add line items

mikkoh:
It seems that the attribute order has changed in the schema. This broke one of the tests.


mikkoh:
:+1: this could be interesting.

PR #71, opened 2017-05-17
[Feature] Split line items into new, modified, and deleted

mikkoh:
This PR creates 3 lists of line items on `CartLineItems`:

* New line items
* Modified line items
* Deleted line items

This will allow us build queries to create, modify, delete line items in the Storefront API.

### DIFF
--- a/Assets/Editor/TestCart.cs
+++ b/Assets/Editor/TestCart.cs
@@ -369,6 +369,8 @@ namespace Shopify.Tests
 
             cart.LineItems.AddOrUpdate(variantId1, 33);
             cart.LineItems.AddOrUpdate(variantId2, 2);
+
+            var lineItem1 = cart.LineItems.Get(variantId1);
         
             cart.LineItems.ClearNew();
 
@@ -389,6 +391,7 @@ namespace Shopify.Tests
 
             Assert.AreEqual(0, cart.LineItems.Deleted().Count, "Has 0 deleted line item after clear");
             Assert.AreEqual(1, cart.LineItems.All().Count, "Has 1 line item at end");
+            Assert.AreEqual(lineItem1, cart.LineItems.All()[0], "lineItem1 is the last remaining line item");
         } 
     }
 }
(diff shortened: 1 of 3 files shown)


Assets/Editor/TestCart.cs
```diff
@@ -272,5 +272,81 @@ public class TestCart {
...
+            cart.LineItems.Delete(variantId3);
+
+            Assert.AreEqual(0, cart.LineItems.New().Count, "No new line items");
+            Assert.AreEqual(0, cart.LineItems.Modified().Count, "No modified line items");
+            Assert.AreEqual(1, cart.LineItems.Deleted().Count, "Has one deleted line item");
+
+            cart.LineItems.ClearDeleted();
+
+            Assert.AreEqual(0, cart.LineItems.Deleted().Count, "Has 0 deleted line item after clear");
+            Assert.AreEqual(1, cart.LineItems.All().Count, "Has 1 line item at end");
```
dengjeffrey: Should we also be testing if the remaining items are what is suppose to be in the cart?

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
```diff
@@ -127,28 +130,38 @@ namespace <%= namespace %>.SDK {
...
                 if (quantity != null) {
-                    input.Quantity = (long) quantity;
+                    modifiedLineItem.Quantity = (long) quantity;
+                    didModify = true;
                 }
 
                 if (customAttributes != null) {
-                    input.CustomAttributes = customAttributes;
+                    modifiedLineItem.CustomAttributes = customAttributes;
+                    didModify = false;
```
dengjeffrey: Could you quickly explain what custom attributes are?  Are line items not considered modified if we update custom attributes?

mikkoh: Sorry do you mean directly modify the `CustomAttributes` or overwrite like this case?

Assets/Editor/TestCart.cs
mikkoh: This may already be tested in another test but I'll double check.

mikkoh: I ended up adding in this test. Better safe than sorry.

mikkoh:
@dengjeffrey I've fixed the above issues. I ended up making watching line item changes more robust where the line item itself notifies `CartLineItems` that it has been changed.

PR #74, opened 2017-05-19
[FIX] CI compile parsing error

dengjeffrey:
+ Removes incorrect whitespace that caused older versions of Mono to have parsing issues
+ Adds ./ to correct test path errors
+ Adds Print statement to output the version of Unity
+ Removes -quit flag from build, -runEditorTests will quit upon failure or success


mikkoh:
I'd merge this one into master.

PR #75, opened 2017-05-19
[Feature] Global GameObject

mikkoh:
This PR allows you to add components to a global `GameObject`. Previously the `UnityLoader` had it's own `GameObject`. Now for checkouts we need to add the ability to poll on an interval we we need to have `GameObject` to add a script to. Instead of all these scripts creating their own `GameObjects` it's probably better we just have one global game object.

### DIFF
--- a/scripts/generator/graphql_generator/csharp/SDK/GlobalGameObject.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/GlobalGameObject.cs.erb
@@ -1,0 +1,17 @@
+#if !SHOPIFY_MONO_UNIT_TEST
+namespace <%= namespace %>.SDK {
+    using UnityEngine;
+
+    public class GlobalGameObject {
+        private static GameObject GameObject;
+
+        public static T AddComponent<T>() where T: Component {
+            if (GameObject == null) {
+                GameObject = new GameObject("Shopify");
+            }
+
+            return GameObject.AddComponent<T>();
+        }
+    }
+}
+#endif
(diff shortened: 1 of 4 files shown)


scripts/generator/graphql_generator/csharp/SDK/GlobalGameObject.cs.erb
```diff
@@ -0,0 +1,17 @@
...
+
+    public class GlobalGameObject {
+        private static GameObject GameObject;
+
+        public static T AddComponent<T>() where T: Component {
+            if (GameObject == null) {
+                GameObject = new GameObject("Shopify");
+            }
+
+            return GameObject.AddComponent<T>();
```
dengjeffrey: Should we be checking if the component doesn't already exist before adding?

mikkoh: I feel like `GlobalGameObject` should not. You should be able to add multiple components at the same time.

For `UnityLoader` for instance lets say you're doing multiple loads at the same time you need to add multiple components. Although I guess we could refactor `UnityLoader` to not destroy itself but you can have multiple loads.

For `UnitySetTimeout` in #78 we check if a component already exists and if it does not just simply use that existing component.

I feel like for this PR `GlobalGameObject` would not check for the existence of components but if we see it being valuable then we can add it in.

dengjeffrey: Thanks for clarifying 👍

PR #77, opened 2017-05-19
[FIX] Created a `unity-buy-sdk` test store

mikkoh:
We have an issue right now where it's hard for us to tell in Splunk whether queries are coming from people testing the SDK or if it's from Integration tests. This PR updates the store url and access token to use a test shop `unity-buy-sdk.myshopify.com`.

This update will allow us to differentiate better in Splunk whether developers are testing the sdk based on the SDK documentation or if it's queries are coming from internal development.


mikkoh:
@dengjeffrey I've removed all example output comments.

PR #80, opened 2017-05-24
[Feature] Xcode project config

dengjeffrey:
+ Adds a Post Process Build listener that configures the .xcodeproj file for iOS 
    + Adds Shopify Test files to the Test Target and removes them from the Unity-iPhone Target 
    + Sets the Framework Runpath Search Keys
    + Changes appends Test to the Product Name for the Test Target
    + Enable testability on the Debug configuration for the Unity Target
    + Sets the minimum iOS version of the project file to iOS 9.0
    + Removes the default Unity_iPhone_Tests from the test target
         + I removed this because it was a tests that waits for Unity to exit.
         + Whilst building from the Test Target we do not have access to the methods that quits Unity properly, so it was not worth having this test case in the target.
        + Since this test case is generated it would be better if we used our own test case rather than modifying it. I'm leaving this Test case for a future PR.


scripts/generator/graphql_generator/csharp/Editor/ShopifyPBXProject.cs.erb
```diff
@@ -0,0 +1,82 @@
...
+                    string fileGuid = FindFileGuidByProjectPath(localFilePath);
+                    AddFileToBuild(testTargetGuid, fileGuid);
+                    RemoveFileFromBuild(unityTargetGuid, fileGuid);
+                }
+
+                foreach (var otherDirectory in directory.GetDirectories()) {
+                    SetFilesInDirectoryToTestTarget(otherDirectory);
+                }
+            }
+            catch(Exception) {
```
mikkoh: nitpick but cuddle this so `} catch(Exception) {`

scripts/generator/graphql_generator/csharp/Editor/iOSPostProcessor.cs.erb
```diff
@@ -0,0 +1,37 @@
...
+                Debug.Log(e.Message);
+            }
+
+            // Removes default Unity_iPhone_Tests from the Test Target
+            project.RemoveFileFromBuild(project.testTargetGuid, project.FindFileGuidByProjectPath("Unity-iPhone Tests/Unity_iPhone_Tests.m"));
+            project.SetBuildProperty(project.GetAllTargetGuids(), ShopifyPBXProject.SwiftVersionKey, "3.0");
+            project.SetBuildProperty(project.unityTargetGuid, ShopifyPBXProject.RunpathSearchKey, "@executable_path/Frameworks");
+            project.SetBuildProperty(project.testTargetGuid, ShopifyPBXProject.RunpathSearchKey, "@loader_path/Frameworks");
+            project.SetBuildProperty(project.testTargetGuid, ShopifyPBXProject.ProjectModuleNameKey, "$(PRODUCT_NAME:c99extidentifier)Tests");
+            project.SetBuildPropertyForConfig(project.GetDebugConfig(project.unityTargetGuid), ShopifyPBXProject.EnableTestabilityKey, "YES");
```
mikkoh: Should setting all these properties happen in `ShopifyPBXProject`. What's the advantage of doing this here?

dengjeffrey: I gave this some thought, I made the mistake of naming it `ShopifyPBXProject`. It should really be named something like `ExtendedPBXProject` as it only provides conveniences over PBXProject. The `iOSPostProcess` should be `ShopifyIosPostProcess` as it is actually doing Shopify related mutations. My motivation for this was to have loosely coupled components which allows us and others to modify or subclass as needed in the future.

scripts/generator/graphql_generator/csharp/Editor/PostProcessor.cs.erb
```diff
@@ -0,0 +1,15 @@
+namespace <%= namespace %>.SDK {
+    using UnityEngine;
+    using UnityEditor;
+    using UnityEditor.Callbacks;
+
+    public class PostProcesser {
+
+        [PostProcessBuildAttribute(0)]
```
dengjeffrey: Since the our process happens first, the settings can be overwritten by their custom processors. The only case where it would conflict is their process changes the iOS Target Version  below 9.0 after our runs. For developers who have experience with native plugins, they will be able to determine if they need this post process.

circle.yml
```diff
@@ -3,9 +3,9 @@ machine:
     PROJECT_ROOT: $PWD/unity-buy-sdk
     UNITY_DOWNLOAD_DIR: $PROJECT_ROOT/unity
     UNITY_PKG_LOCATION: $UNITY_DOWNLOAD_DIR/Unity.pkg
-    UNITY_PKG_URL: http://netstorage.unity3d.com/unity/fdbb5133b820/MacEditorInstaller/Unity-5.3.4f1.pkg
+    UNITY_PKG_URL: http://netstorage.unity3d.com/unity/29055738eb78/MacEditorInstaller/Unity-5.3.6f1.pkg
```
mikkoh: Just for documentation purposes. Why are we bumping the version that CI is running?

dengjeffrey: I've been testing it out, I haven't found a version that works just yet. The backwards compatibility for iOS specific functions are a bit spotty between 5.3-5.6

dengjeffrey:
Will close this for the time being, as there is an issue with installing the iOS Editor Support for Unity on CI

PR #81, opened 2017-05-24
[Fix] Split up MockLoader

mikkoh:
`MockLoader` currently is a massive class. This PR splits up `MockLoader` to a few separate classes that handle individual queries. I've also added a test to ensure that two classes that return responses for queries do not both handle the same query. I think further refactoring could be done but at least this is the right step forward.

I'm merging this into `master` since it's something that should go in. After I'll rebase `beta` from `master`.

Next up I'll be adding a class that implements checkout functionality to `MockLoader`.


Assets/Editor/MockLoaderErrors.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public List<string> GetQueries() {
+            return new List<string>();
+        }
+
+        private bool Is404Query(string query) {
+            return query.Contains(@"after:""404""");
+        }
+
+        private bool IsGraphQLErroQuery(string query) {
```
dengjeffrey: Should this be `IsGraphQLErrorQuery`?

mikkoh: Good 👁

mikkoh:
@dengjeffrey the spelling mistake is fixed

PR #82, opened 2017-05-25
[Feature] iOS payment session

dengjeffrey:
### What 

I want to get some early feedback on this implementation while I write the tests. @mikkoh @dbart01 

+ Fixes PKPaymentToken+Serializable to handle cases where token data is nil. Testing Apple Pay on the simulator returns nil token data.
+ Adds a `PaymentSession` implementation
    + `PaymentSession` is responsible for creating the `PKPaymentRequest` and coordinating the mutations on that request such as adding summary items and shipping methods. 
   + `PaymentSession` is also responsible for displaying the `PKAuthorizationController`  
   + On `PKPaymentAuthorizationViewControllerDelegate` notifications the `PaymentSession` will notify it's own `PaymentSessionDelegate` and block till the `PaymentSession` is notified to unblock using `didAuthorize` or `didFinishUpdate ` 
       + There is no `PaymentSessionDelegate` method for
`paymentAuthorizationViewController(controller:, didSelect paymentMethod:, completion:`  because I do not see a use for the info provided in `PKPaymentMethod` for completing a checkout
       + We use a semaphore here so that we can block the before calling the completion closure on `PKPaymentAuthorizationDelegate`. Blocking before calling the completion, allows us to notify a managed function to update the session's summary items and shipping. Managed functions are asynchronous and have no return values so we must wait till we receive another function notifying that all mutations are finished. That is when we unlock the semaphore, and call the completion closure with the updated summary items.
       + `PaymentSession` will block an asynchronous queue that references the completion closure. The asynchronous queue is blocked until someone calls `didAuthorize` or `didFinishUpdate`. This does not block the main queue, so the user can still interact with the App. Other changes to the shipping address and billing address made by the user during this are put onto a queue by PKPaymentAuthorizationController, in effect other delegate methods are not called until the original completion closure is finished. With the exception of paymentAuthorizationViewControllerDidFinish
   + Step towards #65

### DIFF
--- a/Assets/Plugins/iOS/Shopify/PaymentSession.swift
+++ b/Assets/Plugins/iOS/Shopify/PaymentSession.swift
@@ -58,7 +58,7 @@ import PassKit
     
     /// The last status of the authorization received by passing the token data
     /// to the payment server
-    var lastAuthStatus: PKPaymentAuthorizationStatus!
+    var lastAuthStatus: PKPaymentAuthorizationStatus?
     
     var summaryItems: [PKPaymentSummaryItem]
     var shippingMethods: [PKShippingMethod]
@@ -215,7 +215,7 @@ extension PaymentSession: PKPaymentAuthorizationViewControllerDelegate {
         
         DispatchQueue.global(qos: .userInitiated).async {
             self.updateSemaphore.wait()
-            completion(self.lastAuthStatus);
+            completion(self.lastAuthStatus!);
         }
     }
 }
--- a/Assets/Plugins/iOS/Shopify/PaymentSession.swift
+++ b/Assets/Plugins/iOS/Shopify/PaymentSession.swift
@@ -1,0 +1,221 @@
+//
+//  PaymentSession.swift
+//  UnityBuySDK
+//
+//  Created by Shopify.
+//  Copyright © 2017 Shopify Inc. All rights reserved.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import UIKit
+import PassKit
+
+@objc enum PaymentStatus: Int {
+    case Failed
+    case Cancelled
+    
+    func toString() -> String {
+        switch self {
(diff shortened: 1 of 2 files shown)


Assets/Plugins/iOS/Shopify/PaymentSession.swift
```diff
@@ -0,0 +1,221 @@
...
+            return "Cancelled"
+        case .Failed:
+            return "Failed"
+        }
+    }
+}
+
+@objc protocol PaymentSessionDelegate : class {
+    func paymentSessionDidFinish(session: PaymentSession,with status: PaymentStatus)
+    func paymentSession(_ session: PaymentSession, didSelect shippingMethod: PKShippingMethod)
```
dbart01: Should likely follow the same naming principle as `didSelectShippingContact`. So `didSelectShippingMethod`.

Assets/Plugins/iOS/Shopify/PaymentSession.swift
```diff
@@ -0,0 +1,221 @@
...
+    func paymentSession(_ session: PaymentSession, didSelectShippingContact shippingContact: PKContact)
+    func paymentSession(_ session: PaymentSession, didAuthorize payment: PKPayment)
+}
+
+@objc class PaymentSession: NSObject {
+
+    /// Semaphore that blocks the PKAuthorizationController till summary items
+    /// or shipping methods are updated based on the user's input
+    let updateSemaphore = DispatchSemaphore(value: 0)
```
dbart01: Why is the semaphore necessary?

Assets/Plugins/iOS/Shopify/PaymentSession.swift
```diff
@@ -0,0 +1,221 @@
...
+    
+    /// Presents the PKAuthorizationController with the current shipping methods
+    /// and summary items in this session.
+    /// UnityAppController is set to remain active until the authorization controller
+    /// is dismissed, to enable communication between managed and unmanaged functions
+    func presentAuthorizationController()  {
+        populateRequest();
+        let authViewController = PKPaymentAuthorizationViewController.init(paymentRequest: request)
+        let unityController    = UIApplication.shared.delegate as! UnityBuyAppController
+        let topController      = unityController.topMostController()!
```
dbart01: Alternatively, you can present this on the window's root view controller:
```swift
unityController.window.rootViewController
```
The decision is up to you.

Assets/Plugins/iOS/Shopify/PaymentSession.swift
dengjeffrey: I forgot to mention in the description. But we use a semaphore here so that we can block the before calling the completion closure on `PKPaymentAuthorizationDelegate`. Blocking before calling the completion, allows us to notify a managed function to update the session's summary items and shipping. Managed functions are asynchronous and have no return values so we must wait till we receive another function notifying that all mutations are finished. That is when we unlock the semaphore, and calling the completion closure with the updated summary items.

dbart01: So `PaymentSession` will block the whole queue until someone calls `authorizePayment` on the session?

dengjeffrey: Almost. `PaymentSession` will block an asynchronous queue that will own the completion closure. The asynchronous queue is blocked until someone calls `didAuthorize` or `didFinishUpdate`. This doesn't block the  main queue, the user can still interact with the App. Other changes to the shipping address and billing address made by the user during this are put onto a queue by PKPaymentAuthorizationController, so other delegate methods are not called until the original completion closure is finished. With the exception of `paymentAuthorizationViewControllerDidFinish`

dbart01: I'm a little uneasy with this approach. I'm not sure I understand what the semaphore is trying to solve. What it does is clear but the problem it's trying to solve is unclear to me.

dengjeffrey: As discussed with @dbart01, the cleaner approach would be abstracting the semaphore locking into a `UnityMessage`. I will be refactoring for this change.

Assets/Plugins/iOS/Shopify/PaymentSession.swift
dengjeffrey: Here I'am following the delegate naming as they have in `PKPaymentAuthorizationViewControllerDelegate`. I believe it is `didSelectShippingContact` because the parameter is a `PKContact` which is ambiguous to the implementor which contact they mean, shipping or billing contact. On the other hand `didSelect shippingMethod` removes the shippingMethod from the prefix because it doesn't add any additional info to the parameter `PKShippingMethod`

PR #84, opened 2017-05-29
[Feature] iOS Payment Session

dengjeffrey:
+ Fixes PKPaymentToken+Serializable to handle cases where token data is nil. Testing Apple Pay on the simulator returns nil token data.
+ Adds `UnityBuyAppController`
    + A subclass of  `UnityAppController` that overrides the delegate method that is called when the app is about to resign active. This way we can control whether the app resigns active when the `PKPaymentAuthorizationViewController` is shown. By setting `didResignActive` to false Unity will continue to process messages sent via UnitySendMessage.
+ Adds a `PaymentSession` implementation
    + `PaymentSession` is responsible for creating the `PKPaymentRequest` and displaying a `PKPaymentAuthorizationViewController` on `iOS 9` or `PKPaymentAuthorizationController` on `iOS 10+`
    + I used the protocol `PaymentAuthorizationControlling` to generalize these two classes, so that PaymentSession can be modified to use either
   + On `PKPaymentAuthorizationViewControllerDelegate` notifications the `PaymentSession` will notify it's own `PaymentSessionDelegate` requesting for updated summary items and shipping info
        + There is no `PaymentSessionDelegate` method for
`paymentAuthorizationViewController(controller:, didSelect paymentMethod:, completion:`  because I do not see a use for the info provided in `PKPaymentMethod` for completing a checkout
+ What is different from the closed PR #82 ?
    + I removed the semaphore locks in this class. Instead as discussed the functionality for locking and retrieving updated info will be handled in another object. This keeps objects loosely coupled, and reduces the points of interaction between managed and unmanaged code.

+ Adds Tests and Mocks for PaymentSession
    + We use a subclass of `PKPaymentAuthorizationController`  as a mock to test the functionality for `PKPaymentAuthorizationControllerDelegate` and `PKPaymentAuthorizationViewControllerDelegate` methods. In `PaymentSession` we handle both delegate callbacks with a single method, thus testing with only the controller should be sufficient.

+ Step towards #65

### DIFF
--- a/Assets/Plugins/iOS/Shopify/PaymentAuthorizationControlling/PKPaymentAuthorizationViewController+PaymentAuthorizationControlling.swift
+++ b/Assets/Plugins/iOS/Shopify/PaymentAuthorizationControlling/PKPaymentAuthorizationViewController+PaymentAuthorizationControlling.swift
@@ -34,16 +34,13 @@ extension PKPaymentAuthorizationViewController: PaymentAuthorizationControlling 
             return delegate
         }
         set {
-            if let delegate = newValue as? PKPaymentAuthorizationViewControllerDelegate {
-                self.delegate = delegate
-            } else {
-                self.delegate = nil
-            }
+            self.delegate = newValue as? PKPaymentAuthorizationViewControllerDelegate
         }
     }
     
     func present(completion: ((Bool) -> Void)?) {
-        if let rootViewController = UIApplication.shared.delegate?.window??.rootViewController {
+        if let window = UIApplication.shared.delegate?.window,
+            let rootViewController = window?.rootViewController {
             rootViewController.present(self, animated: true) {
                 completion?(true)
             };
--- a/Assets/Plugins/iOS/Shopify/PaymentAuthorizationControlling/PKPaymentAuthorizationViewController+PaymentAuthorizationControlling.swift
+++ b/Assets/Plugins/iOS/Shopify/PaymentAuthorizationControlling/PKPaymentAuthorizationViewController+PaymentAuthorizationControlling.swift
@@ -1,9 +1,9 @@
 //
-//  MockToken.swift
+//  PKPaymentAuthorizationViewController+PaymentAuthorizationControlling.swift
 //  UnityBuySDK
 //
 //  Created by Shopify.
-//  Copyright (c) 2017 Shopify Inc. All rights reserved.
+//  Copyright © 2017 Shopify Inc. All rights reserved.
 //
 //  Permission is hereby granted, free of charge, to any person obtaining a copy
 //  of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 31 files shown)


Assets/Plugins/iOS/Shopify/PaymentSession.swift
```diff
@@ -0,0 +1,171 @@
...
+//
+
+import UIKit
+import PassKit
+
+@objc enum PaymentStatus: Int {
+    case Failed
+    case Cancelled
+    
+    func toString() -> String {
```
dbart01: `CustomStringConvertible` provides the interface that you're probably looking for.

dengjeffrey: Hmm sounds like a good idea, but I don't think a enum can have the `description` variable. Ideally it would be an enum of type String but it needs to be Objective C compatible

dbart01: `enum` of type `String` is nothing special. It's just an `enum` that conforms to `RawRepresentable` internally. There's nothing wrong with having an `enum` conforming to `CustomStringConvertible`. Are you saying that an `enum` can't have a `description` AND be compatible with ObjC?

dengjeffrey: To conform to `CustomStringConvertible` you need to define the `description` property, but I believe you can't define instance properties in an `enum`

dbart01: You can absolutely define instance properties in `enum`s. In fact, `toString()` is currently an instance property.

dengjeffrey: You're right 😄, good call!

Assets/Plugins/iOS/Shopify/PaymentAuthorizationControlling/PKPaymentAuthorizationViewController+PaymentAuthorizationControlling.swift
```diff
@@ -0,0 +1,64 @@
...
+import PassKit
+
+extension PKPaymentAuthorizationViewController: PaymentAuthorizationControlling {
+    
+    var authorizationDelegate: NSObjectProtocol? {
+        get {
+            return delegate
+        }
+        set {
+            if let delegate = newValue as? PKPaymentAuthorizationViewControllerDelegate {
```
dbart01: This can be simplified to just:
```swift
self.delegate = newValue as? PKPaymentAuthorizationViewControllerDelegate
```

Assets/Plugins/iOS/Shopify/PaymentAuthorizationControlling/PKPaymentAuthorizationViewController+PaymentAuthorizationControlling.swift
```diff
@@ -0,0 +1,64 @@
...
+                self.delegate = delegate
+            } else {
+                self.delegate = nil
+            }
+        }
+    }
+    
+    func present(completion: ((Bool) -> Void)?) {
+        if let rootViewController = UIApplication.shared.delegate?.window??.rootViewController {
```
dbart01: I find long optional chains hard to read. It's up to you, but I prefer to do something like this instead:
```swift
if let window = UIApplication.shared.delegate?.window,
    let rootViewController = window.rootViewController {
    ...
}
```

Assets/Plugins/iOS/Shopify/PaymentSession.swift
```diff
@@ -0,0 +1,222 @@
...
+        case .Cancelled:
+            return "Cancelled"
+        case .Failed:
+            return "Failed"
+        case .Success:
+            return "Success"
+        }
+    }
+    
+    static func from(status: PKPaymentAuthorizationStatus) -> PaymentStatus{
```
dbart01: Nit: `PaymentStatus {`

Assets/Plugins/iOS/Shopify/PaymentSession.swift
```diff
@@ -0,0 +1,222 @@
...
+    }
+}
+
+// ----------------------------------
+//  MARK: - Generic Payment Authorization Delegate Functions -
+//
+extension PaymentSession {
+    
+    func paymentAuthorizationDidFinish() {
+        let paymentStatus: PaymentStatus;
```
dbart01: `;` ... 🤔

Assets/Plugins/iOS/Shopify/PaymentSession.swift
```diff
@@ -0,0 +1,222 @@
...
+        let unityController = UIApplication.shared.delegate as! UnityBuyAppController
+        unityController.shouldResignActive = true
+        
+        if let lastAuthStatus = lastAuthStatus {
+            paymentStatus = PaymentStatus.from(status: lastAuthStatus)
+        } else {
+            paymentStatus = .Cancelled
```
dbart01: Should this logic be encapsulated in `PaymentStatus`?

dengjeffrey:
@dbart01 👍?

PR #85, opened 2017-05-30
[Feature] Web Checkouts

mikkoh:
This PR implements getting a web checkout url from the StoreFront API. It does this in the following way:
- Builds a query which
  + Creates a `Checkout` on the Storefront API using line items
  + Updates the `Checkout` on the Storefront API based on line item changes (add, delete, update)
- Then it polls to see if we need to wait for `ready` if needed
- Then a callback exposes `Checkout.webURL`

This PR currently does not include
- loading all pages of line items
- Handling user errors

The above need to be implemented still.


scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -26,9 +26,27 @@ namespace <%= namespace %> {
...
+        private bool IsCreated {
+            get {
+                return CheckoutId != null;
+            }
+        }
 
+        private const float POLL_DELAY = 0.5f;
+        private CartLineItems _LineItems;
         private ShopifyClient Client;
+        private string CheckoutId = null;
```
JakeCataford: I dont really get why the cart model knows so much about the checkout, is this an attempt to merge these two concepts into one?

mikkoh: Would you prefer that there was an abstraction for checkouts that you could pass a cart to? Yes right now we're merging the two concepts mainly to maintain the existing API which may not be a good reason to do this.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -49,52 +67,135 @@ namespace <%= namespace %> {
...
-        /// \code
-        /// // Example that creates a checkout link and opens the url in the Browser
-        /// string linkToCheckout = ShopifyBuy.Client().Cart().GetWebCheckoutLink();
-        ///
-        /// Application.OpenURL(linkToCheckout);
-        /// \endcode
-        public string GetWebCheckoutLink(string note = null) {
-            StringBuilder url = new StringBuilder();
-            bool hasLineItem = false;
+        public void GetWebCheckoutLink(Action<string> success, Action<List<string>, string> failure) {
```
JakeCataford: it these were separate concepts it could be something like: `var checkout = Checkout.Create(cart); checkout.webUrl();` which i like because it's less implicit

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -49,52 +67,135 @@ namespace <%= namespace %> {
...
+                } else {
+                    success(CheckoutWebUrl);
+                }
+            });
+        }
 
-            url.Append("http://");
-            url.Append(Client.Domain);
-            url.Append("/cart/");
+        private void CheckoutSave(Action<List<string>, string> callback) {
```
JakeCataford: same with these methods, why `Cart.CheckoutSave` vs `Cart.Checkout.Save`? We have a class for checkout already so we could put the methods on it instead?

mikkoh:
@JakeCataford if we were to extend `Checkout` and add functionality for saving the checkout etc. It's definitely clean in that the state of a checkout experience is stored in a `Checkout`.

However there's an issue. It's nice if each query on checkouts always queried the same data but where I see there being an issue is when polling. In an ideal world I think you wouldn't query line items. Especially not load all the pages of line items.

I do agree that it would be nice to separate checkout from cart I'm just not sure we can just simply extend `Checkout` and add methods on it to mutate a `Checkout`.

I'll have to think of this...

mikkoh:
@JakeCataford maybe a `Cart` just has a `Checkout` instance that is the checkout that the cart is associated with. I'm not sure we should add `Save` to `Checkout` but functionality we should maybe add is so that you can merge to checkouts to produce and updated checkout. Something like:
`Checkout newCheckout = oldCheckout + someNewCheckoutData;`. (ignore operator overloading but focus on what its doing)

mikkoh:
@JakeCataford @dengjeffrey I've updated this PR to utilize `Checkout` more efficiently to hold the state of the checkout process. Basically now each time a query is performed on `Checkout` we merge in the update into a `Checkout` instance to hold the state.

It doesn't completely separate checkouts and carts but it's heading in that direction.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -26,9 +26,25 @@ namespace <%= namespace %> {
...
+            }
+        }
 
+        private bool IsCreated {
+            get {
+                return CurrentCheckout != null;
+            }
+        }
+
+        private const float POLL_DELAY = 0.5f;
```
richardmonette: `POLL_DELAY_SECONDS`?

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -26,9 +26,25 @@ namespace <%= namespace %> {
...
+        private CartLineItems _LineItems;
```
richardmonette: Is the `_` a convention thing around `private`? If so, then should we do this on line 47 as well?

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -49,52 +65,141 @@ namespace <%= namespace %> {
...
+                if (errors != null || httpError != null) {
+                    callback(errors, httpError);
+                } else {
+                    LineItems.ClearNew();
+                    LineItems.ClearModified();
+                    LineItems.ClearDeleted();
+
+                    if (doAdd) {
+                        UpdateCheckout(response.checkoutLineItemsAdd().checkout());
+                    } else if (doModify) {
```
dengjeffrey: Will there be a case where both of these are `true`?

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -49,52 +65,141 @@ namespace <%= namespace %> {
...
-        /// \code
-        /// // Example that creates a checkout link and opens the url in the Browser
-        /// string linkToCheckout = ShopifyBuy.Client().Cart().GetWebCheckoutLink();
-        ///
-        /// Application.OpenURL(linkToCheckout);
-        /// \endcode
-        public string GetWebCheckoutLink(string note = null) {
-            StringBuilder url = new StringBuilder();
-            bool hasLineItem = false;
+        public void GetWebCheckoutLink(Action<string> success, Action<List<string>, string> failure) {
```
mikkoh: Yeah maybe for the public methods it makes sense to create a `delegate`. 👍

scripts/generator/graphql_generator/csharp/Cart.cs.erb
mikkoh: Yes there is a case like that but in both cases `Checkout` will return the same thing so there's no need to do it twice.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
mikkoh: `_` is a convention for when the variable is accessible exposed via getter and/or mutated via setter. Basically `_` denotes this variable is "kind of public".

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -49,52 +65,141 @@ namespace <%= namespace %> {
...
-            url.Append("http://");
-            url.Append(Client.Domain);
-            url.Append("/cart/");
+            DefaultQueries.checkout.Create(query, newLineItemInput);
 
-            foreach(CartLineItem lineItem in LineItems.All()) {
-                if (hasLineItem) {
-                    url.Append(",");
+            Client.Mutation(query, (Mutation response, List<string> errors, string httpError) => {
+                if (errors != null || httpError != null) {
```
mikkoh: I'd prefer to keep it this way. Mainly because if there are `errors` or `httpErrors` we want to immediately fail and call the callback. For me it's actually easier to read since it's literally just one line in the if statement.

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
```diff
@@ -400,10 +352,37 @@ namespace <%= namespace %>.SDK {
...
 
-        private void OnCartLineItemChange(CartLineItem modifiedLineItem) {
-            if (!LineItemsNew.Contains(modifiedLineItem) && !LineItemsModified.Contains(modifiedLineItem)) {
-                LineItemsModified.Add(modifiedLineItem);
+        public void UpdateLineItemsFromCheckoutLineItems(List<CheckoutLineItem> checkoutLineItems) {
+            foreach(CheckoutLineItem checkoutLineItem in checkoutLineItems) {
+                bool didUpdateID = false;
+
+                foreach(CartLineItem cartLineItem in LineItems) {    
+                    if (cartLineItem.VariantId == checkoutLineItem.variant().id()) {
```
dengjeffrey: Will there ever be a case where the size of `checkoutLineItems` = `LineItems`?

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
```diff
@@ -400,10 +352,37 @@ namespace <%= namespace %>.SDK {
...
+                        cartLineItem.ID = checkoutLineItem.id();
+                        didUpdateID = true;
```
dengjeffrey: Will `LineItems` ever have multiple items with the same `VariantId`?

mikkoh: No. As far as I know you cannot buy the same variant twice.

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
mikkoh: Sorry why do you ask this? I have an assumption why but just want to make it clear

dengjeffrey: Spoke with @mikkoh, Voiced my concerns about the runtime of the nested for loops

mikkoh:
I've removed managing state of new, and updated line items from `CartLineItem`. Now the code will use the method described here:
https://github.com/Shopify/shopify/pull/118468

Basically every time we request a weblink to ensure that the StoreFront API is in sync we remove all line items and then add all line items. This ensures that we don't need to keep track of new or updated line items. We still need to keep track of deleted line items but this is a lot more lightweight than what was previously here.

Please note in this refactor I may have missed some tests. I'd like to add those in another PR to keep this PR small(ish) but please do make note if you notice something.

scripts/generator/graphql_generator/csharp/SDK/DefaultQueriesCheckout.cs.erb
```diff
@@ -74,7 +54,7 @@ namespace <%= namespace %>.SDK {
                     ),
                 checkoutId: checkoutId,
                 lineItems: lineItems
-            );
+            );
```
dengjeffrey: Nit: Space

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
```diff
@@ -104,22 +94,45 @@ namespace <%= namespace %>.SDK {
...
+                return _IsSaved;
+            }
+        }
+
+        private bool _IsSaved = false;
         private List<CartLineItem> LineItems = new List<CartLineItem>();
-        private List<CartLineItem> LineItemsNew = new List<CartLineItem>();
-        private List<CartLineItem> LineItemsModified = new List<CartLineItem>();
-        private List<CartLineItem> LineItemsDeleted = new List<CartLineItem>();
+        private Action<string> OnDeleteLineItem;
```
JakeCataford: Some of these seem like they might benefit from events over single instance callbacks. With events, we aren't limiting the api to a single event consumer, and we could end up making it easier to interface with especially for view-style components that only need to know when to refresh their information.

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
```diff
@@ -376,10 +362,39 @@ namespace <%= namespace %>.SDK {
             return Delete(variantId);
         }
 
-        private void OnCartLineItemChange(CartLineItem modifiedLineItem) {
-            if (!LineItemsNew.Contains(modifiedLineItem) && !LineItemsModified.Contains(modifiedLineItem)) {
-                LineItemsModified.Add(modifiedLineItem);
+        public void UpdateLineItemsFromCheckoutLineItems(List<CheckoutLineItem> checkoutLineItems) {
```
JakeCataford: missing documentation on a public method, seems like most public methods have docs?

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
```diff
@@ -376,10 +362,39 @@ namespace <%= namespace %>.SDK {
...
+                        break;
+                    }
+                }
+
+                if (!didUpdateID) {
+                    long quantity = checkoutLineItem.quantity();
+                    Dictionary<string, string> customAttributes = null;
+
+                    try {
+                        List<<%= namespace %>.Attribute> attributes = checkoutLineItem.customAttributes();
```
JakeCataford: Seems weird that we have to wiggle back up a namespace here, why are these attributes not nested under the SDK namespace so we dont have to specify this whole path? Alternatively, we can just import the namespace at the top of the file...

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -46,55 +60,130 @@ namespace <%= namespace %> {
...
+                });
             }
+        }
 
-            return url.ToString();
+        private void OnDeleteLineItem(string lineItemId) {
+            DeletedLineItems.Add(lineItemId);
         }
     }
 }
```
JakeCataford: Maybe like a Purchase object that encapsulates the high level operations happening on cart and checkout? we can think about this for a refactor if you want instead of addressing it here, but since this does change the interface I would say that its more favorable to address it before shipping this pr...

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -26,9 +26,23 @@ namespace <%= namespace %> {
...
+            get {
+                return CurrentCheckout != null;
+            }
+        }
+
+        private const float POLL_DELAY_SECONDS = 0.5f;
+        private CartLineItems _LineItems;
+        private List<String> DeletedLineItems = new List<string>();
         private ShopifyClient Client;
+        private Checkout CurrentCheckout = null;
```
mikkoh: 👍  Jake do you know any good linting tools that are not tied to a specific C# dev environment? I think you're right it would be nice to do a linting pass at some point.

I've been doing `_LineItems` for cases where there's an accessors that mutates or exposes the variable. I've used `PascalCase` for class variables since that seems to be the way to go with C# projects.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -46,55 +60,130 @@ namespace <%= namespace %> {
...
-        public string GetWebCheckoutLink(string note = null) {
-            StringBuilder url = new StringBuilder();
-            bool hasLineItem = false;
+        public delegate void GetWebCheckoutLinkSuccess(string url);
+        public delegate void GetWebCheckoutLinkFailure(List<string> graphQLErrors, string httpError);
+
+        public void GetWebCheckoutLink(GetWebCheckoutLinkSuccess success, GetWebCheckoutLinkFailure failure) {
+            CheckoutSave((errors, httpError) => {
+                if (errors != null || httpError != null) {
+                    failure(errors, httpError);
```
mikkoh: We're thinking of refactoring how errors are handled. More about is here https://github.com/Shopify/unity-buy-sdk/issues/105

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -46,55 +60,130 @@ namespace <%= namespace %> {
...
+                } else {
+                    success(CurrentCheckout.webUrl());
+                }
+            });
+        }
+
+        private void CheckoutSave(Action<List<string>, string> callback) {
```
mikkoh: My thought was that we're creating A LOT of delegates all over the place and using an `Action` might not be a bad thing for `private` methods. Of course it might be good to add a comment here for what the `Action` is expecting. Thoughts?

scripts/generator/graphql_generator/csharp/Cart.cs.erb
mikkoh: Yeah I agree that I feel dirty having `DeletedLineItems` line items on the `Cart` but it also is not not convenient in `CartLineItems`. I'm ok coming back to this and working on a refactor.

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
mikkoh: I was thinking before we merge the `beta` branch on `master` to do a full review of documentation.

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
mikkoh: `System` contains a class named `Attribute`.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
JakeCataford: sounds good

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
JakeCataford: first i hear of this beta branch 😮 this changes everything lol

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
JakeCataford: 💀

scripts/generator/graphql_generator/csharp/Cart.cs.erb
JakeCataford: I use rider which uses Resharper under the hood, but i dont think it runs as a standalone thing yet.... 

I can always just do a one-off pr that just brings it into rider and runs a lint on the whole thing... we would have to make sure theres no other work going on that would conflict

JakeCataford: either way its just syntax nitpicking so i dont really care either way...

scripts/generator/graphql_generator/csharp/Cart.cs.erb
JakeCataford: hmm i prefer delegates because they are more explicit, but im ok with action if thats what you wanna do for private callbacks. Its more verbose, but i find it nicer to write when using resharper or intellisense or whatever because it gives you the context of the information, including the named variables so that its clear whats coming back in that callback and how it should be used...

BUT

Implementation is implementation, we can always go a different direction so Action is fine here... Want to agree that we should use delegate types for interfaces and we use Action/doesnt matter for private stuff? WDYT?

mikkoh:
@JakeCataford in the above comments is there something in specific that you absolutely feel should be changed? I can see using `delegates` over actions?

JakeCataford:
Nah my biggest worries were about introducing the wrong interface, but because we can refactor forward (beta branch) im not picky... Lets make sure to follow up with a few of these concerns before we merge into master though

Issue #86, opened 2017-05-31
Fix PlaybackEngines search path in Circle CI


dengjeffrey:
When compiling in CI, Mono does not load the Editor Supports in the PlaybackEngines folder located at `/Applications/Unity/PlaybackEngines/`. This causes the build to fail at compilation since we use a class ,`UnityEditor.iOS.Xcode`, defined in the iOS Editor Support.

Currently this is circumvented by creating a symbolic link from the PlaybackEngines folder to 
`/unity-buy-sdk/Assets/`

After further investigation, I believe this issue is caused by not having a license for the CI

dengjeffrey:
Fixed in #181

PR #87, opened 2017-05-31
[Feature] Xcode project config

dengjeffrey:
+ Adds a Post Process Build listener that configures the .xcodeproj file for iOS 
    + Adds Shopify Test files to the Test Target and removes them from the Unity-iPhone Target 
    + Sets the Framework Runpath Search Keys
    + Changes appends Test to the Product Name for the Test Target
    + Enable testability on the Debug configuration for the Unity Target
    + Sets the minimum iOS version of the project file to iOS 9.0
+ Creates issue #86


scripts/generator/graphql_generator/csharp/Editor/ShopifyIosPostProcessor.cs.erb
```diff
@@ -0,0 +1,43 @@
...
+
+            try {
+                project.SetFilesInDirectoryToTestTarget(testDirectory);
+            } catch (Exception e) {
+                Debug.Log(e.Message);
+            }
+
+            // Removes default Unity_iPhone_Tests from the Test Target
+            project.RemoveFileFromBuild(project.TestTargetGuid, project.FindFileGuidByProjectPath("Unity-iPhone Tests/Unity_iPhone_Tests.m"));
+            project.SetBuildProperty(project.GetAllTargetGuids(), ExtendedPBXProject.SwiftVersionKey, "3.0");
```
mikkoh: Is this needed? I don't know much about Swift versions but what I'm asking is that are we going to use features of Swift that requires the newest version.

scripts/generator/graphql_generator/csharp/Editor/ShopifyIosPostProcessor.cs.erb
```diff
@@ -0,0 +1,43 @@
...
+            var project = new ExtendedPBXProject(buildPath);
+
+            try {
+                project.SetFilesInDirectoryToTestTarget(testDirectory);
+            } catch (Exception e) {
+                Debug.Log(e.Message);
+            }
+
+            // Removes default Unity_iPhone_Tests from the Test Target
+            project.RemoveFileFromBuild(project.TestTargetGuid, project.FindFileGuidByProjectPath("Unity-iPhone Tests/Unity_iPhone_Tests.m"));
```
mikkoh: Does this effect developers who maybe relying on this existing?

scripts/generator/graphql_generator/csharp/Editor/ShopifyIosPostProcessor.cs.erb
```diff
@@ -0,0 +1,43 @@
...
+            project.SetBuildProperty(project.TestTargetGuid, ExtendedPBXProject.ProjectModuleNameKey, "$(PRODUCT_NAME:c99extidentifier)Tests");
+            project.SetBuildPropertyForConfig(project.GetDebugConfig(project.UnityTargetGuid), ExtendedPBXProject.EnableTestabilityKey, "YES");
+
+            #if UNITY_5_5_OR_NEWER
+            if (float.Parse(UnityEditor.PlayerSettings.iOS.targetOSVersionString) < 9.0) {
+                project.SetBuildProperty(project.GetAllTargetGuids(), ExtendedPBXProject.DeploymentTarget, "9.0");
+            }
+            #elif UNITY_5_4_2_OR_NEWER
+            if (UnityEditor.PlayerSettings.iOS.targetOSVersion < iOSTargetOSVersion.iOS_9_0) {
+                project.SetBuildProperty(project.GetAllTargetGuids(), ExtendedPBXProject.DeploymentTarget, "9.0");
```
mikkoh: Should this just be:
`float iosTargetVersion = UnityEditor.PlayerSettings.iOS.targetOSVersio`
and above
`float iosTargetVersion = float.Parse(UnityEditor.PlayerSettings.iOS.targetOSVersionString)`
and then in else
`float iosTargetVersion = 9.0`

just have one conditional that does:
```
if (iosTargetVersion < 9.0) {
  project.SetBuildProperty(project.GetAllTargetGuids(), ExtendedPBXProject.DeploymentTarget, "9.0");
}
```

dengjeffrey: I see what you are getting at, `UnityEditor.PlayerSettings.iOS.targetOSVersion` is a enum, so instead I will use a boolean that will denote whether we change to 9.0 or not

scripts/generator/graphql_generator/csharp/Editor/ShopifyIosPostProcessor.cs.erb
dengjeffrey: Yes, Swift 2.0 and Swift 3.0 differ drastically in terms of naming conventions and other features. With Swift 4.0 coming out in the next WWDC, it would be best that we keep the code base closest to the newest version.

scripts/generator/graphql_generator/csharp/Editor/ShopifyIosPostProcessor.cs.erb
dengjeffrey: I suppose for now we can remove having this, till we implement automated test cases for our plugin. Perhaps we can have this added specifically for the CI. It would only affect those that have modified the file directly.

mikkoh: I like that approach of having a build flag to ensure this only happens for CI. Since it's only an issue for our CI from what I've understood.

scripts/generator/graphql_generator/csharp/Editor/ShopifyIosPostProcessor.cs.erb
mikkoh: :+1: I will need to trust you on this.

PR #91, opened 2017-06-05
[Feature] iOS unity messages

dengjeffrey:
#### What/TLDR
+ Creates an architecture for sending/receiving messages to and from Unity
+ Adds the functionality for sending `PassKit` objects to Unity

#### How/Why
+ Does not include test cases, will add those in separate PRs to keep this one short
+ Decided to set `enum` of type `string` to have values in Camel Caps, since the standard in `C#` is to use `CamelCaps` for enums. Which decreases the amount of extra work we have to do when creating JSON strings.


Assets/Plugins/iOS/Shopify/UnityMessageSender.swift
```diff
@@ -0,0 +1,69 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+import PassKit
+
+typealias MessageCompletion = (_ result: String?) -> Void
```
dbart01: Move into `UnityMessageSender`

dbart01: Ohh, unless it's needed in ObjC? 🤔

Assets/Plugins/iOS/Shopify/UnityMessageSender.swift
```diff
@@ -0,0 +1,69 @@
...
+
+@objc class UnityMessageSender: NSObject {
+
+    static var pendingMessageResponses = [String: UnityMessageResponse]()
+    
+    static func sendMessage(_ message: String, toObject objectName: String, havingMethod methodName: String) -> String? {
+        
+        let message  = UnityMessage.init(message: message);
+        let response = UnityMessageResponse.init(identifier: message.identifier, completion: DispatchSemaphore(value: 0))
```
dbart01: `completion` is a semaphore? That's kinda misleading. If it's a semaphore, just call it that.

Assets/Plugins/iOS/Shopify/UnityMessageSender.swift
```diff
@@ -0,0 +1,69 @@
...
+
+typealias MessageCompletion = (_ result: String?) -> Void
+
+@objc class UnityMessageSender: NSObject {
+
+    static var pendingMessageResponses = [String: UnityMessageResponse]()
+    
+    static func sendMessage(_ message: String, toObject objectName: String, havingMethod methodName: String) -> String? {
+        
+        let message  = UnityMessage.init(message: message);
```
dbart01: Why the explicit `.init` ? Why not just `UnityMessage(message: message)`... Also, that's quite redundant.

dbart01:
Why the distinction between a Unity "request" and "response"? They are always very tightly coupled with a unique identifier. Just curios about the reasoning for that decision.

Assets/Plugins/iOS/Shopify/UnityMessageSender.swift
```diff
@@ -0,0 +1,69 @@
...
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+import PassKit
+
+typealias MessageCompletion = (_ result: String?) -> Void
+
+@objc class UnityMessageSender: NSObject {
```
dbart01: It's not just a sender, it's also the receiver. I'd call it something different like `MessageCenter` or `UnityMessageCenter`. Although I do find the `Unity` prefix quite redundant.

Assets/Plugins/iOS/Shopify/UnityMessageSender.swift
dengjeffrey: I totally forgot that we can do that 😅 . Are you referring to the keyword `init` being redundant or creating a `UnityMessage`?

dengjeffrey:
They're tightly coupled, but I decided to split the two because the other classes don't need to know that a message has a unique identifier when sending a request. Similarly when waiting for a response they would not need the message portion of the request. Some requests don't require a response, so it's not always the case where a unique identifier is needed

Assets/Plugins/iOS/Shopify/UnityMessageSender.swift
```diff
@@ -0,0 +1,69 @@
...
+            return result
+        }
+        
+        return nil
+    }
+    
+    static func sendMessage(_ message: String, toObject objectName: String, havingMethod methodName: String, completionHandler: MessageCompletion?) {
+        
+        let message  = UnityMessage.init(message: message);
+        let response = UnityMessageResponse.init(identifier: message.identifier, completion: DispatchSemaphore(value: 0))
```
dbart01: Ideally, the semaphore would be an implementation detail. Why not have your `MessageResponse` create the semaphore internally and expose only a `wait()` method that would internally wait on the semaphore. To signal the semaphore, you can expose something like `completeWith(response: String? = nil)` that would, again, signal the semaphore internally and update the state with Unity's response.

dbart01: Better still, you can expose a wait method that creates the queue internally:
```swift
response.wait { result in
    pendingMessageResponses[response.identifier] = nil
    completionHandler?(result)
}
```
and in your `_UnityRespondToMessage` binding, you'll update the response,
```objc
[response completeWithResult:result];
```
which would signal the semaphore and run the above completion.

Assets/Plugins/iOS/Shopify/UnityMessageSender.swift
dbart01: The fact that a `UnityMessageResponse` is always created within the same scope as a `UnityMessage` tells me that there's no benefit to having two separate objects represent these actions. To me, a `UnityMessage` is a single unit of work where we send a message, wait for the receiver and complete with an optional result. Splitting these up into two objects that share most of the state doesn't provide any benefit in this particular case. If anything, it adds complexity in keeping the two objects in sync. I think an ideal interface would look something like this:
```swift
// Somewhere at the call site
let message = Message(to: object, method: method, message: "do_something")
UnityMessageSender.send(message)

// inside UnityMessageSender
static func send(_ message: Message) {
    let identifier = message.identifier

    messages[identifier] = message
    UnityInterfaceWrapper.sendMessage(try! message.content(), toObject: message.object, havingMethodName: message.method)

    message.wait { result in
        messages[identifier] = nil
        completionHandler(result)
    }
}
```
The idea is that you're passing `Message` around instead of loose values like `message`, `objectName` and `methodName`. The `Message` should encapsulate those.

Assets/Plugins/iOS/Shopify/UnityMessage.swift
```diff
@@ -31,22 +31,49 @@ enum UnityMessageField: String {
...
+    func wait(withCompletion completion: MessageCompletion?) {
+        DispatchQueue.global(qos: .userInitiated).async {
+            self.semaphore.wait()
+            completion?(self.result)
+        }
+    }
+    
+    func complete(withResult result: String? = nil) {
+        self.result = result;
+        semaphore.signal();
```
dbart01: `;` 🔥

Assets/Plugins/iOS/Shopify/UnityMessageResponder.mm
```diff
@@ -28,13 +28,8 @@
...
-    void _UnityRespondToMessage(const char *identifier, const char *message) {
-        UnityMessageResponse *response = [UnityMessageSender.pendingMessageResponses objectForKey:[NSString stringWithUTF8String:identifier]];
-        
-        if (message != nil) {
-            response.result = [NSString stringWithUTF8String:message];
-        }
-        
-        dispatch_semaphore_signal(response.completion);
+    void _UnityRespondToMessage(const char *identifier, const char *response) {
+        UnityMessage *message = [MessageCenter.messages objectForKey:[NSString stringWithUTF8String:identifier]];
```
dbart01: `messages` should be an implementation detail. Instead, you can expose something like:
```swift
[MessageCenter messageWithIdentifier:identifier];
```

Assets/Plugins/iOS/Shopify/UnityMessage.swift
dengjeffrey: Damn thanks, we need some linter in place

Assets/Plugins/iOS/Shopify/UnityMessageResponder.mm
dengjeffrey: Would it be better if it was all encapsulated in `MessageCenter` then? 

`[MessageCenter respondToMessageHavingIdentifier: identifier, withResult: response]`

dbart01: I think that's needlessly coupling the classes since `MessageCentee` doesn't add any value here. It just proxies the call. It's also less testable.

Assets/Plugins/iOS/Shopify/UnityMessage.swift
```diff
@@ -64,7 +64,7 @@ enum UnityMessageField: String {
     }
     
     func complete(withResult result: String? = nil) {
-        self.result = result;
+        self.result = result
         semaphore.signal();
```
dbart01: `;`

Assets/Plugins/iOS/Shopify/UnityMessage.swift
```diff
@@ -0,0 +1,79 @@
...
+    }
+    
+    func wait(withCompletion completion: MessageCompletion?) {
+        DispatchQueue.global(qos: .userInitiated).async {
+            self.semaphore.wait()
+            completion?(self.response)
+        }
+    }
+    
+    func complete(withResponse response: String? = nil) {
```
dbart01: A little redundant. Would prefer `complete(with response: String? = nil)`

Assets/Plugins/iOS/Shopify/UnityMessage.swift
```diff
@@ -0,0 +1,79 @@
...
+    let identifier: String
+    
+    internal let semaphore: DispatchSemaphore
+    
+    var response: String?
+    
+    // ----------------------------------
+    //  MARK: - Init -
+    //
+    init(content: String, recipientObjectName: String, recipientMethodName: String) {
```
dbart01: Somewhat redundant as well. Could be something like this:
```swift
init(content: String, object: String, method: String) {
```

Assets/Plugins/iOS/Shopify/UnityMessage.swift
```diff
@@ -0,0 +1,79 @@
...
+    public typealias MessageCompletion = (_ response: String?) -> Void
+
+    let recipientObjectName: String
+    let recipientMethodName: String
+    let content: String
+    let identifier: String
+    
+    internal let semaphore: DispatchSemaphore
+    
+    var response: String?
```
dbart01: If we expose a `complete` method that sets the response, that should be the only way to set the response. You can declare this as:
```swift
private(set) var response: String?
```
Doing so only allows `private` mutations but is internal for reading.

Assets/Plugins/iOS/Shopify/UnityMessage.swift
dengjeffrey: Would it not be a bit misleading what `object` and `method` is for someone who is not familiar with `UnitySendMessage`

Assets/Plugins/iOS/Shopify/UnityMessage.swift
```diff
@@ -0,0 +1,79 @@
...
+import Foundation
+
+enum UnityMessageField: String {
+    case message    = "Message"
+    case identifier = "Identifier"
+}
+
+@objc class UnityMessage: NSObject {
+    
+    public typealias MessageCompletion = (_ response: String?) -> Void
```
dbart01: `public typealias` in an `internal` class doesn't make much sense. You can just omit `public` here.

Assets/Plugins/iOS/Shopify/UnityInterfaceWrapper.mm
```diff
@@ -0,0 +1,36 @@
...
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+
+#import "UnityInterfaceWrapper.h"
+
+@implementation UnityInterfaceWrapper
+
++ (void)sendMessage:(NSString *)message toObject:(NSString *)objectName havingMethodName:(NSString *)methodName {
```
dbart01: The Objective-C naming convention for this can be taken from the existing target-selector pattern:
```objc
+ (void)sendMessage:(NSString *)message object:(NSString *)object method:(NSString *)method {
```
In this case, instead of `selector:(SEL)selector`, we replace it with `method`.

Assets/Plugins/iOS/Shopify/UnityMessage.swift
dbart01: The type makes it very clear, in my opinion. Swift naming conventions state that we should optimize for the readability at the call site, which I think it does very well:
```swift
let message = UnityMessage(content: "", object: "MyObject", method: "doSomething")
```
With basic Unity knowledge, the reader would know right away what it's supposed to be doing.

Assets/Plugins/iOS/Shopify/UnityMessage.swift
dengjeffrey: Following Swift naming I believe the `Response` needs to be there since String here is a fundamental type, so it provides weak type information. It's not evidently clear what the String is to someone who is just reading it. It could be an error or message identifier etc.

Assets/Plugins/iOS/Shopify/UnityMessage.swift
```diff
@@ -0,0 +1,79 @@
...
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+enum UnityMessageField: String {
```
dbart01: You can nest this inside `UnityMessage` and avoid the long name:
```swift
class UnityMessage: NSObject {
    enum Field: String {
        case message = "Message"
        case identifier = "Identifier"
    }
    ...
}
```

Assets/Plugins/iOS/Shopify/UnityMessage.swift
dbart01: Someone implementing `UnityMessage` will almost always be aware that messages to Unity are stringly-typed. Regardless, though, I don't think this:
```swift
let response = getResponseFromSomewhere()
message.complete(withResponse: response)
```
is any more clear than:
```swift
let response = getResponseFromSomewhere()
message.complete(with: response)
```
Guidelines for using an API should be the burden of syntax. That's why we have comments and documentation. After someone understands that `response` is always a string returned by Unity, they'll still continue to incur the cognitive and visual overhead of reading `complete(withResponse: response)`.

Have a look at [Swift API design guidelines](https://swift.org/documentation/api-design-guidelines/), it explains what I'm trying to say much better.

dengjeffrey: I was specifically looking at this part of the [Swift API design guidelines](https://swift.org/documentation/api-design-guidelines/#weak-type-information), but I believe in this case you're right. Since there is only one use case it is pretty clear what is happening at the use-site.

Assets/Plugins/iOS/Shopify/UnityMessage.swift
dengjeffrey: This might be worthwhile to revisit in a different PR. Since classes that implement Serializable have the `Field` outside class scope, it would be worthwhile to refactor all of them to avoid namespace collision.

Assets/Plugins/iOS/Shopify/UnityMessage.swift
dbart01: I guess the optimal solution depends on what we're trying to achieve. If, for example, `UnityMessage` is a `public` type and will be used outside of the the module, exposing weak type information for a public API is not desirable. To solve that you can introduce a wrapper type:
```swift
extension UnityMessage {
    struct ReturnValue {
        let value: String  
    }
}
``` 
which would make the API very explicit.
```swift
let response = getResponseFromSomewhere()
let value = UnityMessage.ReturnValue(response)
message.complete(with: value)
```
Not sure if the benefit is worth the added cost of testing and maintenance but that is an alternative.

Assets/Plugins/iOS/Shopify/UnityMessage.swift
dbart01: Another thing to note here is that internal implementation of classes should not pollute the module namespace. `UnityMessageField` is implementation detail so it's best nested and declared `private`. I've also adjusted my code snippet above to reflect nesting inside an extension rather than the class itself.

Assets/Plugins/iOS/Shopify/MessageCenter.swift
```diff
@@ -32,14 +32,14 @@ import PassKit
...
-        UnityInterfaceWrapper.sendMessage(try! message.serializedString(), toObject: message.recipientObjectName, havingMethodName: message.recipientMethodName)
+        MessageSender.sendMessage(try! message.serializedString(), object: message.object, method: message.method)
     }
     
     static func send(_ message: UnityMessage, completionHandler: UnityMessage.MessageCompletion?) {
         
         messages[message.identifier] = message
         
-        UnityInterfaceWrapper.sendMessage(try! message.serializedString(), toObject: message.recipientObjectName, havingMethodName: message.recipientMethodName)
+        MessageSender.sendMessage(try! message.serializedString(), object: message.object, method: message.method)
```
dbart01: Why not just pass the `UnityMessage` in here?

dengjeffrey: I believe that since `MessageSender` is included in the `Unity-iPhone-Bridging-Header.h` we would be unable to include the Swift bridging header here. And `UnityMessage` is defined in Swift. `MessageSender` is really just an Objective-C wrapper, but named as `MessageSender` to make it more clear of what it is doing

dbart01: Then why does `UnityMessage` inherit from `NSObject` and is declared `@objc`?

Assets/Plugins/iOS/Shopify/MessageSender.h
```diff
@@ -0,0 +1,35 @@
...
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+
+#import <Foundation/Foundation.h>
+#import "UnityInterface.h"
```
dbart01: Does this have to be exposed publicly? Can it not be included in the `.mm` file?

Assets/Plugins/iOS/Shopify/MessageCenter.swift
dengjeffrey: It's declared @objc so it can be used in 'MessageResponder', 'MessageResponder' has no functions exposed to swift so it is fine.

dbart01: Why does it matter if it has functions exposed to Swift?

dengjeffrey: I was mislead when I got a compile error when trying to include the bridging header, so I thought It had something to do with recursive imports. But I've found the fix.

dengjeffrey: Actually I realized we don't need this class at all, we can define the C++ method in the bridging header

Assets/Plugins/iOS/Shopify/MessageCenter.swift
```diff
@@ -31,15 +31,15 @@ import PassKit
...
-        
+
         messages[message.identifier] = message
+    
+        let object  = message.object.cString(using: .utf8)
+        let method  = message.method.cString(using: .utf8)
+        let content = try! message.serializedString().cString(using: .utf8)
         
-        MessageSender.sendMessage(try! message.serializedString(), object: message.object, method: message.method)
+        UnitySendMessage(object, method, content);
```
dbart01: Very nice 👍

Assets/Plugins/iOS/Shopify/Unity-iPhone-Bridging-Header.h
```diff
@@ -26,4 +26,4 @@
 
 #import "UnityBuyAppController.h"
 #import "UnityAppController+ViewHandling.h"
-#import "MessageSender.h"
+#import "UnityInterface.h"
```
dbart01: Awesome 👍

Assets/Plugins/iOS/Shopify/MessageCenter.swift
```diff
@@ -0,0 +1,53 @@
...
+    
+    static func send(_ message: UnityMessage, completionHandler: UnityMessage.MessageCompletion?) {
+
+        messages[message.identifier] = message
+    
+        let object  = message.object.cString(using: .utf8)
+        let method  = message.method.cString(using: .utf8)
+        let content = try! message.serializedString().cString(using: .utf8)
+        
+        UnitySendMessage(object, method, content);
```
dbart01: `;` 🔥

Assets/Plugins/iOS/Shopify/ApplePayEventDispatcher.swift
```diff
@@ -0,0 +1,133 @@
...
+    //
+    init(recieverName: String) {
+        self.receiverName = recieverName
+        super.init()
+    }
+    
+    func call(method: MethodName, withValue value: String, completionHandler: AppleEventCompletion?) {
+        
+        let message = UnityMessage(content: value, object: receiverName, method: method.rawValue)
+        MessageCenter.send(message) { (serializedResponse: String?) in
```
dbart01: Nit: Can just write `MessageCenter.send(message) { serializedResponse in`

dbart01: This format can also be applied to other instance of completion handler invocations. Declaring explicit types isn't necessary.

Assets/Plugins/iOS/Shopify/ApplePayEventDispatcher.swift
```diff
@@ -0,0 +1,133 @@
...
+import UIKit
+import PassKit
+
+typealias AppleEventCompletion = (_ response: ApplePayEventResponse?) -> Void
+
+class ApplePayEventDispatcher: NSObject {
+    
+    let receiverName: String
+    
+    enum MethodName: String {
```
dbart01: `private`

Assets/Plugins/iOS/Shopify/ApplePayEventDispatcher.swift
```diff
@@ -0,0 +1,133 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import UIKit
+import PassKit
+
+typealias AppleEventCompletion = (_ response: ApplePayEventResponse?) -> Void
```
dbart01: Nest under `ApplePayEventDispatcher`.

Assets/Plugins/iOS/Shopify/ApplePayEventDispatcher.swift
dbart01: You can also simplify the name to just `Method`.

Assets/Plugins/iOS/Shopify/ApplePayEventDispatcher.swift
```diff
@@ -0,0 +1,133 @@
...
+    
+    // ----------------------------------
+    //  MARK: - Init -
+    //
+    init(recieverName: String) {
+        self.receiverName = recieverName
+        super.init()
+    }
+    
+    func call(method: MethodName, withValue value: String, completionHandler: AppleEventCompletion?) {
```
dbart01: This method signature can be optimized for call site clarity:
```swift
func call(_ method: Method, value: String, completionHandler: AppleEventCompletion? = nil)
```
Example call site:
```swift
call(.didFinishCheckoutSession, value: status.description)
```

Assets/Plugins/iOS/Shopify/ApplePayEventDispatcher.swift
```diff
@@ -0,0 +1,133 @@
...
+            completion(.failure, [PKPaymentSummaryItem]())
+            return
+        }
+        
+        call(method: .updateSummaryItemsForShippingIdentifier, withValue: identifier) { response in
+            guard
+                let response     = response,
+                let authStatus   = response.authorizationStatus,
+                let summaryItems = response.summaryItems
+            else {
```
dbart01: I like this `guard` formatting a lot 👍

Assets/Plugins/iOS/Shopify/ApplePayEventDispatcher.swift
```diff
@@ -0,0 +1,133 @@
...
+                return
+            }
+            completion(authStatus, summaryItems)
+        }
+    }
+    
+    func paymentSession(_ session: PaymentSession, didSelectShippingContact shippingContact: PKContact, completion: @escaping (PKPaymentAuthorizationStatus, [PKShippingMethod], [PKPaymentSummaryItem]) -> Void) {
+        
+        guard let contactString = try? shippingContact.serializedString() else {
+            completion(.failure, [PKShippingMethod](), [PKPaymentSummaryItem]())
```
dbart01: Type information isn't necessary here. You can write this as:
```swift
completion(.failure, [], [])
```

Assets/Plugins/iOS/Shopify/ApplePayEventResponse.swift
```diff
@@ -47,43 +47,37 @@ struct ApplePayEventResponse {
...
     //
-    init?(serialized: String) {
-
-        guard
-            let data = serialized.data(using: .utf8),
-            let jsonObject = (try? JSONSerialization.jsonObject(with: data)) as? JSON else {
-                return nil
-        }
-
+    static func deserialized(_ serialized: JSON) -> ApplePayEventResponse? {
```
dbart01: A method name should be a verb. Perhaps
```swift
static func deserialize(_ json: JSON) -> ApplePayEventResponse?
```

Assets/Plugins/iOS/Shopify/Deserializable/Deserializable.swift
```diff
@@ -0,0 +1,60 @@
...
+
+import Foundation
+
+protocol Deserializable {
+    static func deserialized(_ serialized: JSON) -> Self?
+}
+
+extension Deserializable {
+    
+    static func deserialized(_ serialized: String) -> Self? {
```
dbart01: `serialized` is pretty ambiguous. Why not just call it like it is - `_ string: String`?

Assets/Plugins/iOS/Shopify/Deserializable/Deserializable.swift
```diff
@@ -0,0 +1,60 @@
...
+            let data = serialized.data(using: .utf8),
+            let jsonObject = (try? JSONSerialization.jsonObject(with: data)) as? JSON
+        else {
+            return nil
+        }
+
+        return deserialized(jsonObject)
+    }
+    
+    static func items(fromSerialized jsons: [JSON]) -> [Self]? {
```
dbart01: Why not continue following the naming convention here? After all, a collection of `JSON` objects is still a "serialized value":
```swift
static func deserialize(_ jsonCollection: [JSON]) -> [Self]?
```

Assets/Plugins/iOS/Shopify/Deserializable/Deserializable.swift
```diff
@@ -0,0 +1,60 @@
...
+        
+        for json: JSON in jsons {
+            
+            guard let item = deserialized(json) else {
+                return nil
+            }
+            
+            items.append(item)
+        }
+        return items
```
dbart01: The entire method body can be written more concisely as:
```swift
return jsonCollection.flatMap {
    self.deserialize($0)
}
```

Assets/Plugins/iOS/Shopify/Deserializable/PKShippingMethod+Deserializable.swift
```diff
@@ -27,46 +27,37 @@
...
-enum ShippingMethodField: String {
-    case detail
-    case identifier
-}
-
 extension PKShippingMethod {
     
-    convenience init?(serializedShippingMethod json:JSON) {
-        
+    enum Field: String {
```
dbart01: `private`

Assets/Plugins/iOS/Shopify/Deserializable/PKPaymentSummaryItem+Deserializable.swift
```diff
@@ -27,42 +27,34 @@
...
-enum PaymentSummaryItemField: String {
-    case amount
-    case label
-    case type
-}
-
-extension PKPaymentSummaryItem {
+extension PKPaymentSummaryItem: Deserializable {
+    
+    enum Field: String {
```
dbart01: `private`

Assets/Plugins/iOS/Shopify/MessageDispatcher.swift
```diff
@@ -0,0 +1,55 @@
...
+
+protocol MessageDispatcher {
+    typealias DispatchCompletion<T: Deserializable> = (_ response: T?) -> Void
+    var receiver: String { get }
+}
+
+extension MessageDispatcher {
+    
+    func call<T: RawRepresentable>(_ method: T, value: String) where T.RawValue == String {
+        UnitySendMessage(receiver, method.rawValue, value)
```
dbart01: `UnitySendMessage` shouldn't be called outside of `MessageCenter`. You should call through to `MessageCenter` instead.

Assets/Plugins/iOS/Shopify/MessageCenter.swift
```diff
@@ -0,0 +1,53 @@
...
+
+        messages[message.identifier] = message
+    
+        let object  = message.object.cString(using: .utf8)
+        let method  = message.method.cString(using: .utf8)
+        let content = try! message.serializedString().cString(using: .utf8)
+        
+        UnitySendMessage(object, method, content)
+        
+        message.wait { response in
```
dbart01: The `wait` is actually just "wasted effort" if `completionHandler` is `nil`. Instead, you should only store the message and wait for completion if a callback is present. Also, you can give `completionHandler` a default value so that you have a unified interface for method calls that don't provide a callback.
```swift
static func send(_ message: UnityMessage, completionHandler: UnityMessage.MessageCompletion? = nil) {
    
     if let completion = completionHandler {
         messages[message.identifier] = message
         
         defer {
             message.wait { response in
                 messages[message.identifier] = nil
                 completion(response)
             }
         }
     }
 
     let object  = message.object.cString(using: .utf8)
     let method  = message.method.cString(using: .utf8)
     let content = try! message.serializedString().cString(using: .utf8)
     
     UnitySendMessage(object, method, content)
 }
```

Assets/Plugins/iOS/Shopify/ApplePayEventResponse.swift
dengjeffrey: Should we not be using the `-ed` ending here since it is a nonmutating method?
[Side Effect Naming](https://swift.org/documentation/api-design-guidelines/#name-according-to-side-effects)

Assets/Plugins/iOS/Shopify/Deserializable/Deserializable.swift
dengjeffrey: Awesome 👍

Assets/Plugins/iOS/Shopify/ApplePayEventResponse.swift
dbart01: That convention is designed to disambiguate methods that **can** potentially be mutating. Such as method that have behaviour on the received. For example `array.append(value)`. This is a mapping function. It takes `X` returns `Y`. This simply cannot be mutating by nature. Think of it as a `map` or `flatMap`.

Assets/Plugins/iOS/Shopify/Deserializable/Deserializable.swift
```diff
@@ -0,0 +1,58 @@
...
+
+        return deserialize(jsonObject)
+    }
+    
+    static func deserialize(_ jsonCollection: [JSON]) -> [Self]? {
+        let items = jsonCollection.flatMap {
+            self.deserialize($0)
+        }
+        
+        if (items.count < jsonCollection.count) {
```
dbart01: `( )` 🔥

Assets/Plugins/iOS/Shopify/UnityMessage.swift
dbart01: It seems that you've nested other `Field` structures but not this one?

Assets/Plugins/iOS/Shopify/MessageDispatcher.swift
```diff
@@ -0,0 +1,55 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+protocol MessageDispatcher {
+    typealias DispatchCompletion<T: Deserializable> = (_ response: T?) -> Void
+    var receiver: String { get }
```
dbart01: Can you not just pass this as an argument to `MessageCenter.send()`? This protocol doesn't seem to have much value.

dbart01: There's already an object that couples and encapsulates everything required to send a message to Unity - `UnityMessage`. This protocol decouples the `receiver` from the other values. I think that just makes the flow hard to follow and more confusing.

dengjeffrey: I believe having this protocol will create a safer 1-to-1 mapping between Unity classes and the classes in plugin. Each Unity class we need to interact with will be represented as a `dispatcher` and defined methods `enum`.

dbart01: How exactly makes it safer? The nice thing about the `UnityMessage` abstraction is that it isn't coupled with any specific Unity class. Imagine at some point later on, you might introduce a class on the Unity side that handles all communication with Swift (inverse of what we have here). That 1-1 coupling no longer makes sense.

dengjeffrey: Instead of a 1-1 per class, this abstraction is more like 1-1 per `GameObject`. It makes it safer because it removes the possibility of making a mistake where you mismatch the `GameObject` name and the methods that they handle. In Unity a `GameObject` holds multiple `components`, which are classes, at any moment these components can be added or removed. 

In my thought process, I thought that it would be more familiar to the Unity developer if each `component` was represented by `Dispatcher`

dengjeffrey: After giving it some thought, I was wrong to assume that this would create more safety, since we can't enforce the call-site to use a method enum defined by the dispatcher.

Assets/Plugins/iOS/Shopify/String/PKPaymentAuthorizationStatus+String.swift
```diff
@@ -27,35 +27,53 @@
...
-                return .pinIncorrect
-            } else {
-                return nil
-            }
-        case "PinLockout":
-            if #available(iOS 9.2, *) {
-                return .pinLockout
-            } else {
-                return nil
+        for status in allItems {
```
dbart01: Why that change? The `switch` implementation is easier to understand and is more efficient (not that it makes a difference).

dengjeffrey: I needed the backwards compatibility to create mocks for the test cases. The alternative was to redefine them in tests but that would not be very DRY

dbart01: Why not just let the class create itself:
```swift
extension PKPaymentAuthorizationStatus: RawRepresentable {

    init?(rawValue: String) {
        switch rawValue {
             ...
        }
    }

    var rawValue: String {
        switch self {
            ...
        }
    }
}
```

Assets/Plugins/iOS/Shopify/UnityMessage.swift
```diff
@@ -0,0 +1,81 @@
...
+    }
+    
+    func complete(with response: String? = nil) {
+        self.response = response
+        semaphore.signal()
+    }
+}
+
+extension UnityMessage {
+    enum Field: String {
```
dbart01: `private` / `fileprivate`

Assets/Plugins/iOS/Shopify/ApplePayEventResponse.swift
```diff
@@ -0,0 +1,86 @@
...
+        
+        return self.init(authorizationStatus: authorizationStatus, summaryItems: summaryItems, shippingMethods: shippingMethods)
+    }
+}
+
+extension ApplePayEventResponse {
+    
+    /// Expect the response from events to follow [String: Any] JSON format.
+    /// These are the keys that they are expected to follow
+    enum ResponseKey: String {
```
dbart01: `private` / `fileprivate`

dengjeffrey:
Good to go 👍 ?

Assets/Plugins/iOS/Shopify/ApplePayEventDispatcher.swift
```diff
@@ -0,0 +1,131 @@
...
+    }
+    
+    private func call(method: Method, value: String, completionHandler: AppleEventCompletion? = nil) {
+        
+        let message = UnityMessage(content: value, object: receiver, method: method.rawValue)
+        
+        if let completionHandler = completionHandler {
+            
+            MessageCenter.send(message) { response in
+                if let response = response {
```
dbart01: This check is detrimental. The `completionHandler` accepts an optional response, you should just pass it straight through. Otherwise, if the response is `nil`, the `completionHandler` will never be called.

dengjeffrey: nice catch 👍 !

PR #93, opened 2017-06-08
[Feature] iOS test scripts

dengjeffrey:
#### What
+ Adds a test script that builds and runs tests on the Xcode project
+ Use the build pipeline instead of listening for PostProcessing events
+ Does not add integration into CI

Requires #96 
<hr>

#### How/Why
+ Using the build pipeline we don't enforce everyone to use our PostProcessor

### DIFF
--- a/scripts/generator/graphql_generator/csharp/Editor/ShopifyBuildPipeline.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/Editor/ShopifyBuildPipeline.cs.erb
@@ -1,0 +1,61 @@
+namespace <%= namespace %>.SDK {
+    using System.Collections;
+    using System.Collections.Generic;
+    using System;
+    using System.IO;
+    using UnityEngine;
+    using UnityEditor;
+
+    public class ShopifyBuildPipeline : MonoBehaviour {
+
+        private static string PlayerPathArg = "-buildPlayerPath";
+        private static string PlayerPath {
+            get {
+                string[] arguments = Environment.GetCommandLineArgs();
+                string playerPath = ShopifyBuildPipeline.PlayerPathFromArguments(arguments);
+
+                if (playerPath == null) {
+                    playerPath = "./Shopify-iOS";
+                }
+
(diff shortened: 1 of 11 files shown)


scripts/generator/graphql_generator/csharp/Editor/ShopifyBuildPipeline.cs.erb
```diff
@@ -0,0 +1,61 @@
+namespace <%= namespace %>.SDK {
```
mikkoh: Is there a reason why will be bundling this into the `.unitypackage` that will be distributed to users?

scripts/generator/graphql_generator/csharp/Editor/ShopifyBuildPipeline.cs.erb
```diff
@@ -0,0 +1,61 @@
...
+    using System.Collections;
+    using System.Collections.Generic;
+    using System;
+    using System.IO;
+    using UnityEngine;
+    using UnityEditor;
+
+    public class ShopifyBuildPipeline : MonoBehaviour {
+        private static string PlayerPathArg = "-buildPlayerPath";
```
mikkoh: This should be `private const string PlayerPathArg = "-buildPlayerPath";`

scripts/generator/graphql_generator/csharp/Editor/ShopifyBuildPipeline.cs.erb
dengjeffrey: I guess that depends if we are considering bundling the sample app into the `.unitypackage`. 
If we are including a sample app, we would ideally have our own Build pipeline that would configure the Xcode project so that the sample app will run.

mikkoh: When you talk about the sample app is that what @richardmonette is building? If so maybe this script should be distributed in that app. The same app won't live in this same repo.

PR #94, opened 2017-06-08
[Feature] Merge utils

mikkoh:
This PR implements the ability to merge response json objects.

When working with checkouts we want the ability to update a previously queried `Checkout`. For instance let's say that we are polling we want to update `ready` on an existing `Checkout` we've previously queried.

One curious thing in this PR is `DoNotMergeIfExists`. There maybe cases where we do not want to merge one field but merge all other fields. This for instance could be used when you want to handle a field on a case by case basis. For instance with `Checkouts` sometimes we may want to merge in line items sometimes we may want to leave them be.

### DIFF
--- a/Assets/Editor/TestResponeMergeUtil.cs
+++ b/Assets/Editor/TestResponeMergeUtil.cs
@@ -2,6 +2,7 @@ namespace Shopify.Tests
 {
     using NUnit.Framework;
     using Shopify.Unity.SDK;
+    using System.Collections;
     using System.Collections.Generic;
 
     [TestFixture]
@@ -39,6 +40,38 @@ namespace Shopify.Tests
             Assert.AreEqual("i am b", merged["fieldInBoth"], "Overwrote field in A from B");
             Assert.AreNotEqual(responseA["fieldHasDictionary"], merged["fieldHasDictionary"]);
             Assert.AreNotEqual(responseB["fieldHasDictionary"], merged["fieldHasDictionary"]);
+        }
+
+        [Test]
+        public void CustomFieldMerger() {
+            Dictionary<string, object> responseA = GetResponseA();
+            Dictionary<string, object> responseB = GetResponseB();
+
+            ResponseMergeUtil merger = new ResponseMergeUtil();
+
+            merger.AddFieldMerger("fieldInBoth", (string field, IDictionary into, IDictionary dataA, IDictionary dataB) => {
+                into[field] = (string) dataA[field] + (string) dataB[field];
+            });
+
+            Dictionary<string, object> merged = merger.Merge(responseA, responseB);
+
+            Assert.AreEqual("i am ai am b", merged["fieldInBoth"], "Custom merger worked");
+        }
+
+        [Test]
+        public void MergeListsConcat() {
+            Dictionary<string, object> responseA = GetResponseA();
+            Dictionary<string, object> responseB = GetResponseB();
+
+            ResponseMergeUtil merger = new ResponseMergeUtil();
+
+            merger.AddFieldMerger("fieldHasList", ResponseMergeUtil.MergeListsConcat);
+
+            Dictionary<string, object> merged = merger.Merge(responseA, responseB);
+
+            Assert.AreEqual(2, ((List<object>) merged["fieldHasList"]).Count, "Has two elements");
+            Assert.AreEqual(0, ((List<object>) merged["fieldHasList"])[0], "First value was correct");
+            Assert.AreEqual(1, ((List<object>) merged["fieldHasList"])[1], "Second value was correct");
         }
 
         private Dictionary<string,object> GetResponseA() {
--- a/Assets/Editor/TestResponeMergeUtil.cs
+++ b/Assets/Editor/TestResponeMergeUtil.cs
@@ -82,11 +82,12 @@ namespace Shopify.Tests
 
             ResponseMergeUtil merger = new ResponseMergeUtil();
 
-            merger.AddFieldMerger("fieldInBoth", ResponseMergeUtil.DoNotMerge);
+            merger.AddFieldMerger("fieldInBoth", ResponseMergeUtil.DoNotMergeIfExists);
 
             Dictionary<string, object> merged = merger.Merge(responseA, responseB);
 
-            Assert.AreEqual("i am a", merged["fieldInBoth"], "Custom merger worked");
+            Assert.AreEqual("i am a", merged["fieldInBoth"], "Did not merge when existed");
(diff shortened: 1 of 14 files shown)


scripts/generator/graphql_generator/csharp/SDK/ResponseMergeUtil.cs.erb
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public void AddObjectMerger(string field, ResponseMergeUtil merger) {
+            ObjectMergers[field] = merger;
+        }
+
+        public Dictionary<string, object> Merge(IDictionary responseA, IDictionary responseB) {
+            Dictionary<string, object> MergedResponse = new Dictionary<string, object>();
+
+            foreach(string field in responseA.Keys) {
```
richardmonette: Are you just trying to copy responseA into MergedResponse? I think you could just pass responseA as an argument on line 47 to have the same effect.

scripts/generator/graphql_generator/csharp/SDK/ResponseMergeUtil.cs.erb
```diff
@@ -0,0 +1,69 @@
...
+        public static void MergeOverwrite(string field, IDictionary into, IDictionary responseA, IDictionary responseB) {
+            into[field] = responseB[field];
+        }
+
+        public static void MergeListsConcat(string field, IDictionary into, IDictionary responseA, IDictionary responseB) {
+            List<object> newList = new List<object>();
+            IList listA = (IList) responseA[field];
+            IList listB = (IList) responseB[field];
+
+            foreach(object item in listA) {
```
richardmonette: I think you can just do `newList.AddRange(listA)`

mikkoh: Good catch :+1: previously I had `newList` be and `IList` so this is a left over from this.

Assets/Editor/TestResponeMergeUtil.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            ResponseMergeUtil merger = new ResponseMergeUtil();
+
+            merger.AddObjectMerger("fieldHasDictionary", new ResponseMergeUtil());
+
+            Dictionary<string, object> merged = merger.Merge(responseA, responseB);
+
+            Assert.AreEqual(3, merged["fieldOnlyInA"], "Brought in field from A");
+            Assert.AreEqual("i am b", merged["fieldInBoth"], "Overwrote field in A from B");
+            Assert.IsTrue(!Object.ReferenceEquals(responseA["fieldHasDictionary"], merged["fieldHasDictionary"]), "Creted a new nested object that was not responseA");
```
dengjeffrey: Nit: `Created a new ...`

Assets/Editor/TestResponeMergeUtil.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            Dictionary<string, object> merged = merger.Merge(responseA, responseB);
+
+            Assert.AreNotEqual(responseA, merged, "the merged dictionary is a dictionary. Is not responseA");
+            Assert.AreNotEqual(responseB, merged, "the merged dictionary is a dictionary. Is not responseB");
+            Assert.AreEqual(3, merged["fieldOnlyInA"], "Brought in field from A");
+            Assert.AreEqual("i am b", merged["fieldInBoth"], "Overwrote field in A from B");
+            Assert.AreEqual(1, ((List<int>) merged["fieldHasList"])[0], "Overwrote field in A from B that is a list");
+            Assert.AreEqual("monkey", ((Dictionary<string, string>) merged["fieldHasDictionary"])["keyA"], "Overwrote field in A from B that is a list");
+            Assert.AreEqual(responseB["fieldHasDictionary"], merged["fieldHasDictionary"]);
```
dengjeffrey: If we make this assertion, will we need line 25?

scripts/generator/graphql_generator/csharp/SDK/ResponseMergeUtil.cs.erb
```diff
@@ -0,0 +1,69 @@
...
+        public void AddFieldMerger(string field, MergeFieldDelegate merger) {
+            FieldMergers[field] = merger;
+        }
+
+        public void AddObjectMerger(string field, ResponseMergeUtil merger) {
+            ObjectMergers[field] = merger;
+        }
+
+        public Dictionary<string, object> Merge(IDictionary responseA, IDictionary responseB) {
+            Dictionary<string, object> MergedResponse = new Dictionary<string, object>();
```
dengjeffrey: `mergedResponse`?

scripts/generator/graphql_generator/csharp/SDK/ResponseMergeUtil.cs.erb
mikkoh: We want to clone the list instead of assign by reference. Just trying to ensure that no mutations happen down the line to the originally queried data.

Assets/Editor/TestResponeMergeUtil.cs
mikkoh: Yes because what if it down the line someone implements copying/cloning `fieldHasDictionary` instead of the expected default outcome of set by reference. Although this should use `Object.ReferenceEquals` instead.

scripts/generator/graphql_generator/csharp/SDK/ResponseMergeUtil.cs.erb
richardmonette: IIRC using the copy constructor achieves what you have in mind, that is, changing the copy or original does not affect the other https://www.dotnetperls.com/copy-dictionary

mikkoh:
@richardmonette @dengjeffrey I converted all references to `IDictionary` to be `Dictionary<string, object>` since we're dealing with JSON data and `MiniJSON` always just uses `Dictionary<string, object>`.

This allows us to clone by doing syntax like:
```
Dictionary<string, object> mergedResponse = new Dictionary<string, object>(responseA);
```

Assets/Editor/TestResponeMergeUtil.cs
```diff
@@ -17,6 +17,8 @@ public class TestResponseMergeUtil {
 
             Dictionary<string, object> merged = merger.Merge(responseA, responseB);
 
+            Assert.IsTrue(!Object.ReferenceEquals(responseA, merged));
```
dengjeffrey: I think this could be more readable if we used `Assert.IsFalse`

PR #95, opened 2017-06-08
Query products using nodes instead of multiple node calls

richardmonette:
Currently, we are querying for products by doing repeated trips to the server querying a single product at a time. This seem sub-optimal and doesn't really take advantage of how graphql works. 

This PR changes this around so instead we do one query on nodes, with all the ids we require.


richardmonette:
Rebased!

scripts/generator/graphql_generator/csharp/ShopifyBuy.cs.erb
```diff
@@ -3,6 +3,7 @@ namespace <%= namespace %> {
     using System.Text;
     using System.Collections;
     using System.Collections.Generic;
+    using System.Diagnostics;
```
sleroux: Nit: Don't think we need this using

scripts/generator/graphql_generator/csharp/ShopifyBuy.cs.erb
```diff
@@ -405,31 +406,25 @@ namespace <%= namespace %> {
...
-                if (error != null) { 
+                if (error != null) {
                     callback(null, error);
                 } else {
                     List<Product> products = new List<Product>();
 
-                    for (int i = 0; i < productIds.Count; i++) {
-                        Product product = (Product) response.data.node(DefaultQueries.GetAliasFromIndex(i));
-
+                    foreach (Shopify.Unity.Product product in response.data.nodes()) {
```
sleroux: Nit: You can probably use var here

richardmonette: Tried using `var`, but then there is a compiler error about the `Add()` not having a correct overload.

PR #98, opened 2017-06-14
[Feature] Added Enum String Tests

dengjeffrey:
+ Adds tests for Enums with string extensions


dbart01:
My opinion on this approach is that it seems a little over-engineered. Two separate files add extensions to enums to expose `.allItems`, which require writing additional tests and now maintaining not only the extensions but the tests for them as well. Why no just explicitly test each case:
```swift
XCTAssertEqual(PKPaymentAuthorizationStatus.value1, PKPaymentAuthorizationStatus(rawValue: "value1"))
XCTAssertEqual(PKPaymentAuthorizationStatus.value2, PKPaymentAuthorizationStatus(rawValue: "value2"))
XCTAssertEqual(PKPaymentAuthorizationStatus.value3, PKPaymentAuthorizationStatus(rawValue: "value3"))
...
```

dengjeffrey:
I believe by storing the possible types in an array it reduces the amount of clutter and repetitiveness. Perhaps it should be just a local array?

dbart01:
It's entirely up to you. I actually think it **adds** clutter instead of reducing it. The less code you write, the less code you have to maintain. Especially when it comes to tests. I think tests are one of the very few places where the DRY principle is of less importance if it means the tests are easier to read and maintain. Just my opinion, though.

dengjeffrey:
@dbart01 I do agree with having it explicitly shown, so you know exactly which assert failed. I've made some changes trying to reduce the amount of extra functionality, while still reducing the amount of boiler text.

Assets/Plugins/iOS/Shopify/BuyTests/StringTests.swift
```diff
@@ -29,37 +29,60 @@ import PassKit
...
     
     /// Test that rawValues correspond to the init
     func testPaymentAuthorizationStatusInit() {
-        for status in PKPaymentAuthorizationStatus.allItems {
-            XCTAssertEqual(status, PKPaymentAuthorizationStatus(rawValue: status.rawValue as String))
+        
+        let assertInitIdempotency = { (status: PKPaymentAuthorizationStatus) in
+            XCTAssertEqual(
+                status,
+                PKPaymentAuthorizationStatus.init(rawValue: status.rawValue as String),
```
dbart01: `.init` isn't necessary -> `PKPaymentAuthorizationStatus(rawValue: ...)`

Assets/Plugins/iOS/Shopify/BuyTests/StringTests.swift
```diff
@@ -29,37 +29,60 @@ import PassKit
...
+        XCTAssertEqual(types.count, Set(types).count)
     }
     
     /// Test that rawValues correspond to the init
     func testPaymentAuthorizationStatusInit() {
-        for status in PKPaymentAuthorizationStatus.allItems {
-            XCTAssertEqual(status, PKPaymentAuthorizationStatus(rawValue: status.rawValue as String))
+        
+        let assertInitIdempotency = { (status: PKPaymentAuthorizationStatus) in
+            XCTAssertEqual(
```
dbart01: What is this testing? When will it ever fail?

dengjeffrey: Tests that that the 'rawValue' of an state, is the same value used for initialization of that state.

This fails when a 'rawValue' of a state is different than the raw value that would be given to the initialization to create that state.

Assets/Plugins/iOS/Shopify/BuyTests/StringTests.swift
```diff
@@ -0,0 +1,88 @@
...
+        assertInitIdempotency(.pinLockout)
+        assertInitIdempotency(.pinIncorrect)
+    }
+    
+    func testPaymentSummaryItemTypeInit() {
+        
+        let assertInitIdempotency = { (status: PKPaymentSummaryItemType) in
+            XCTAssertEqual(
+                status,
+                PKPaymentSummaryItemType.init(rawValue: status.rawValue as String),
```
dbart01: `.init` 🔥

Assets/Plugins/iOS/Shopify/BuyTests/StringTests.swift
```diff
@@ -0,0 +1,88 @@
...
+        XCTAssertEqual(types.count, Set(types).count)
+    }
+    
+    /// Test that rawValues correspond to the init
+    func testPaymentAuthorizationStatusInit() {
+        
+        let assertInitIdempotency = { (status: PKPaymentAuthorizationStatus) in
+            XCTAssertEqual(
+                status,
+                PKPaymentAuthorizationStatus(rawValue: status.rawValue as String),
```
dbart01: Is the cast necessary?

dengjeffrey: Yes, the compiler doesn't know if we are referencing to the `init(rawValue: int)` or `init(rawValue: string)`

dbart01: How can there be two duplicate conformances to `RawRepresentable`?

dengjeffrey: Spoke with @dbart01, the issue of having an ambiguous conformance only happens on test cases

PR #100, opened 2017-06-14
[Feature] iOS Unity Messages Tests

dengjeffrey:
+ Adds tests for sending messages
+ Sends a message to a Unity object and waits for a repeat of that message
+ Fixes compile issue regarding deployment target 

<hr>

#### Extra details
+ A test scene with a `Tester` object is used to facilitate message sending
+ The iOS test cases must wait till the `Tester` object has been loaded into the scene before attempting to send messages to it


Assets/Plugins/iOS/Shopify/BuyTests/UnityTestHelpers/Tester.h
```diff
@@ -0,0 +1,34 @@
...
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+#include <Foundation/Foundation.h>
+
+@interface Tester : NSObject
+
+@property (class) BOOL hasLoaded;
+@property (class, nullable) void (^loadCompletion)();
```
dbart01: Nit: Might be cleaner as a `typedef void (^TesterCompletion)`. Then it can be declared as:
```objc
@property (class, nullable) TesterCompletion completion;
```

Assets/Plugins/iOS/Shopify/BuyTests/MessageTests.swift
```diff
@@ -0,0 +1,59 @@
...
+//
+
+import XCTest
+@testable import ProductName
+
+class MessageTests: XCTestCase {
+    
+    let timeout = 10.0
+    
+    override func setUp() {
```
mikkoh: Nit pick but shouldn't this be `setup`

dbart01: Nope, it's defined in `XCTest`. Awkward, I know.

Assets/Plugins/iOS/Shopify/BuyTests/MessageTests.swift
```diff
@@ -0,0 +1,59 @@
...
+        if Tester.hasLoaded == false {
```
sleroux: Nit: Not sure if it effects things but I've usually seen `super.setUp()` called when overriding XCTestCases.

PR #103, opened 2017-06-19
[Feature] iOS Apple Dispatcher Tests

dengjeffrey:
- Tests that the correct Apple Pay messages are successfully sent to managed functions
- Tests that the correct Apple Pay response messages are successfully serialized and sent back


scripts/generator/graphql_generator/csharp/SDK/iOS/SummaryItem.cs.erb
```diff
@@ -0,0 +1,19 @@
+#if UNITY_IPHONE
+namespace <%= namespace %>.SDK.iOS {
+    using System.Collections;
+    using System;
+
+    [Serializable]
+    public class SummaryItem: Serializable {
+        #pragma warning disable 0414
+        public string Label;
```
sleroux: Nit: Are these properties meant to be `readonly`? We could add that specifier to these types of properties to strengthen that.

sleroux: Same for the `ShippingMethod` above

Assets/Plugins/iOS/Shopify/BuyTests/Models/Models.swift
```diff
@@ -116,6 +117,10 @@ struct Models {
             shippingMethod: createShippingMethod())
     }
     
+    static func createSummaryItem() -> PKPaymentSummaryItem {
+        return PKPaymentSummaryItem(label: "SubTotal", amount: NSDecimalNumber.init(value: 1.00))
```
sleroux: Nit: Don't need to explicitly call out `.init` - `NSDecimalNumber(value: 1.00)` works!

scripts/generator/graphql_generator/csharp/SDK/iOS/SummaryItem.cs.erb
dengjeffrey: Since the class is `Serializable`, if we leave them public without readonly, we can utilize `JsonUtility.FromJson<SummaryItem>`. Although we don't use that function right now, it would be confusing if we remove that functionality from a `Serializable`

Assets/UnityTestTools/NativeTesting/Native/MessageCenterTestsHelper.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Shopify.Tests {
+    using UnityEngine;
+    using System.Runtime.InteropServices;
```
mikkoh: What are we using from this namespace?

dengjeffrey: Nice catch!

dengjeffrey:
👍 ?

scripts/generator/graphql_generator/csharp/SDK/iOS/ShippingMethod.cs.erb
```diff
@@ -0,0 +1,20 @@
+#if UNITY_IPHONE
+namespace <%= namespace %>.SDK.iOS {
+    using System.Collections;
+    using System;
+
+    [Serializable]
+    public class ShippingMethod : SummaryItem {
+
+        #pragma warning disable 0414
```
mikkoh: I think you might be able to kill these `warning disable` statements. You are using theser variabes in the constructor.

dengjeffrey: They give off a warning saying that they are set but unused

scripts/generator/graphql_generator/csharp/SDK/iOS/ApplePayEventReceiver.cs.erb
```diff
@@ -0,0 +1,14 @@
+#if UNITY_IPHONE
+namespace <%= namespace %>.SDK.iOS {
+    using System.Collections;
+    using System.Collections.Generic;
+    using UnityEngine;
+
+    public interface ApplePayEventReceiver {
```
mikkoh: Can we name this `IApplePayEventReceiver`?

PR #104, opened 2017-06-19
[Feature] iOS Cart Apple Pay

dengjeffrey:
+ Defines the methods that managed code can access in order to interact with a `PaymentSession` 
+ Adds tests that check arguments are parsed into a `PaymentSession` correctly

##### Decisions
+ The serialized summary items and shipping items are stored in a `[String: String]` with a key of `Items` 
    + The built in serializable in `C#` for `List` must have it as a value of a `Dictionary`, hence the extra wrapping

### DIFF
--- a/Assets/Plugins/iOS/Shopify/BuyTests/CartTests.swift
+++ b/Assets/Plugins/iOS/Shopify/BuyTests/CartTests.swift
@@ -1,0 +1,68 @@
+//
+//  CartTests.swift
+//  UnityBuySDK
+//
+//  Created by Shopify.
+//  Copyright © 2017 Shopify Inc. All rights reserved.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import XCTest
+@testable import ProductName
+
+class CartTests: XCTestCase {
+        
+    func testCreatePaymentSession() {
+        
+        let dataKey        = "Items"
+        let itemLabel      = "SubTotal"
+        let itemAmount     = "1.00"
+        let shippingLabel  = "Free Shipping"
+        let shippingAmount = "0"
+        let identifier     = "FreeShipping"
+        let detail         = "10-15 Days"
+        
+        let summaryItems        = [Models.createSummaryItemJson(amount: itemAmount, label: itemLabel)]
+        let shippingMethods     = [Models.createShippingMethodJson(amount: shippingAmount, label: shippingLabel, identifier: identifier, detail: detail)]
+        let summaryItemData     = try! JSONSerialization.data(withJSONObject: [dataKey: summaryItems])
+        let shippingMethodData  = try! JSONSerialization.data(withJSONObject: [dataKey: shippingMethods])
+        
+        let serializedSummaryItems  = String.init(data: summaryItemData,    encoding: .utf8)!.cString(using: .utf8)
+        let serializedShippingItems = String.init(data: shippingMethodData, encoding: .utf8)!.cString(using: .utf8)
+        
+        let merchantID        = "merchantID".cString(using: .utf8)
+        let countryCode       = "US".cString(using: .utf8)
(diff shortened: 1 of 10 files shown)


Assets/Plugins/iOS/Shopify/BuyTests/CartTests.swift
```diff
@@ -0,0 +1,68 @@
...
+        
+        let merchantID        = "merchantID".cString(using: .utf8)
+        let countryCode       = "US".cString(using: .utf8)
+        let currencyCode      = "USD".cString(using: .utf8)
+        let requiringShipping = true
+        let unityObjectName   = "Tester".cString(using: .utf8)
+
+        _CreateApplePaySession(merchantID, countryCode, currencyCode, requiringShipping, unityObjectName, serializedSummaryItems, serializedShippingItems)
+        
+        XCTAssertEqual(_Session.request.paymentSummaryItems.count, 1)
```
sleroux: Nit: Might be valuable to have a XCTAssertNil check before we create the session just to verify that we're not caching anything.

sleroux: Also if possible we could also nil our the _Session object in the tearDown of the test.

Assets/Plugins/iOS/Shopify/Cart+ApplePay.mm
```diff
@@ -0,0 +1,97 @@
...
+    } else {
+        return false;
+    }
+
+    /// Parse Shipping Items
+    NSDictionary *shippingItemsJson = [NSJSONSerialization JSONObjectWithData:[shippingString dataUsingEncoding:NSUTF8StringEncoding]
+                                                                      options:kNilOptions
+                                                                        error:&error];
+    
+    NSArray *shippingMethods;
```
sleroux: Question: I noticed for `summaryItems` you check to see if the result is nil and return false if so. Is it okay if `shippingMethods` is nil when creating the `PaymentSession`?

sleroux: Related: You're performing close to the same logic when extracting the items from the JSON/string. Does it make sense to pull this out into a helper method? Maybe something like:

```objc
- (NSArray *)itemsFromJSONString:(NSString)jsonString {
}
```

dengjeffrey: + We can create a `PaymentSession` with no `shippingMethods` because we may not have the necessary information to provide shipping information without the user's address

dengjeffrey:
Since at any given point the user can make one payment at a time we can utilize one global PaymentSession. Alternatively we can use opaque pointers and pass a pointer to a session to Unity, but that seems a bit much for what we want to achieve. When the user cancels the state remains. But when the user tries to checkout again, a new session will be created.

Assets/Plugins/iOS/Shopify/Cart+ApplePay.mm
```diff
@@ -0,0 +1,94 @@
...
+bool _CanShowApplePaySetup() {
+    return [PaymentSession canShowSetup];
+}
+
+void _ShowApplePaySetup() {
+    [PaymentSession showSetup];
+}
+
+bool _CreateApplePaySession(const char *merchantID, const char *countryCode, const char *currencyCode, bool requiringShipping, const char *unityDelegateObjectName, const char *serializedSummaryItems, const char *serializedShippingMethods) {
+
```
mikkoh: Should the implementation of these functions be written in Swift and maintain Objective C as just a thin bridge?

dbart01: I agree ^

Assets/Plugins/iOS/Shopify/Cart+ApplePay.h
```diff
@@ -0,0 +1,48 @@
...
+#import <PassKit/PassKit.h>
+#import "Cart+ApplePay.h"
+
+@class PaymentSession;
+PaymentSession *session;
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+    bool _CreateApplePaySession(const char *merchantID, const char *countryCode, const char *currencyCode, bool requiringShipping, const char *unityDelegateObjectName, const char *serializedSummaryItems, const char *serializedShippingMethods);
```
mikkoh: Maybe `unityDelegateObjectName` should be the first parameter passed in. It just feels odd to have that one parameter be in the middle of parameters that describe the payment session.

Assets/Plugins/iOS/Shopify/Cart+ApplePay.mm
```diff
@@ -0,0 +1,94 @@
...
+}
+
+void _ShowApplePaySetup() {
+    [PaymentSession showSetup];
+}
+
+bool _CreateApplePaySession(const char *merchantID, const char *countryCode, const char *currencyCode, bool requiringShipping, const char *unityDelegateObjectName, const char *serializedSummaryItems, const char *serializedShippingMethods) {
+    
+    NSString *itemsString = [NSString stringWithUTF8String:serializedSummaryItems];
+    NSArray *itemJsons = ItemsForString(itemsString);
```
mikkoh: Should this be `itemsJson`

mikkoh: Should these be called `summaryItemsString` and `summaryItemJsons`?

Assets/Plugins/iOS/Shopify/Cart+ApplePay.mm
```diff
@@ -0,0 +1,94 @@
...
+    
+    NSString *shippingString = [NSString stringWithUTF8String:serializedShippingMethods];
+    NSArray *shippingJsons = ItemsForString(shippingString);
+    
+    if (itemJsons == nil) {
+        return false;
```
mikkoh: So basically we fail if we can't deserialize summary items. But what if deserialization of shipping info fails. Should we return a string/error message instead.

dengjeffrey: This call will be internally made, I think it suffices to return just a boolean. A failure on creating a session means there is some underlying error with our SDK

dbart01: If it's an error with the SDK, it should be an assertion. Not a runtime error.

dbart01: If the SDK is responsible for all values being serialized, it can guarantee that deserialization will always succeed. We shouldn't return "won't ever happen" errors just because a framework indicates that errors are a possibility.

Issue #105, opened 2017-06-20
Error Handling


dengjeffrey:
Discussed with @sleroux and @mikkoh, we would like to handler errors in a consistent way throughout the SDK. 

We thought about handling the errors as a list of `ShopifyError` 
```C#
interface ShopifyError { // We leave this implementation empty for now }
delegate void Failure(List<ShopifyError>);

someCall( (errors) => {
    foreach (var error in errors) {
        if (error is httpError) {
        
        } else if (error is GraphQLError) {

        } else if (error is UserError) {

        }
    }
});
```

This way we gain the following:

+ By having `ShopifyError` as an interface we can add new functionality to all errors at any given time
+ Since the `If else` dance is unavoidable by using an array of errors we can return new types of errors without changing thing signature
    
The problems this solution causes:
+ The implementor does not know explicitly what possible errors can be returned, and which ones need to be handled

mikkoh:
I wonder if we could use `Tuples` https://msdn.microsoft.com/en-us/library/system.tuple(v=vs.110).aspx

Tuples def aren't something that Unity devs use though :(

sleroux:
Do you mean to use Tuples as the parameter type for errors instead of an array? Not sure if we can define a Tuple type with a variable number of entries. Would be cool though to wrap the errors into a single tuple though.

dengjeffrey:
@richardmonette Feel free to give some opinions on this solution

sleroux:
﻿As I'm starting to look more into this, I had a couple of additional ideas:

1. One way we could clean this up a bit a more more is to wrap the errors in a class so we don't return an array directly from the API.

2. I wonder if we need strongly typed errors for each error type. I'm probably missing something but for most errors we only need maybe a textual message along with a few other properties.

```csharp
class ShopifyError {
    enum Code {
        httpError, graphqlError, ...
    }

    public readonly Code code;

    // Textual message describing the error
    public readonly string reason;

    // Generic dictionary for putting more specific things in. Inspired by how iOS bubbles up various error types.
    public readonly Dictionary<String, Object> info;

    ...
}

class ShopifyFailure {
    public readonly List<ShopifyError> errors;
    public ShopifyFailure(List<ErrorInfo> errors) {
        this.errors = errors;
    }
}

// ... From the game dev's perspective ...
client.products((products, failure) => {
    if (failure != nil) {
        handleFailure(failure);
    }    
});

void handleFailure(ShopifyFailure failure) {
[...]

PR #107, opened 2017-06-22
Error Handling Refactor

sleroux:
Discussion: https://github.com/Shopify/unity-buy-sdk/issues/105

Instead of returning both a httpError and list of graphQL errors, the API now returns a single ShopifyError which wraps around both those error types. Also, instead
of returning all possible errors at once, we only return the first, most prominent error back to the developer.


scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
```diff
@@ -0,0 +1,17 @@
+namespace <%= namespace %>.SDK {
+    using System.Collections.Generic;
+
+    public class ShopifyError {
+        public enum ErrorType {
```
sleroux: Not sure if `ErrorType` or `Type` is a good name for this.

scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
```diff
@@ -0,0 +1,17 @@
...
+            HTTP, GraphQL
+        }
+
+        public readonly ErrorType Error;
+        public readonly string Reason;
```
sleroux: Would `Message` make more sense than `Reason`?

scripts/generator/graphql_generator/csharp/SDK/ErrorFormatter.cs.erb
```diff
@@ -0,0 +1,22 @@
+namespace <%= namespace %> {
+    using <%= namespace %>.SDK;
+
+    public static class ErrorFormatter {
```
sleroux: I felt like transforming the errors from the response into a ShopifyError doesn't make sense inside the `TopLevelResponse` since generating the `ShopifyError` seems like an abstraction on top of it. Not sure if `ErrorFormatter` makes sense though.

scripts/generator/graphql_generator/csharp/SDK/ConnectionLoader.cs.erb
```diff
@@ -9,7 +9,7 @@ namespace <%= namespace %>.SDK {
     public delegate object GetConnectionFromResponseDelegate(object response);
     public delegate void BuildQueryOnNodeDelegate(QueryRootQuery query, List<ConnectionQueryInfo> connectionInfosToBuildQuery, string nodeId, string alias);
     public delegate void BuildQueryOnEdgesNodeDelegate(object node, string after);
-    public delegate void ResponseNodeHandler(List<Node> nodes, List<string> errors, string httpError);
+    public delegate void ResponseNodeHandler(List<Node> nodes, ShopifyError error);
```
sleroux: I wasn't sure if we should only be generating the `ShopifyError` from the `ShopifyBuy` layer only versus having it here as well.

sleroux:
Left some inline thinking/questions

scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
dengjeffrey: I believe `Type` might be better, so it reads `ShopifyError.Type`, 🤷‍♂️ just my opinion

scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
dengjeffrey: We should probably follow [Error](https://msdn.microsoft.com/en-us/library/t9zk6eay(v=vs.100).aspx), so `ShopifyError.description`

scripts/generator/graphql_generator/csharp/SDK/ErrorFormatter.cs.erb
mikkoh: Maybe a nice thing would be the ability to cast the `TopLevelResponse` to a `ShopifyError`. If there were no issues then `null` is returned. In this case it would be an `explicit` cast and not an `implicit` cast.

scripts/generator/graphql_generator/csharp/SDK/ErrorFormatter.cs.erb
```diff
@@ -0,0 +1,22 @@
...
+        /// <summary>
+        /// Computes the ShopifyError we want to give back to the developer from a response. This method will return the first available error.
+        /// </summary>
+        /// <param name="response">response to format an error from</param>
+        public static ShopifyError FormatErrorFromResponse(TopLevelResponse response) {
```
dengjeffrey: Should this be `ErrorFromResponse`, since this is a non mutating function

scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
mikkoh: I don't mind `ErrorType`.

scripts/generator/graphql_generator/csharp/SDK/ErrorFormatter.cs.erb
```diff
@@ -0,0 +1,22 @@
...
+            if (response.HTTPError != null) {
+                return new ShopifyError(ShopifyError.ErrorType.HTTP, response.HTTPError);
+            } 
+            
+            if (response.errors != null && response.errors.Count > 0) {
+                // Take the first error we received from GraphQL and bubble it up.
+                return new ShopifyError(ShopifyError.ErrorType.GraphQL, response.errors[0]);
```
mikkoh: We can't just pass the first error back. There maybe multiple syntax errors from graphql. I think at a minimum we should just concat all errors with newlines.

I need to think about this more.

scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
mikkoh: I like Jeffreys suggestion.

sleroux:
Updated the property changes, explicit operator cast and a way to pass the GraphQL errors in the string as a JSON array.

scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
```diff
@@ -0,0 +1,31 @@
...
+
+        public readonly ErrorType error;
+        public readonly string description;
+
+        public ShopifyError(ErrorType error, string description) {
+            this.error = error;
+            this.description = description;
+        }
+
+        public static explicit operator ShopifyError(TopLevelResponse response) {
```
mikkoh: Can this `static` method be placed at the top of the file. This is typically what's been done in other files.

```
Public Static Variables
Private Static Variables
Public Static methods
Private Static Methods
Public Property definitions
Private Property definitions
Constructor
Public Methods
Private Methods
```

scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
```diff
@@ -0,0 +1,31 @@
...
+            if (response.HTTPError != null) {
+                return new ShopifyError(ShopifyError.ErrorType.HTTP, response.HTTPError);
+            } 
+            
+            if (response.errors != null && response.errors.Count > 0) {
+                string graphQLErrorsString = Json.Serialize(response.errors);
```
mikkoh: In this PR could you refactor `TopLevelResponse` to simply serialize the `errors` as a string. Basically change.
`private List<string> ParseErrors(Dictionary<string,object> dataJSON) {`
to
`private string ParseErrors(Dictionary<string,object> dataJSON) {`

Assets/Editor/TestResponseDeserialization.cs
```diff
@@ -73,7 +73,7 @@ public class TestResponseDeserialization {
 
             Assert.IsNull(response.DataJSON);
             Assert.IsNotNull(response.errors);
-            Assert.AreEqual("Field 'doesntExist' doesn't exist on type 'Shop'", response.errors[0]);
+            Assert.AreEqual("[\"Field 'doesntExist' doesn't exist on type 'Shop'\"]", response.errors);
```
mikkoh: In this case it really doesn't matter but the returned list would look something like this:

```
 [
    {
      "message": "Field 'doesntExist' doesn't exist on type 'Shop'",
      "locations": [
        {
          "line": 6,
          "column": 5
        }
      ],
      "fields": [
        "query",
        "shop",
        "doesntExist"
      ]
    }
  ]
```

scripts/generator/graphql_generator/csharp/SDK/ConnectionLoader.cs.erb
mikkoh: I think it's good to use it here also.

scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
```diff
@@ -0,0 +1,30 @@
+namespace <%= namespace %>.SDK {
+    using System.Collections.Generic;
+    using MiniJSON;
+
+    public class ShopifyError {
+        public static explicit operator ShopifyError(TopLevelResponse response) {
```
mikkoh: Should we write a test to ensure this cast works properly?

sleroux: Definitely - I'll add one.

Assets/Editor/TestResponseDeserialization.cs
sleroux: Right now `ParseErrors` from `TopLevelResponse` only pulls out the message of the error - not the entire dictionary. Do we want to modify that to take return back the entire GraphQL error dictionary?

mikkoh:
I was just thinking today how are we gonna handle user errors. Which are key value pairs. keys denote fields that were incorrect. values are messages for errors.

sleroux:
What's an example of a user error? Are they the same or similar to GraphQL errors?

I also mentioned above that currently when parsing the errors from a response, we're only storing a list of the GraphQL messages - not the entire dictionary object. Should we be stringifying the entire error object?

It's starting to sound like the structure of each error is quite different from each other and contains information we want to surface. Would moving back to the subclassing model instead of the enum make more sense for this?

mikkoh:
I think you're handling GraphQL errors correctly. 

Hmmm. It would feel bad to have to serialize user errors back to a string and then to have a subclass deserialize it for ease of use.

If you want to see the form of a `UserError` checkout the generated file `UserError.cs`.

```
field: [String!]
Path to input field which caused the error.

message: String!
The error message.
```

I'm sort of not clear what `field` will look like when there's an error. Cause technically it should point to an input object.

sleroux:
I'm going to go ahead and land this so we can follow up with the UserError and other API adjustment work.

PR #108, opened 2017-06-22
Link with UnityEngine.dll for testing

sleroux:
Since NUnit is obsolete for newer versions of Mono, the developer would need to install an older version to run the ./scripts/test.sh script properly. This patch uses the
Mono installation bundled with Unity which is older (4.0.5.0).

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -3,8 +3,10 @@
 which mcs &> /dev/null || die "Mono is not installed"
 which nunit-console &> /dev/null || die "nunit-console is not installed"
 
+UNITY_MONO_DIR="/Applications/Unity/MonoDevelop.app/Contents/Frameworks/Mono.framework/Versions/Current"
+
 # we're defining SHOPIFY_TEST here so that source can check whether we're running assertions/tests
-mcs -debug -define:SHOPIFY_TEST -define:SHOPIFY_MONO_UNIT_TEST -recurse:'Assets/Shopify/*.cs' -recurse:'Assets/Editor/*.cs' -reference:nunit.framework.dll -target:library -out:test.dll
+"$UNITY_MONO_DIR/bin/mcs" -debug -define:SHOPIFY_TEST -define:SHOPIFY_MONO_UNIT_TEST -recurse:'Assets/Shopify/*.cs' -recurse:'Assets/Editor/*.cs' -reference:nunit.framework.dll -target:library -out:test.dll
 
 if [ $? = 0 ] ; then
     nunit-console test.dll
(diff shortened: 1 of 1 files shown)


scripts/test.sh
```diff
@@ -3,8 +3,18 @@
...
-mcs -debug -define:SHOPIFY_TEST -define:SHOPIFY_MONO_UNIT_TEST -recurse:'Assets/Shopify/*.cs' -recurse:'Assets/Editor/*.cs' -reference:nunit.framework.dll -target:library -out:test.dll
+"$UNITY_MONO_DIR/bin/mcs" \
+    -debug \
+    -define:SHOPIFY_TEST \
+    -define:SHOPIFY_MONO_UNIT_TEST \
+    -recurse:'Assets/Shopify/*.cs' \
+    -recurse:'Assets/Editor/*.cs' \
+    -reference:nunit.framework.dll \
+    -reference:'/Applications/Unity/Unity.app/Contents/Managed/UnityEngine.dll' \
+    -target:library -out:test.dll
```
dengjeffrey:
```
...
-target:library \
-out
```

mikkoh:
😕
screen shot 2017-06-22 at 3 51 34 pm

sleroux:
Strangely you have to run MonoDevelop at least once for it to generate that /tmp/ file 😔 I wonder if there is a way around that

scripts/test.sh
```diff
@@ -3,8 +3,19 @@
 which mcs &> /dev/null || die "Mono is not installed"
```
mikkoh: This line should be changed to use the path to `mcs`

sleroux:
I don't think there is a way around the thing MonoDevelop does when it is opened that generates that /tmp folder. I've reverted back to use the system mcs installation but kept the UnityEngine.dll reference to resolve testing errors we were getting from NativeMessage.cs linking to Unity.

sleroux:
I noticed that for classes which link to UnityEngine.dll we wrap in SHOPIFY_MONO_TESTS. Is that the best approach instead of linking in the engine? If so we can just close this PR.

dengjeffrey:
I think we should link to UnityEngine.dll, this allows us to test anything that might refer to one of the classes located in that namespace. Probably should wait for @mikkoh's thoughts though

mikkoh:
One of the reasons I've tried to keep a separation between the library and Unity is that there might be an offshoot chance that in the future we'll need to support another C# library. Up until this point it's been pretty easy to do that. The reason why I think this is important is that I've been seeing other game engines which are using C#. Maybe it's overkill but at the same time I think it's pretty easy to separate the files that are Unity specific.

sleroux:
Makes sense and I think anything that is Unity-dependent is something we can wrap in some kind of interface as well. Closing this down.

PR #110, opened 2017-06-22
Move Delegates to Single File

sleroux:
Moves all of the delegates we use into a single SDK/Delegates.cs.erb file. Also changes the only Action<> we used for CartLineItems to be a delegate instead.

### DIFF
--- a/scripts/generator/graphql_generator/csharp.rb
+++ b/scripts/generator/graphql_generator/csharp.rb
@@ -102,6 +102,7 @@ module GraphQLGenerator
         SDK/DefaultQueriesProducts
         SDK/DefaultQueriesCollections
         SDK/DefaultQueriesCheckout
+        SDK/Delegates
         SDK/GlobalGameObject
         SDK/QueryLoader
         SDK/ConnectionLoader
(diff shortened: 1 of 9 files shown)


sleroux:
Is this what you were thinking of doing @mikkoh? We only currently use Actions in two places: OnChange for CartLineItems and the Unity timeout script but I've gone and changed the first one to be a delegate instead. 

Just WIP for now - I'll go back and clean up comments and remove the testing commits I pulled in from the other branch.

mikkoh:
I think this is headed in the right direction for sure. Thoughts @dengjeffrey? At some point I'd like to review that there are no redundant delegates.

dengjeffrey:
This looks good, however I think in a separate PR we should rename these delegate methods. 
Now without the context of the class, as a developer who is looking possible to re-use some of these delegate it may be really hard to determine which each delegate is for. 

Possible could follow: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members or https://stackoverflow.com/questions/2346065/proper-naming-convention-for-a-net-delegate-type

PR #111, opened 2017-06-22
[Feature] Checkouts Shipping

dengjeffrey:
#### What
+ Adds default queries for shipping and email mutations
+ Adds functions in cart that makes those mutations
+ Adds extra fields to the default checkout query such as subtotal, tax, and total
     + Prepares for Apple Pay

#### What this doesn't include
+ Test for shipping functions in cart

#### Decisions
+ Decided to make the mutations on shipping and email info private
    + As of right now our checkout experience does not allow you to pre-add shipping and email information


dengjeffrey:
Will split this into multiple PRs

PR #112, opened 2017-06-22
[Feature] Ready Poll

dengjeffrey:
+ Puts the polling functionality for a checkout into separate method

#### Motivation
+ By moving the polling function out, we can now poll for any `Ready` fields on `Checkout` specifically `ShippingLines`


mikkoh:
As far as I remember. `ready` will only be set to `true` once everything is good to go. I'm not sure we need to check anything else than `ready`

Issue #106, opened 2017-06-21
Unity native/web interface for checking out


sleroux:
**Problem**

Jeffrey and I spent some time today talking about how we want to handle the checkout integration into the Unity side of the project. Initially we were thinking
of having the checkout logic live in `Cart.cs` but when thinking ahead to the web/Android implementations we some some short comings:

1. When the game developer checks out they pass in callback actions which would be stored on `Cart.cs`. This adds more responsibility to the `Cart.cs` class and adds 
more potentially platform-specific state.

2. The `Cart.cs` would contain both logic for iOS and Android checkout logics which lengthens the implementation and adds more sprinkling of #ifdefs.

**Solution**

To break up the concerns of the Cart and Checkout methods, 

1. Extract checkout logic from Cart.cs and forward API calls to an implementor.
2. Split up the checkout implementation across platform lines.

To accomplish this, we define a new interface for handling checking out:

```csharp
interface ICartCheckout {
    void StartWebCheckout(Action success, Action failure);
    bool CanShowNativePaySetup();
    void ShowNativePaySetup();
    void PrepareNativeCheckout();
    void CompleteNativeCheckout(Action success, CheckoutFailure failure);
}
```

This will allow us to define different implementations for the various platforms as well has having a single place to store state related to checking out:

```csharp
// iOS
partial class IOSCartCheckout: ICartCheckout {
    // Platform specific plugin methods.
    [DllImport ("__internal")]
    private static extern void _StartWebCheckout
    
    // Storage for callbacks, etc.
    private Action success, failure;

    // Implementation of checkout methods that previously would have been in Cart.cs.
    void StartWebCheckout(Action success, Action failure) {}
        // Setup global game object with receiver
        this.success = success;
        this.failure = faiure;
        _StartWebCheckout(...);
    }
}
```

We can also extend this class to implement the different interfaces for the native -> Unity messages:

```csharp
partial class IOSCartCheckout: ApplePayEventsReceiver {

}

partial class IOSCartCheckout: WebPayEventsReceiver {
    void DidCompleteWebCheckout(string result) {
        result ? success() : failure();
    }
}
```

With the implementations contained in the ICartCheckout objects, we can now create a platform-specific instance when the Cart is created:

```csharp
// Checkout methods and implementation
class Cart  {
    private ICartCheckout cartCheckout;

    public Cart() {
        // Load in platform-specific implementor. Pass in cart for mutations if needed.
    #if UNITY_IOS
[...]

PR #114, opened 2017-06-23
[Unity] Add IWebPayable with Web Checkout API

sleroux:
First pass at implementing an interface and the basics of a web checkout implementor for iOS.

### DIFF
--- a/scripts/generator/graphql_generator/csharp/Cart.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/Cart.cs.erb
@@ -5,6 +5,10 @@ namespace <%= namespace %> {
     using <%= namespace %>.SDK;
     using <%= namespace %>.GraphQL;
 
+    #if UNITY_IPHONE
+    using <%= namespace %>.SDK.iOS;
+    #endif
+
     /// <summary>
     /// This exception is thrown when interacting with a cart to add, update, or delete line items and no matching
     /// variant could be found.
@@ -12,7 +16,6 @@ namespace <%= namespace %> {
     public class NoMatchingVariantException : Exception {
         public NoMatchingVariantException(string message) : base(message) {}
     }
-
     /// <summary>
     /// Manages line items for an order. Can also be used to generate a web checkout link to check out in browser.
     /// </summary>
@@ -44,6 +47,8 @@ namespace <%= namespace %> {
         private ShopifyClient Client;
         private Checkout CurrentCheckout = null;
 
+        private IWebCheckout WebCheckout;
+
         /// <summary>
         /// Constructs a new cart using a <see ref="ShopifyClient">ShopifyClient </see>. Typically, carts won't be
         /// instantiated directly, but will rather be instatiated using <see ref="ShopifyClient.Cart">ShopifyBuy.Client().Cart() </see>.
@@ -61,6 +66,10 @@ namespace <%= namespace %> {
         public Cart(ShopifyClient client) {
             Client = client;
             _LineItems = new CartLineItems(OnDeleteLineItem);
+
+            #if UNITY_IOS
+            WebCheckout = new iOSWebCheckout();
+            #endif
         }
 
         public delegate void GetWebCheckoutLinkSuccess(string url);
@@ -74,6 +83,14 @@ namespace <%= namespace %> {
                 }
 
                 success(CurrentCheckout.webUrl());
+            });
+        }
+
+        public void CheckoutWithNativeWebView(CheckoutSuccessCallback success, CheckoutFailureCallback failure, CheckoutCancelCallback cancelled) {
+            GetWebCheckoutLink(url => {
+                WebCheckout.Checkout(url, success, failure, cancelled);
+            }, error => {
+                failure(error);
             });
         }
 
--- a/scripts/generator/graphql_generator/csharp/SDK/GlobalGameObject.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/GlobalGameObject.cs.erb
@@ -5,6 +5,16 @@ namespace <%= namespace %>.SDK {
     public class GlobalGameObject {
         private static GameObject GameObject;
 
+        public static string Name {
+            get {
+                if (GameObject != null) {
+                    return GameObject.name;
+                } else {
+                    return null;
+                }
+            }
+        }
+
         public static T AddComponent<T>() where T: Component {
             if (GameObject == null) {
                 GameObject = new GameObject("Shopify");
--- a/scripts/generator/graphql_generator/csharp/SDK/IWebCheckout.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/IWebCheckout.cs.erb
@@ -1,0 +1,7 @@
+namespace <%= namespace %>.SDK {
+    using System;
+    
+    interface IWebCheckout {
+        void Checkout(string checkoutURL, CheckoutSuccessCallback success, CheckoutFailureCallback failure, CheckoutCancelCallback cancelled);
+    }
+}
--- a/scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
@@ -1,0 +1,46 @@
+#if UNITY_IPHONE
+namespace <%= namespace %>.SDK.iOS {
+    using UnityEngine;
+    using System.Runtime.InteropServices;
+    using <%= namespace %>.SDK;
+
+    class iOSWebCheckout: IWebCheckout {
+        private static iOSWebViewMessageReceiver _messageReceiver;
+
+        [DllImport ("__Internal")]
+        private static extern void _CheckoutWithWebView(string unityDelegateObjectName, string checkoutURL);
+
+        public void Checkout(string checkoutURL, CheckoutSuccessCallback success, CheckoutFailureCallback failure, CheckoutCancelCallback cancelled) {
+            SetupNativeMessageReceiver(success, failure, cancelled);
+            _CheckoutWithWebView(GlobalGameObject.Name, checkoutURL);
+        }
+
+        private static void SetupNativeMessageReceiver(CheckoutSuccessCallback success, CheckoutFailureCallback failure, CheckoutCancelCallback cancelled) {
+            if (_messageReceiver == null) {
+                _messageReceiver = GlobalGameObject.AddComponent<iOSWebViewMessageReceiver>();
+            }
+        
+            _messageReceiver.OnSuccess = success;
+            _messageReceiver.OnFailure = failure;
+            _messageReceiver.OnCancelled = cancelled;
+        }
+    }
+
+    class iOSWebViewMessageReceiver: MonoBehaviour {
+        internal CheckoutSuccessCallback OnSuccess;
+        internal CheckoutFailureCallback OnFailure;
+        internal CheckoutCancelCallback OnCancelled;
+
+        void OnNativeMessage(string jsonMessage) {
+            var message = NativeMessage.CreateFromJSON(jsonMessage);
+            if (message.Content == "loadedThankYou") {
+                // Do a round trip to the server to see if the purchase is solid.
+                // Invoke callbacks on cart to pass to Unity
+                OnSuccess();
+            } else if (message.Content == "cancelled") {
+                OnCancelled();
+            }
+        }
+    }
+}
+#endif
--- a/scripts/generator/graphql_generator/csharp.rb
+++ b/scripts/generator/graphql_generator/csharp.rb
@@ -120,10 +120,12 @@ module GraphQLGenerator
         SDK/NativeMessage
         SDK/Serializable
         SDK/INativeCheckout
+        SDK/IWebCheckout
         SDK/iOS/ApplePayAuthorizationStatus
         SDK/iOS/ApplePayEventReceiver
         SDK/iOS/ApplePayEventResponse
         SDK/iOS/ShippingMethod
+        SDK/iOS/iOSWebCheckout
         SDK/iOS/SummaryItem
         SDK/iOS/Cart.ApplePayEventReceiver
         SDK/iOS/iOSNativeCheckout
--- a/scripts/generator/graphql_generator/csharp/SDK/Delegates.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/Delegates.cs.erb
@@ -25,4 +25,8 @@ namespace <%= namespace %>.SDK {
     public delegate void OnCartLineItemChange();
 
     public delegate bool PollUpdatedHandler(QueryRoot updatedQueryRoot);
+
+    public delegate void CheckoutSuccessCallback();
+    public delegate void CheckoutFailureCallback(ShopifyError error);
+    public delegate void CheckoutCancelCallback();
 }


sleroux:
@dengjeffrey Pushed up some code so we can align on how on the interface for the different checkouts. I was thinking that when you've implemented the C# bits of Apply pay we can move that code out of Cart.cs and into another partial definition for `iOSCartCheckout` defined in `iOSCartCheckout.Native.cs.erb` while forwarding the calls from Cart to it. I just put in `Action`s for the `ICartCheckout` stuff as a place holder but we'll update that with the API you're adding to Cart.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -47,6 +54,13 @@ namespace <%= namespace %> {
         public Cart(ShopifyClient client) {
             Client = client;
             _LineItems = new CartLineItems();
+
+            // TODO: When we add more platform we can move this into a factory.
+            #if UNITY_IOS
+            checkout = new iOSCartCheckout();
```
dengjeffrey: These `Checkout` classes are going to need a reference to `Cart`, so it can make mutations on the cart. 
Or they need to set the `Cart` as a delegate to checkout events

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSCartCheckout.Web.cs.erb
```diff
@@ -0,0 +1,47 @@
...
+
+    partial class iOSWebPayCheckout {
+        internal WebCheckoutSucceeded success;
+        internal WebCheckoutFailed failure;
+
+        void DidCompleteCheckout(bool didSucceed) {
+            if (didSucceed) {
+                success();
+            } else {
+                failure(null);
```
dengjeffrey: Might need some more info here. Was it an HTTP error? or did the user cancel

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSCartCheckout.Web.cs.erb
```diff
@@ -0,0 +1,47 @@
...
+            }
+        }
+    }
+
+    #if !SHOPIFY_MONO_UNIT_TEST
+    partial class iOSWebPayCheckout: MonoBehaviour {}
```
dengjeffrey: Nice!

sleroux:
@dengjeffrey @JakeCataford @mikkoh 

I've updated this PR to include discussion points from https://github.com/Shopify/unity-buy-sdk/issues/106 as well as a functioning web pay backend for iOS. This PR paired up with iOS native implementation https://github.com/Shopify/unity-buy-sdk/pull/120 rounds out the feature end-to-end.

Changes:

- `ICartCheckout` is broken out into `INativePayable` and `IWebPayable` interfaces
- Web checkout API has been updated to include 2 callbacks: `onUIStateChanged` and `didCompletePurchase`. 

The new `onUIStateChanged` callback will allow game developers to respond to the UI state of the web checkout. For example, if the game needs to be paused before we go into the web checkout process, the developer can do so by handline the `willPresent` state.

`didCompletePurchase` allows a developer to respond to the end of the checkout process. It's worth noting that this is invoked when the user is taken to the `thank_you` checkout page. A user could potentially spoof this so it might be better to rename this to `didCompleteCheckout` to make it more ambigious...

sleroux:
Forgot to mention, I'm going to look into some ways to test this from both the Unity and iOS side in follow up PRs.

scripts/generator/graphql_generator/csharp/SDK/IWebPayable.cs.erb
```diff
@@ -0,0 +1,14 @@
+namespace <%= namespace %>.SDK {
+    using System;
+    
+    interface IWebPayable {
+        void StartWebCheckout(string checkoutURL, WebCheckoutUIStateChanged uiStateChanged, WebCheckoutDidCompletePurchase completedPurchase);
+    }
+
+    public enum WebCheckoutUIState {
+        willPresent, presenting, willDismiss, dismissed, invalid
```
dengjeffrey: Nit: C# Enums are preferred to be in PascalCase

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebPayable.cs.erb
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public void DidCompletePurchase() {
+            completedPurchase();
+        }
+
+        private static WebCheckoutUIState StateFromString(string state) {
+            if (state == "willPresent") {
+                return WebCheckoutUIState.willPresent;
+            } else if (state == "presenting") {
```
dengjeffrey: Nit: Might be better to use:
```
state == WebCheckoutUIState.presenting.ToString("G")
```
To avoid making spelling mistakes

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebPayable.cs.erb
```diff
@@ -0,0 +1,54 @@
...
+                webComponent = GlobalGameObject.AddComponent<iOSWebPayBehaviour>();
+            }
+            webComponent.uiStateChanged = uiStateChanged;
+            webComponent.completedPurchase = completedPurchase;
+            _StartWebCheckout(GlobalGameObject.GameObjectName, checkoutURL);
+        }
+    }
+
+    internal class iOSWebPayBehaviour: MonoBehaviour {
+        internal WebCheckoutUIStateChanged uiStateChanged;
```
dengjeffrey: Nit: C# instance variables should be in PascalCase

scripts/generator/graphql_generator/csharp/SDK/Serializable.cs.erb
```diff
@@ -1,3 +1,4 @@
+#if !SHOPIFY_MONO_UNIT_TEST
 namespace <%= namespace %>.SDK {
 #if !SHOPIFY_MONO_UNIT_TEST
```
dengjeffrey: Probably should remove this

scripts/generator/graphql_generator/csharp/SDK/GlobalGameObject.cs.erb
```diff
@@ -5,6 +5,12 @@ namespace <%= namespace %>.SDK {
     public class GlobalGameObject {
         private static GameObject GameObject;
 
+        public static string GameObjectName {
+            get {
+                return GameObject.name;
```
dengjeffrey: This might blow up when `GameObject` is still null

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidCartCheckout.Web.cs.erb
```diff
@@ -0,0 +1,12 @@
+namespace <%= namespace %>.SDK.Android {
+#if !SHOPIFY_MONO_UNIT_TEST
```
dengjeffrey: `#if !SHOPIFY_MONO_UNIT_TEST` Can probably go above the namespace

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -13,6 +15,9 @@ namespace <%= namespace %> {
         public NoMatchingVariantException(string message) : base(message) {}
     }
 
+    public delegate void WebCheckoutSucceeded();
+    public delegate void WebCheckoutFailed(object error);
```
dengjeffrey: Should these be in `Delegates.cs`?

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
```diff
@@ -0,0 +1,36 @@
...
+
+        public void StartWebCheckout(string checkoutURL) {
+            if (webComponent == null) {
+                webComponent = GlobalGameObject.AddComponent<iOSWebCheckoutBehaviour>();
+            }
+            _StartWebCheckout(GlobalGameObject.GameObjectName, checkoutURL);
+        }
+    }
+
+    internal class iOSWebCheckoutBehaviour: MonoBehaviour {
```
dengjeffrey: I believe the default is `internal` already

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
```diff
@@ -0,0 +1,42 @@
...
+        internal Cart CurrentCart; 
+
+        internal void OnNativeMessage(string jsonMessage) {
+            var message = NativeMessage.CreateFromJSON(jsonMessage);
+            if (message.Content == "loadedThankYou") {
+                // Do a round trip to the server to see if the purchase is solid.
+                // Invoke callbacks on cart to pass to Unity
+                CurrentCart.onCheckoutSucceeded.Invoke();
```
richardmonette: I'm still wondering if maybe this should be two levels of events, so instead of reaching into Cart, iOSWebCheckout also have events, which Cart registers from above, and then re-broadcasts as it sees fit.

sleroux: Yea I agree - it would be cleaner to do so. We could pass in delegates into iOSWebCheckout prior to checking out like so:

```csharp
// (in Cart.cs)
void StartWebCheckout() {
    webCheckout.OnCheckoutSucceeded = () => {
        // Call external calback
    };
    webCheckout.OnCheckoutFailed = () => {
        ....
    };
    webCheckout.Start(url);
}
```

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
```diff
@@ -0,0 +1,43 @@
+namespace <%= namespace %>.SDK.iOS {
+#if !SHOPIFY_MONO_UNIT_TEST
```
dengjeffrey: Nit: Should be the first line

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
```diff
@@ -0,0 +1,43 @@
...
+    using UnityEngine;
+    using System.Runtime.InteropServices;
+    using <%= namespace %>.SDK;
+
+    class iOSWebCheckout: IWebCheckout {
+        private static iOSWebCheckoutBehaviour webComponent;
```
dengjeffrey: `WebComponent`

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
```diff
@@ -0,0 +1,43 @@
...
+            webComponent.OnCancelled = cancelled;
+
+            _StartWebCheckout(GlobalGameObject.GameObjectName, checkoutURL);
+        }
+    }
+
+    class iOSWebCheckoutBehaviour: MonoBehaviour {
+        internal CheckoutSuccessCallback OnSuccess;
+        internal CheckoutFailureCallback OnFailure;
+        internal CheckoutCancelCallback OnCancelled;
```
dengjeffrey: Are these always going to be set? If so should these be part of the constructor?

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
sleroux: If I put it outside the project won't compile because there is no other place in the project where we declare the SDK.iOS namespace. Should I define an empty namespace somewhere else or leave the check inside?

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
sleroux: Since we're creating this behaviour through `AddComponent` we won't be able to pass the values into the constructor. I can probably wrap this in a small private helper method though to clean it up.

dengjeffrey: Ahh gotcha 👍

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
dengjeffrey: Should probably wrap this in `#if UNITY_IPHONE`, as well as where you end up using this in `Cart`

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -77,6 +86,14 @@ namespace <%= namespace %> {
             });
         }
 
+        public void StartWebCheckout(CheckoutSuccessCallback success, CheckoutFailureCallback failure, CheckoutCancelCallback cancelled) {
```
mikkoh: I feel like this method should be called something like `CheckoutWithNativeWebView` or if we want to be inline with naming with Apple Pay checkout `CompleteNativeWebViewCheckout`.

This being called `Start*` I feel like there should be a `FinishWebCheckout` or `FinalizeWebCheckout`.

scripts/generator/graphql_generator/csharp/SDK/GlobalGameObject.cs.erb
```diff
@@ -5,6 +5,16 @@ namespace <%= namespace %>.SDK {
     public class GlobalGameObject {
         private static GameObject GameObject;
 
+        public static string GameObjectName {
```
mikkoh: Nit. Maybe this should be simply called `name` since `AddComponent` is actually a method of a `GameObject`.

scripts/generator/graphql_generator/csharp/SDK/IWebCheckout.cs.erb
```diff
@@ -0,0 +1,7 @@
+namespace <%= namespace %>.SDK {
+    using System;
+    
+    interface IWebCheckout {
+        void Start(string checkoutURL, CheckoutSuccessCallback success, CheckoutFailureCallback failure, CheckoutCancelCallback cancelled);
```
mikkoh: Again when I see a method named `Start` I expect there to be a matching `Finish` or `Finalize` method.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
```diff
@@ -0,0 +1,51 @@
...
+        private static extern void _StartWebCheckout(string unityDelegateObjectName, string checkoutURL);
+
+        public void Start(string checkoutURL, CheckoutSuccessCallback success, CheckoutFailureCallback failure, CheckoutCancelCallback cancelled) {
+            SetupWebCheckoutComponent(success, failure, cancelled);
+
+            _StartWebCheckout(GlobalGameObject.GameObjectName, checkoutURL);
+        }
+
+        private static void SetupWebCheckoutComponent(CheckoutSuccessCallback success, CheckoutFailureCallback failure, 
+                CheckoutCancelCallback cancelled) {
```
mikkoh: Just keep it on the same line. Or drop all the params on different lines.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
```diff
@@ -0,0 +1,51 @@
...
+        internal CheckoutFailureCallback OnFailure;
+        internal CheckoutCancelCallback OnCancelled;
+
+        void OnNativeMessage(string jsonMessage) {
+            var message = NativeMessage.CreateFromJSON(jsonMessage);
+            if (message.Content == "loadedThankYou") {
+                // Do a round trip to the server to see if the purchase is solid.
+                // Invoke callbacks on cart to pass to Unity
+                OnSuccess();
+            } else if (message.Content == "cancelled") {
```
mikkoh: What if there's no internet connection? Can we get a failure here?

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
```diff
@@ -0,0 +1,51 @@
+#if UNITY_IPHONE
+namespace <%= namespace %>.SDK.iOS {
+#if !SHOPIFY_MONO_UNIT_TEST
+    using UnityEngine;
+    using System.Runtime.InteropServices;
+    using <%= namespace %>.SDK;
+
+    class iOSWebCheckout: IWebCheckout {
+        private static iOSWebCheckoutBehaviour WebComponent;
```
mikkoh: I feel that `iOSWebCheckoutBehaviour` isn't a really descriptive name. Really what does `iOSWebCheckoutBehaviour` do? It receives messages from the native module. Should we call `iOSWebCheckoutBehaviour` something like `iOSWebviewMessageReceiver`. I'm not sure the names should ever include `Behaviour` or `Component` because in Unity it's essentially noise because everything is either a `MonoBehaviour` or an instance / `Component` of a `MonoBehaviour`.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
sleroux: I like `CheckoutWithNativeWebView` over `CompleteNativeWebViewCheckout` mostly because we beginning of the checkout process.

scripts/generator/graphql_generator/csharp/SDK/IWebCheckout.cs.erb
sleroux: Only other name I can think of is `Checkout(...)`. Would that work?

sleroux:
Updated with clarified naming for API and other methods/classes.

mikkoh:
Just to be clear are you going to handle HTTP errors in another PR or this one? Cause I can re-review if needed keeping a second PR in mind.

mikkoh:
I approved this since we'll be adding http error handling in another pr.

sleroux:
I've opened an issue to earmark the HTTP error handling https://github.com/Shopify/unity-buy-sdk/issues/136

sleroux:
@dengjeffrey 👍?

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
dengjeffrey: Do we need `SHOPIFY_MONO_UNIT_TEST ` now that we have `UNITY_IPHONE `

PR #117, opened 2017-06-26
[Feature] Checkouts Shipping Mutations

dengjeffrey:
- Adds functionality to the Cart to set the Shipping and Email for the Checkout
- Adds Polling for Available shipping rates

### DIFF
--- a/scripts/generator/graphql_generator/csharp/Cart.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/Cart.cs.erb
@@ -151,6 +151,51 @@ namespace <%= namespace %> {
             } catch(NoQueryException exception) {}
         }
 
+        private void SetShippingAddress(MailingAddressInput mailingAddressInput, Action<List<string>, string> callback) {
+            MutationQuery query = new MutationQuery();
+
+            DefaultQueries.checkout.ShippingAddressUpdate(query, CurrentCheckout.id(), mailingAddressInput);
+
+            Client.Mutation(query, (Mutation response, List<string> errors, string httpError) => {
+                if (errors != null || httpError != null) {
+                    callback(errors, httpError);
+                } else {
+                    UpdateCheckout(response.checkoutShippingAddressUpdate().checkout());
+                    AvailableShippingRatesPoll(callback);
+                }
+            });
+        }
+
+        private void SetShippingLine(string shippingRateHandle, Action<List<string>, string> callback) {
+            MutationQuery query = new MutationQuery();
+
+            DefaultQueries.checkout.ShippingLineUpdate(query, CurrentCheckout.id(), shippingRateHandle);
+
+            Client.Mutation(query, (Mutation response, List<string> errors, string httpError) => {
+                if (errors != null || httpError != null) {
+                    callback(errors, httpError);
+                } else {
+                    UpdateCheckout(response.checkoutShippingLineUpdate().checkout());
+                    CheckoutPoll(callback);
+                }
+            });
+        }
+
+        private void SetEmail(string email, Action<List<string>, string> callback) {
+            MutationQuery query = new MutationQuery();
+
+            DefaultQueries.checkout.EmailUpdate(query, CurrentCheckout.id(), email);
+
+            Client.Mutation(query, (Mutation response, List<string> errors, string httpError) => {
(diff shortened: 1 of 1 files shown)


richardmonette:
The body of these methods looks pretty similar, since we are using ERB anyways, would this be a good opportunity to have a helper which writes these methods? (then you could just call it three times with the right args for the few differences)

sleroux:
I like the idea of using the ERB stuff to not only generate the SDK files but methods for those files. To do this, would we need to expand on the `csharp.rb` script? I wonder if adding this kind of generation will make it more complicated. The other option is to find a way a to DRY the methods up but I'm not sure if that's an option here since the methods are similar but different in what they invoke.

I'm down for 👍 this for now and revisiting the method-level templating in a refactoring discussion.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -151,6 +151,51 @@ namespace <%= namespace %> {
...
+                if (errors != null || httpError != null) {
+                    callback(errors, httpError);
+                } else {
+                    UpdateCheckout(response.checkoutShippingAddressUpdate().checkout());
+                    AvailableShippingRatesPoll(callback);
+                }
+            });
+        }
+
+        private void SetShippingLine(string shippingRateHandle, Action<List<string>, string> callback) {
```
mikkoh: Was the decision that we'd be using `delegates` here on in everywhere for consistency and readability?

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -151,6 +151,51 @@ namespace <%= namespace %> {
             } catch(NoQueryException exception) {}
         }
 
+        private void SetShippingAddress(MailingAddressInput mailingAddressInput, Action<List<string>, string> callback) {
```
mikkoh: Is the plan that the native module will send a JSON object via a message that can be used to create a `MailingAddressInput`?

mikkoh:
I'm not sure it makes sense to generate the code here. Might add complexity without much benefit. There could be a huge benefit in generating default queries though.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -192,6 +237,15 @@ namespace <%= namespace %> {
             );
         }
 
+        private void AvailableShippingRatesPoll(Action<List<string>, string> callback) {
+            PollTillReady(
+                checkout => {
+                    return checkout.availableShippingRates().ready();
+                },
+                callback
```
mikkoh: Should you update `CurrentCheckout` before calling the `callback` otherwise `CurrentCheckout` will still have `availableShippingRates().ready() == false` even though it is good to go.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
dengjeffrey: Yes but it seems to be a refactoring effort. I think it would be better if it went into https://github.com/Shopify/unity-buy-sdk/pull/110. So we don't repeat ourselves

scripts/generator/graphql_generator/csharp/Cart.cs.erb
dengjeffrey: yes

scripts/generator/graphql_generator/csharp/Cart.cs.erb
dengjeffrey: `PollTillReady` will handle updating the checkout. Here we are passing the `callback` to `PollTillReady`

scripts/generator/graphql_generator/csharp/Cart.cs.erb
dengjeffrey: Actually since this is private, I'll make the change here

dengjeffrey: Okay so I realized, there are other methods in `Cart` that uses these actions. We should do this refactor work in another PR

mikkoh: 👍

PR #119, opened 2017-06-26
[Feature] Input JSON Constructor

dengjeffrey:
- Adds a constructor for Input types to be created from a JSON dictionary. 

#### Motivation
- By adding this constructor, Input types such as `MailingAddressInput` can be constructed from the serialized json objects sent from unmanaged functions

#### Decisions
- Decided to re-throw any exceptions. But at the call site we would probably not use a try and catch since a failure to initialize meant there is an error with our SDK

### DIFF
--- a/scripts/generator/graphql_generator/csharp/type.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/type.cs.erb
@@ -146,6 +146,27 @@ namespace <%= namespace %>.GraphQL {
                     }
                 <% end %>
             }
+
+            <%= params_doc(type.input_fields) %>
+            public <%= type.classify_name %>(Dictionary<string, object> dataJSON) {
+                <%# handle adding required fields %>
+                <% if type.required_input_fields.length > 0 %>
+                try {
+                    <% type.required_input_fields.each do |field| %>
+                        Set("<%= field.name %>", dataJSON["<%= escape_reserved_word(field.name) %>"]);
+                    <% end %>
+                } catch {
+                    throw;
+                }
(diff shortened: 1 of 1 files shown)


scripts/generator/graphql_generator/csharp/type.cs.erb
```diff
@@ -146,6 +146,27 @@ namespace <%= namespace %>.GraphQL {
...
+            <%= params_doc(type.input_fields) %>
+            public <%= type.classify_name %>(Dictionary<string, object> dataJSON) {
+                <%# handle adding required fields %>
+                <% if type.required_input_fields.length > 0 %>
+                try {
+                    <% type.required_input_fields.each do |field| %>
+                        Set("<%= field.name %>", dataJSON["<%= escape_reserved_word(field.name) %>"]);
+                    <% end %>
+                } catch {
+                    throw;
```
mikkoh: The only thing I wonder is whether we should throw something meaningful. Create our own exception.

dengjeffrey: In this case it would throw a key not found exception, we should be able to infer from the call stack that the we passed in a dataJSON with incorrect keys

richardmonette: Could be more nicely phrased though, with perhaps some basic info on how to fix.

dengjeffrey: Sure, I'll create a ticket for this issue and come back to this later

PR #120, opened 2017-06-26
[iOS] iOS implementation for Web Checkouts

sleroux:
This patch implements the _StartWebCheckout(..) API from the Unity side on iOS. From a high level,

* The Unity code calls into the native plugin via `_StartWebCheckout(unityDelegateName, checkoutURL)`.
* The API creates a `WebCheckoutSession` which encapsulates the web checkout.
* `WebCheckoutSession` creates the `WebCheckoutViewController` which contains all of the custom UI for checking out.
* `WebCheckoutSession` implements the `WebCheckoutDelegate` which receives events of what happened during the web checkout. In a later patch this will 
  interact with the `UnityMessage` class to bubble up callbacks to the Unity layer.
* `WebCheckoutViewController` contains a few helper classes inside for creating the views it displays.

A couple of notes about the checking out experience:

* I've made the web checkout look like an overlay modal so the user can see the game in the background instead of fully covering the screen. Makes it feel like
  you're still in the game.
* The button at the bottom changes from Cancel -> Finish when the web view navigates to the last page of the checkout process (.../thank_you).
* The web view doesn't get auto-dismissed but is something we can add. I'm not sure if we want this since as a customer I'd like to see the final screen before closing.

I have some tests I'm working on still as well so I'll follow up with those and the C# implementation.


Assets/Plugins/iOS/Shopify/WebCheckoutViewController.swift
```diff
@@ -0,0 +1,236 @@
...
+    func endInitialLoad() {
+        UIView.animate(withDuration: 0.3) {
+            self.loadingView.alpha = 0
+            self.initialLoading = false
+        }
+    }
+}
+
+// Loading view overlay presented on top of the web view when its first loading.
+private class LoadingView: UIView {
```
sleroux: I add this loading view on top of the web view on initial load to make the experience a little cleaner while the user is waiting for the page to render. I'm using the default loading indicator but we can do something custom here. We can also expose this to the Unity developer to add their own branded loading indicator.

Assets/Plugins/iOS/Shopify/WebCheckoutViewController.swift
```diff
@@ -0,0 +1,236 @@
...
+                self.checkoutView.endInitialLoad()
+            }
+            
+            return
+        }
+        
+        if let url = webView.url, isThankYou(url: url) {
+            // Change the color and title of the button
+            checkoutView.setButton(state: .finished)
+            delegate?.didFinishCheckout()
```
dengjeffrey: `didFinishCheckout()` signals that the web view has gone away? Should this be called once `checkoutView`s `cancelButton` is pressed?

sleroux: Instead of a single `didCloseWebView` method I split it up into two calls: `didFinishCheckout` and `didCancelCheckout`. My thoughts are that maybe the game developer would want to know when a user completes a purchase vs cancelling out of the checkout process.

The more I think about it, it might make sense to have a single callback for when the web view dismissed (`didDismissWebCheckout`) and have another one specifically for when the user completes a purchase `didCompletePurchase` which is fired when we get on the thank_you page.

sleroux:
@dengjeffrey Ready for another pass! 

I included some updates to the colors to be in line with Polaris as well as cleaned up the delegate events.

Assets/Plugins/iOS/Shopify/UIColor+Hex.swift
```diff
@@ -0,0 +1,48 @@
+//
+//  UIColor+Hex.swift
+//  Unity-iPhone
```
dengjeffrey: Nit: need the copyright notice

Assets/Plugins/iOS/Shopify/UIColor+Hex.swift
```diff
@@ -0,0 +1,48 @@
...
+        scanner.scanHexInt64(&rgbValue)
+        
+        let r = (rgbValue & 0xff0000) >> 16
+        let g = (rgbValue & 0xff00) >> 8
+        let b = rgbValue & 0xff
+        
+        self.init(
+            red: CGFloat(r) / 0xff,
+            green: CGFloat(g) / 0xff,
+            blue: CGFloat(b) / 0xff, alpha: 1
```
dengjeffrey: nit: 
```
blue: CGFloat(b),
alpha: 1
```

Assets/Plugins/iOS/Shopify/UIColor+Hex.swift
```diff
@@ -0,0 +1,48 @@
...
+import UIKit
+
+extension UIColor {
+    convenience init(hex: String) {
+        let scanner = Scanner(string: hex)
+        scanner.scanLocation = 0
+        
+        var rgbValue: UInt64 = 0
+        
+        scanner.scanHexInt64(&rgbValue)
```
dengjeffrey: Should this be a failable init? What happens when `scanHexInt64` returns false

sleroux:
@dengjeffrey I've updated the PR to include the comment fixes along with rounding out the UI and Unity message passing. On top of the changes,

- I've added a slight black overlay that gets displayed over the game content in the background to give the effect of a modal.
- Messages are now properly sent to Unity instead of `print` functions.
- Updated the callbacks to match the new web checkout API here: `https://github.com/Shopify/unity-buy-sdk/pull/114`

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -0,0 +1,145 @@
...
+        let session = WebCheckoutSession(unityDelegateObjectName: unityDelegateObjectName,
+                                         checkoutURL: url)
+        activeWebPaySession = session
+        return session;
+    }
+    
+    fileprivate var webViewController: WebCheckoutViewController?
+    fileprivate var overlay: UIView?
+    
+    private(set) var checkoutURL: String
```
dengjeffrey: Should this be a constant?

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -0,0 +1,145 @@
...
+    private(set) var unityDelegateObjectName: String
```
dengjeffrey: `let`?

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -0,0 +1,145 @@
...
+            return false;
+        }
+        
+        webViewController = WebCheckoutViewController(url: url, delegate: self)
+        webViewController!.modalPresentationStyle = .overFullScreen
+        
+        toggleOverlay(show: true, onVC: root)
+        
+        willPresentWebCheckout()
+        toggleOverlay(show: true, onVC: root)
```
dengjeffrey: Since `willPresentWebCheckout` is a call made asynchronously there are no guarantees that the Web Checkout isn't already presented by the time we receive this message on `Unity`. Will this be a problem?

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -0,0 +1,145 @@
...
+        root.present(webViewController!, animated: true) {
+            self.didPresentWebCheckout()
+        }
+        
+        return true;
+    }
+    
+    fileprivate func toggleOverlay(show: Bool, onVC vc: UIViewController) {
```
dengjeffrey: I believe this could be made more clear, `toggle` means to flip a state, but in this case overlay is set to `show`.
Maybe splitting it into two, `show` and `hide`? Similar to `StartAnimating` and `StopAnimating`

sleroux:
@dengjeffrey What do you think about renaming these classes WebPay* instead of WebCheckout*?

dengjeffrey:
@sleroux That sounds good to me

dengjeffrey:
I'm still not convinced that we should handle the loading view for the implementor by default. This requires us to "implement" our own loading screens on Android and iOS. If they want to change this then they need to write/modify them on the native side specifically for each. They might want to re-use their own loading animation they have already made.  

 Can we not load the `WebView` with the URL without adding it as a subview. When . the `WebView` delegate notifies us that the URL has been loaded, we can send a `willPresent` state to the implementor, and wait for a response before we actually add the `WebView`. 

Before `willPresent` they can display their own loading screen. Or we can have a flag that says to use our default loading screen.

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -29,14 +29,15 @@ import WebKit
...
+
 protocol WebCheckoutDelegate: class {
-    func willPresentWebCheckout()
-    func didPresentWebCheckout()
-    
-    func didDismissWebCheckout()
-    func willDismissWebCheckout()
-    
-    func didCompletePurchase()
+    func onChangedWebUIState(newState: WebUIState, receivedUnityResponseCallback: UnityMessage.MessageCompletion?)
```
dbart01: Nit: Seems very Javascripty to me. A more conventional name would be something like:
```swift
func checkout(checkout: WebCheckoutSession, didChange state: State, completion: UnityMessage.MessageCompletion?)
```

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -29,14 +29,15 @@ import WebKit
 
 private var activeWebPaySession: WebCheckoutSession?
 
+enum WebUIState: String {
```
dbart01: Can this not be nested under `WebCheckoutSession` as `State`?

sleroux:
Since it's kind of a pain to run this:

web-1
web-2

sleroux:
I've folded down the commits and added fixes to the feedback throughout as well as changes from the C# side of things here: https://github.com/Shopify/unity-buy-sdk/pull/114

Assets/Plugins/iOS/Shopify/WebCheckoutViewController.swift
```diff
@@ -0,0 +1,345 @@
...
+// MARK: Actions
+extension WebCheckoutViewController {
+    func loadCheckoutURL() {
+        let request = URLRequest(url: url)
+        webView.load(request)
+    }
+    
+    func didPressCancel() {
+        delegate?.willDismiss()
+        self.dismiss(animated: true)
```
dengjeffrey: Do we want to wait for a response from `willDismiss`? The oder of execution for `self.dismiss` and `willDismiss` is not guaranteed, since `willDismiss()` is sent async

sleroux:
@dengjeffrey @dbart01 Ready for another pass. Sorry for all of the iteration on this PR - should be solid now.

Assets/Plugins/iOS/Shopify/WebCheckoutViewController.swift
```diff
@@ -0,0 +1,345 @@
...
+    private let cornerRadius: CGFloat = 3
+    private let viewMargin: CGFloat = 10
+    private let buttonHeight: CGFloat = 46
+    private let loadingIndicatorOffset: CGFloat = 5
+    private let loadingIndicatorTiming: TimeInterval = 0.3
+    
+    private(set) var containerView: UIView!
+    private(set) var webView: WKWebView!
+    private(set) var actionButton: CheckoutButton!
+    private(set) var progressIndicator: WebPageProgressIndicator!
```
dengjeffrey: Do these variables need to be implicitly unwrapped?

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -0,0 +1,128 @@
...
+        let root = UnityAppController.root
+        guard let url = URL(string: checkoutURL) else {
+            return false;
+        }
+        
+        webViewController = WebCheckoutViewController(url: url, delegate: self)
+        webViewController!.modalPresentationStyle = .overFullScreen
+        
+        showOverlay()
+        root.present(self.webViewController!, animated: true, completion: nil)
```
dengjeffrey: Doesn't seem like `webViewController` is used anywhere other than here. We can implicitly unwrap `webViewController`

```
fileprivate var webViewController: WebCheckoutViewController!
```

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -0,0 +1,128 @@
...
+        guard let overlay = overlay else {
+            return
+        }
+        
+        let root = UnityAppController.root
+        root.view.addSubview(overlay)
+        root.view.bringSubview(toFront: overlay)
+        UIView.animate(withDuration: overlayAnimationDuration, animations: {
+            overlay.alpha = 0.3
+        }) { _ in }
```
dengjeffrey: The completion is optional

Assets/Plugins/iOS/Shopify/WebCheckoutViewController.swift
```diff
@@ -0,0 +1,345 @@
...
+    public func webView(_ webView: WKWebView, didStartProvisionalNavigation navigation: WKNavigation!) {
+        checkoutView.showProgressIndicator()
+    }
+    
+    public func webView(_ webView: WKWebView, didFailProvisionalNavigation navigation: WKNavigation!, withError error: Error) {}
+    
+    public func webView(_ webView: WKWebView, didCommit navigation: WKNavigation!) {
+        if let url = webView.url, isThankYouPage(url) {
+            delegate?.didLoadThankYouPage()
+            self.checkoutView.isFinished = true
```
dengjeffrey: `self` is not needed here to keep consistency

Assets/Plugins/iOS/Shopify/WebCheckoutViewController.swift
sleroux: Nope! Was used to declaring them on a view controller but since we set them up in `init` it's good.

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -70,10 +70,10 @@ protocol WebCheckoutDelegate: class {
         }
         
         webViewController = WebCheckoutViewController(url: url, delegate: self)
-        webViewController!.modalPresentationStyle = .overFullScreen
+        webViewController.modalPresentationStyle = .overFullScreen
         
         showOverlay()
-        root.present(self.webViewController!, animated: true, completion: nil)
+        root.present(self.webViewController, animated: true, completion: nil)
```
dengjeffrey: Nit: `self`

dbart01:
The number of classes in `WebCheckoutViewController.swift` makes me very uneasy. Is there no way to split them up into logical files?

sleroux:
Yup we can break out the classes into their own files or move the view classes into a `WebCheckoutViews.swift` file. What do you think would be better?

dbart01:
I would split into separate files. I personally really like the name spacing approach as well. It would look something like this:
```swift
// Checkout.swift
enum Checkout {} // Dummy namespace, we use `enum` because it can't be instantiated

// Checkout.WebViewController.swift
extension Checkout {
    class WebViewController: UIViewController {
        ...
    }
}

// Checkout.WebView.swift
extension Checkout {
    class WebView: WKWebView {
        ...
    }
}

// Checkout.State.swift
extension Checkout {
    enum State {
        ...
    }
}
```

This allow much cleaner and more concise names within the classes and any `public` classes can be used with the `Checkout` prefix.

sleroux:
@dbart01 Updated the PR to include the namespace breakdown. I really like that approach!

Would you be able to give it another look before EOD today? 🙏

Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebContentView.swift
```diff
@@ -0,0 +1,142 @@
...
+    private let cornerRadius: CGFloat = 3
+    private let viewMargin: CGFloat = 10
+    private let buttonHeight: CGFloat = 46
+    private let loadingIndicatorOffset: CGFloat = 5
+    private let loadingIndicatorTiming: TimeInterval = 0.3
+    
+    private(set) var containerView: UIView
+    private(set) var webView: WKWebView
+    private(set) var actionButton: Button
+    private(set) var progressIndicator: WebPageProgressIndicator
```
dbart01: Do these all have to be `var`? Are they getting modified anywhere? Can they potentially be converted to constant?

Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebContentView.swift
```diff
@@ -0,0 +1,142 @@
...
+        actionButton = Button(cornerRadius: cornerRadius)
+        actionButton.translatesAutoresizingMaskIntoConstraints = false
+        
+        progressIndicator = WebPageProgressIndicator(cornerRadius: 3)
+        progressIndicator.translatesAutoresizingMaskIntoConstraints = false
+        progressIndicator.alpha = 0
+        
+        super.init(frame: frame)
+        
+        webView.addObserver(self, forKeyPath: #keyPath(WKWebView.estimatedProgress), options: .new, context: nil)
```
dbart01: If this class can be subclassed, we should pass in an explicit `context` value and check for it in the observation method. If not, the alternative is to declare this as `final class`. Would also prefer to see something like KVO be very explicit in a separate method:
```swift
private func registerObservers()
private func unregisterObservers()

override func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey : Any]?, context: UnsafeMutableRawPointer?)
```

Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebPageProgressIndicator.swift
```diff
@@ -0,0 +1,85 @@
...
+
+import UIKit
+
+extension Checkout {
+
+// Custom loading indicator that it shown at the top when loading a web page.
+class WebPageProgressIndicator: UIView {
+    var progress: CGFloat = 0 {
+        didSet {
+            DispatchQueue.main.async {
```
dbart01: Why is the dispatch necessary?

Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebViewController.swift
```diff
@@ -0,0 +1,118 @@
...
+        
+        super.init(nibName: nil, bundle: nil)
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    override func loadView() {
+        self.view = WebContentView(frame: UIScreen.main.bounds)
```
dbart01: You should still call `super.loadView()` and use `self.view.frame` instead of `UIScreen.main.bounds`. Much more bulletproof.

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -0,0 +1,128 @@
...
+    }
+    
+    fileprivate func hideOverlay() {
+        guard let overlay = overlay else {
+            return
+        }
+        
+        UIView.animate(withDuration: overlayAnimationDuration, animations: {
+            overlay.alpha = 0
+        }) { _ in
```
dbart01: Nit: For cases like this, it's much more readable if the trailing closure is not trailing:
```swift
UIView.animate(withDuration: overlayAnimationDuration, animations: {
    overlay.alpha = 0
}, completion: { _ in
    overlay.removeFromSuperview()
    self.overlay = nil
})
```

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -0,0 +1,128 @@
...
+            view.backgroundColor = .black
+            view.alpha = 0
+            return view
+        }()
+        
+        super.init()
+    }
+    
+    func startCheckout() -> Bool {
+        let root = UnityAppController.root
```
dbart01: Nit: This local variable seems to only be used in one place. Can just call this inline:
```swift
UnityAppController.root.present(webViewController, animated: true, completion: nil)
```

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -0,0 +1,128 @@
...
+    func didLoadThankYouPage()
+    func willDismiss(completionHandler: UnityMessage.MessageCompletion?)
+}
+
+@objc class WebCheckoutSession: NSObject {
+    static func createSession(unityDelegateObjectName: String, url: String) -> WebCheckoutSession {
+        let session = WebCheckoutSession(unityDelegateObjectName: unityDelegateObjectName,
+                                         checkoutURL: url)
+        activeWebPaySession = session
+        return session;
```
dbart01: `;` 🔥

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -0,0 +1,128 @@
...
+            return view
+        }()
+        
+        super.init()
+    }
+    
+    func startCheckout() -> Bool {
+        let root = UnityAppController.root
+        guard let url = URL(string: checkoutURL) else {
+            return false;
```
dbart01: `;` 🔥

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -0,0 +1,128 @@
...
+        }
+        
+        webViewController = Checkout.WebViewController(url: url, delegate: self)
+        webViewController.modalPresentationStyle = .overFullScreen
+        
+        showOverlay()
+        root.present(webViewController, animated: true, completion: nil)
+        
+        return true;
```
dbart01: `;` 🔥

Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebContentView.swift
sleroux: Nope! Moving them to constants.

Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebPageProgressIndicator.swift
sleroux: I had some animation logic which only worked properly if this was called on the main thread but that's no longer the case.

sleroux:
Thanks for all the feedback 👌 Should be good for another pass.

dengjeffrey:
🚢

PR #121, opened 2017-06-26
[Feature] Cart Apple Pay Shipping

dengjeffrey:
### What
+ Adds an Apple Pay Receiver to `Cart`
+ Received serialized Shipping and Email information from Apple Pay are deserialized and sent as GraphQL mutations
+ Combines email and shipping mutation queries, since they are not used separately 

### What is not included
+ Unit Tests
    + Will write the unit tests once we can create an apple pay session from Unity 
+ Error Refactors
    + Will leave the refactoring of Cart.cs later on to use Shopify Error

### DIFF
--- a/scripts/generator/graphql_generator/csharp/SDK/iOS/CartApplePayEventReceiver.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/iOS/CartApplePayEventReceiver.cs.erb
@@ -10,6 +10,7 @@ namespace <%= namespace %> {
     public partial class Cart : IApplePayEventReceiver {
 
         public void UpdateSummaryItemsForShippingIdentifier(string serializedMessage) {
+
             var message = NativeMessage.CreateFromJSON(serializedMessage);
 
             SetShippingLine(message.Content, (httpError, graphQLError) => {
@@ -23,6 +24,7 @@ namespace <%= namespace %> {
         }
 
         public void UpdateSummaryItemsForShippingContact(string serializedMessage) {
+
             var message = NativeMessage.CreateFromJSON(serializedMessage);
             var contentDictionary = (Dictionary<string, object>)Json.Deserialize(message.Content);
             var mailingAddressInput = new MailingAddressInput(contentDictionary);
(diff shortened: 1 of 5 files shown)


scripts/generator/graphql_generator/csharp/SDK/iOS/CartApplePayEventReceiver.cs.erb
```diff
@@ -0,0 +1,112 @@
...
+
+            try {
+                var availableShippingRates = CurrentCheckout.availableShippingRates().shippingRates();
+
+                foreach (var shippingRate in availableShippingRates) {
+                    shippingMethods.Add(new ShippingMethod(shippingRate.title(), shippingRate.price().ToString(), shippingRate.handle()));
+                }
+
+            } catch {
+                throw;
```
sleroux: Is this just a placeholder for the error handling code?

scripts/generator/graphql_generator/csharp/SDK/iOS/CartApplePayEventReceiver.cs.erb
```diff
@@ -0,0 +1,112 @@
...
+                        respondWithSuccess();
+                    } else {
+                        updateEmailSuccess = true;
+                    }
+                } else {
+                    message.Respond(new ApplePayEventResponse(ApplePayAuthorizationStatus.Failure).ToJsonString());
+                }
+            });
+
+            SetShippingAddress(mailingAddressInput, (httpError, graphqlError) => {
```
sleroux: Is there any issues with these two async calls racing each other? For example, `responseWithSuccess` will be called from different places at different times.

sleroux: I also noticed that `summaryItems` is declared above with `null` and is used in `respondWithSuccess`. If `SetEmail` is called before `SetShippingAddress` it will be null but if it's the other way around it will not be null.

sleroux: Might be best to chain these to force some kind of ordering.

sleroux:
Also I forgot to mention:

For partial class definitions I've been following this SO post https://stackoverflow.com/questions/1478610/naming-conventions-for-partial-class-files so for example this would be `Cart.ApplePayEventReceiver.cs.erb`. What do you think about that convention?

scripts/generator/graphql_generator/csharp/SDK/iOS/CartApplePayEventReceiver.cs.erb
dengjeffrey: I suppose it would be more clear if they were chained.

scripts/generator/graphql_generator/csharp/SDK/iOS/CartApplePayEventReceiver.cs.erb
dengjeffrey: No this just rethrows the exception, I suppose it would be best if we created our own exception to give more details to what is happening.

PR #122, opened 2017-06-26
[Feature] User Error Handling

mikkoh:
I was working on an example of how we could handle user errors.

Really user errors describe one of the states of the checkout that is not defined in `Checkout`. This state is that the `Checkout` can be in erroneous state. So my thought was to add the list of user errors to the `Cart` beside `CurrentCheckout`.

This list should be exposed to the user. I think with the work that @sleroux has been doing on refactoring errors we could send an error that is of type user error and in documentation we tell the user to check user errors on cart.

Thoughts? This PR should be finished after #107.

### DIFF
--- a/scripts/generator/graphql_generator/csharp/Cart.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/Cart.cs.erb
@@ -177,14 +177,16 @@ namespace <%= namespace %> {
         }
 
         private void UpdateLineItemFromCheckout(Checkout checkout) {
-            if (checkout != null) {
-                // sometimes we may not query line items for instance when polling is being performed
-                try {
-                    List<CheckoutLineItem> lineItems = (List<CheckoutLineItem>) checkout.lineItems();
-
-                    LineItems.UpdateLineItemsFromCheckoutLineItems(lineItems);
-                } catch(NoQueryException exception) {}
+            if (checkout == null) {
+                return;
             }
+            
+            // sometimes we may not query line items for instance when polling is being performed
+            try {
+                List<CheckoutLineItem> lineItems = (List<CheckoutLineItem>) checkout.lineItems();
+
+                LineItems.UpdateLineItemsFromCheckoutLineItems(lineItems);
+            } catch(NoQueryException exception) {}
         }
 
         private void HandleUserError(Action<ShopifyError> callback) {
--- a/scripts/generator/graphql_generator/csharp/Cart.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/Cart.cs.erb
@@ -43,6 +43,7 @@ namespace <%= namespace %> {
         private List<String> DeletedLineItems = new List<string>();
         private ShopifyClient Client;
         private Checkout CurrentCheckout = null;
+        private List<UserError> CurrentUserErrors = null;
 
         /// <summary>
         /// Constructs a new cart using a <see ref="ShopifyClient">ShopifyClient </see>. Typically, carts won't be
@@ -100,12 +101,15 @@ namespace <%= namespace %> {
                     return;
                 }
 
-                UpdateCheckout(response.checkoutCreate().checkout());
-
-                if (CurrentCheckout.ready()) {
-                    callback(null);
+                if (UpdateState(response.checkoutCreate().checkout(), response.checkoutCreate().userErrors())) {
+                    if (CurrentCheckout.ready()) {
+                        callback(null);
+                    } else {
+                        PollCheckoutAndUpdate(PollCheckoutReady, callback);
+                    }
                 } else {
-                    PollCheckoutAndUpdate(PollCheckoutReady, callback);
+                    // HANDLE USER ERROR
+                    throw new NotImplementedException("Mikko implement handling user error");
                 }
             });
         }
@@ -129,17 +133,25 @@ namespace <%= namespace %> {
                 }
 
                 DeletedLineItems.Clear();
-                UpdateCheckout(response.checkoutLineItemsAdd().checkout());
 
-                if (CurrentCheckout.ready()) {
-                    callback(null);
+                if (UpdateState(response.checkoutLineItemsAdd().checkout(), response.checkoutLineItemsAdd().userErrors())) {
+                    if (CurrentCheckout.ready()) {
+                        callback(null);
+                    } else {
+                        PollCheckoutAndUpdate(PollCheckoutReady, callback);
+                    }
                 } else {
-                    PollCheckoutAndUpdate(PollCheckoutReady, callback);
+                    // HANDLE USER ERROR
+                    throw new NotImplementedException("Mikko implement handling user error");
                 }
             });
         }
 
-        private void UpdateCheckout(Checkout checkout) {
+        private bool UpdateState(Checkout checkout) {
+            return UpdateState(checkout, new List<UserError>());
+        }
+
+        private bool UpdateState(Checkout checkout, List<UserError> userErrors) {
             if (CurrentCheckout == null) {
                 CurrentCheckout = checkout;
             } else {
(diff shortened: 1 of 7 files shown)


sleroux:
Just for my understanding, is a user error is something the user does that is not valid? For example, if I enter some crazy postal code I'll get back a `UserError` with the fields that are not valid. 

I think this makes sense. After each mutation we make, call `UpdateState` to analyze if there are any user errors present and add them to the Cart for the user to look up. Is there an approach where we can fold in the error along with the userErrors into `ShopifyError` instead of having the user need to look in 2 places for the errors? 

----
Some other random related thoughts on the error stuff:

It's definitely challenging to bubble up all of our varied error types through a single callback. I keep thinking of it like a synchronous API which would throw. In that case it would look something like;

```csharp
try {
   cart.DoSomething
} catch (UserError e) {
 ...
} catch (HttpError e) {
 ...
} catch (GraphQLError e) {
 ...
}
```

Is it worth doing that kind of model here with the `ShopifyError` callback param?

```csharp
cart.DoSomething(stuff, e => {
    if (var userError = (UserError)e) {
     ...
    } else if (var httpError = (HttpError)e) {
     ...
    } else if (var graphQLError = (GraphQLError)e) {
     ...
    }
})
```

Another approach we could do is instead of using a callback block that only returns an error we allow the developer to pass in an object that listens for these errors:

```csharp
interface IShopifyErrorReceiver {
    void OnUserError(UserError e);
    void OnHTTPError(HTTPError e);
    void OnGraphQLError(GraphQLError e);
}

partial class GameDevStuff {
    cart.DoSomething(result => {
        // Always a happy path
    }, this);
}

partial class GameDevStuff: IShopifyErrorReceiver {
    void OnUserError(UserError e) {}
    void OnHTTPError(HTTPError e) {}
    void OnGraphQLError(GraphQLError e) {}
}
```

This might help consolidate the error checks to a single place instead of having the game developer write boilerplate for each callback block.

mikkoh:
Just to be clear I dropped in `NotImplementedException` to denote that this part needs to be coded still. My thought was not to throw exceptions. Here's a code example of usage:

```
cart.DoSomething(stuff, (e) => {
  switch(e.type) {
    case ErrorType.HTTP:
        ...
    case ErrorType.GraphQL:
        ...
    case ErrorType.UserError:
        foreach(UserError error in cart.UserErrors) {
            ...
        }
  }
});
```

You're right that right now there's nothing that meaningful that `e` would contain in the `description`. I'm not sure we can serialized anything useful here.

dengjeffrey:
If it helps here is what a user error looks like:
```
"userErrors": [
        {
          "field": [
            "input",
            "email"
          ],
          "message": "Email is invalid"
        }
```

mikkoh:
@sleroux @dengjeffrey I've updated this PR based on the PRs merged yesterday. Could I get some 👀 on this.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -101,8 +109,11 @@ namespace <%= namespace %> {
                     return;
                 }
 
-                UpdateCheckout(response.checkoutCreate().checkout());
-                CheckoutPoll(callback);
+                if (UpdateState(response.checkoutCreate().checkout(), response.checkoutCreate().userErrors())) {
```
dengjeffrey: Why isn't `userErrors()` part of `ShopifyError` from `TopLevelResponse` ?

mikkoh: No those are GraphQL errors. This is basically syntax errors when writing the GraphQL query.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -139,16 +154,35 @@ namespace <%= namespace %> {
...
+
+                    LineItems.UpdateLineItemsFromCheckoutLineItems(lineItems);
+                } catch(NoQueryException exception) {}
+            }
+        }
+
+        private void HandleUserError(List<UserError> userErrors, Action<ShopifyError> callback) {
+            ShopifyError error = new ShopifyError(
+                ShopifyError.ErrorType.UserError,
+                "There were issues with some of the fields sent. Checkout `cart.UserErrors`"
```
mikkoh: I think `UserErrors` as errors from a form. You might have multiple errors in the form at any given time. What error is most important? In the end the developer might want to handle all errors at the same time. I mentioned this issue in #105 that for `UserErrors` its tough to return any kind of meaningful data in the `description`.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -26,6 +26,12 @@ namespace <%= namespace %> {
             }
         }
 
+        public List<UserError> UserErrors {
```
dengjeffrey: I've been thinking about this, and I'm wondering have you considered subclassing `ShopifyError` -> `ShopifyUserError`, that includes this list of UserErrors?

`UserErrors` can become out of sync on the fly, if implementor is always accessing it like so `Cart.UserErrors` they might catch `UserErrors` from another Query.

mikkoh: This is true. But you also want to persist `UserErrors` instead of just an event/callback firing.

sleroux: I guess you could also see it as part of the Cart state. Like the cart is in an invalid state and this is why.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -139,16 +154,35 @@ namespace <%= namespace %> {
...
             UpdateLineItemFromCheckout(CurrentCheckout);
+
+            return _UserErrors == null;
         }
 
         private void UpdateLineItemFromCheckout(Checkout checkout) {
-            // sometimes we may not query line items for instance when polling is being performed
-            try {
-                List<CheckoutLineItem> lineItems = (List<CheckoutLineItem>) checkout.lineItems();
+            if (checkout != null) {
```
sleroux: Nit: Question too - do we favor early returns vs adding another level of code? 

For example,

```csharp
if (checkout == null) {
    return;
}

... continue on your merry way!
```

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -139,16 +154,35 @@ namespace <%= namespace %> {
...
+
+                    LineItems.UpdateLineItemsFromCheckoutLineItems(lineItems);
+                } catch(NoQueryException exception) {}
+            }
+        }
+
+        private void HandleUserError(Action<ShopifyError> callback) {
+            ShopifyError error = new ShopifyError(
+                ShopifyError.ErrorType.UserError,
+                "There were issues with some of the fields sent. Checkout `cart.UserErrors`"
```
sleroux: Nit: Super minor but maybe reword 'Checkout' to 'See' since Checkout is a bit overloaded 🙃

scripts/generator/graphql_generator/csharp/Cart.cs.erb
mikkoh: I've never been a fan of early returns but it seems that most of the team does so I'm ok changing.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
mikkoh: HAH good catch.

mikkoh:
@sleroux yeah I've always found `UserError` to be weird. It seems harsh for some reason to point fingers at the user and say they created an error. HA. There are two reasons why I think it's important to continue to use `UserError`:

1. This a part of Shopify's nomenclature. People who are used to the Shopify eco-system and apis will find `UserError` to be familiar.
2. A lot of the documentation outside the Unity SDK also talks about user errors

scripts/generator/graphql_generator/csharp/SDK/NativeMessage.cs.erb
```diff
@@ -1,3 +1,4 @@
+#if !SHOPIFY_MONO_UNIT_TEST
 namespace <%= namespace %>.SDK {
 #if !SHOPIFY_MONO_UNIT_TEST
```
dengjeffrey: 🤔 this came back

PR #125, opened 2017-06-27
[Feature] Checkouts Token Payment

dengjeffrey:
+ Adds Token Payment queries for creating a payment and polling to default queries
+ Adds tests


scripts/generator/graphql_generator/csharp/SDK/DefaultQueriesCheckout.cs.erb
```diff
@@ -29,6 +43,14 @@ namespace <%= namespace %>.SDK {
             );
         }
 
+        public void PaymentPoll(QueryRootQuery query, string paymentId) {
+            query.node(
+                buildQuery: node => node
+                .onPayment(payment => Payment(payment)),
```
richardmonette: nitpick: I think this line needs to be tabbed in

PR #129, opened 2017-06-28
[Feature] Poll refactor

dengjeffrey:
+ Adds a generalized polling function to `ShopifyBuy` Client

#### Motivation
+ Since are a few graphql queries that use `ready` we need to be able to poll for those
    + Checkout
    + Available Shipping Rates
    + Payment
+ Decided to put this function in `ShopifyBuy`  since it is generally just making a graphql query
+ Chose to put the polling methods into a separate partial class to keep the `Cart` file from being bloated

### DIFF
--- a/scripts/generator/graphql_generator/csharp.rb
+++ b/scripts/generator/graphql_generator/csharp.rb
@@ -87,6 +87,7 @@ module GraphQLGenerator
       %w(
         ShopifyBuy
         Cart
+        Cart.Poll
         SDK/Arguments
         SDK/InputBase
         SDK/InputValueToString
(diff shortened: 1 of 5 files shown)


mikkoh:
I still wonder if we should just use `checkout.ready()` instead of drilling down to `checkout. availableShippingRates().ready()`.

dengjeffrey:
There is the case where `availableShippingRates` is not ready and `Checkout` is 😞:

```
{
  "data": {
    "checkoutCreate": {
      "checkout": {
        "id": "gid://shopify/Checkout/2516049ec594c7f5d2bc8b78375c7437?key=471d49f9a819ccd507197e954c35b4ce",
        "webUrl": "https://checkout.myshopify.io/1/checkouts/2516049ec594c7f5d2bc8b78375c7437?key=471d49f9a819ccd507197e954c35b4ce",
        "requiresShipping": true,
        "availableShippingRates": {
          "shippingRates": null,
          "ready": false
        },
        "subtotalPrice": "
```

dengjeffrey:
@mikkoh 👍 ?

mikkoh:
Wow. This should be an issue on core. Ok lets go with this for now.

PR #132, opened 2017-06-29
[Feature] iOS Native Payable Structure

dengjeffrey:
+ Adds interface for native payable
+ Adds iOS native pay implementation 

#### Decisions
+ Chose to have `key` passed into `StartCheckout(string key)` instead of when we instantiate a `iOSNativePayable`.
     + When `iOSNativePayable` is instantiated in the constructor of `Cart` following #114 we do not have the `merchantId` yet

### DIFF
--- a/scripts/generator/graphql_generator/csharp.rb
+++ b/scripts/generator/graphql_generator/csharp.rb
@@ -126,6 +126,7 @@ module GraphQLGenerator
         SDK/iOS/ShippingMethod
         SDK/iOS/SummaryItem
         SDK/iOS/Cart.ApplePayEventReceiver
+        SDK/iOS/iOSNativeCheckout
       ).each do |class_file_name|
         directory = "#{path}/#{File.dirname(class_file_name)}"
 
--- a/scripts/generator/graphql_generator/csharp.rb
+++ b/scripts/generator/graphql_generator/csharp.rb
@@ -119,6 +119,7 @@ module GraphQLGenerator
         SDK/Log
         SDK/NativeMessage
         SDK/Serializable
+        SDK/INativeCheckout
         SDK/iOS/ApplePayAuthorizationStatus
         SDK/iOS/ApplePayEventReceiver
         SDK/iOS/ApplePayEventResponse
--- a/scripts/generator/graphql_generator/csharp/SDK/INativeCheckout.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/INativeCheckout.cs.erb
@@ -1,0 +1,10 @@
+namespace <%= namespace %>.SDK {
+
+    interface INativeCheckout {
+        bool CanShowPaymentSetup();
+        void ShowPaymentSetup();
+        bool CanStart();
+        /// Key is the public key that is used for encrypting a native payment's token data
+        void Start(string key);
+    }
+}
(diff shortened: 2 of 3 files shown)


scripts/generator/graphql_generator/csharp/SDK/INativePayable.cs.erb
```diff
@@ -0,0 +1,10 @@
+namespace <%= namespace %>.SDK {
+
+    interface INativePayable {
+        bool CanShowPaymentSetup();
+        void ShowPaymentSetup();
+        bool CanCheckout();
+        /// Key is the public key that is used for encrypting a native payment's token data
+        void StartCheckout(string key);
```
sleroux: I like that you've named it `StartCheckout` instead of `StartNativeCheckout`. I wonder if calling it just `Start` would be too much? On my PR I've updated `IWebPayable` to be `IWebCheckout` and named the variable in Cart `webCheckout`. This way, the code reads like `webCheckout.Start(url)`. What do you think?

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativePayable.cs.erb
```diff
@@ -0,0 +1,66 @@
...
+        /// Starts the process of making a payment through Apple Pay.
+        /// </summary>
+        /// <remarks>
+        ///  Displays a payment interface to the user based on the contents of the Cart
+        /// </remarks>
+        /// <param name="key">Merchant ID for Apple Pay from the Apple Developer Portal</param>
+        public void StartCheckout(string key) {
+            // TODO
+        }
+    }
```
sleroux: I ended up defining the MonoBehaviour we attach to the Global game object at the bottom of the same file. What do you think about doing that? Not sure if it needs its own file or not.

scripts/generator/graphql_generator/csharp/SDK/INativePayable.cs.erb
dengjeffrey: I like it 👍 !

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativePayable.cs.erb
dengjeffrey: I think here it ends up being different, on Web your MonoBehaviour can invoke events straight away. On native side we need to access `Cart` so I prefer that the MonoBehaviour just be a really thin wrapper.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativePayable.cs.erb
```diff
@@ -0,0 +1,66 @@
...
+
+        [DllImport ("__Internal")]
+        private static extern bool _CreateApplePaySession(string merchantID, string countryCode, string currencyCode, string unityDelegateObjectName, string serializedSummaryItems, string serializedShippingMethods, bool requiringShipping);
+
+        [DllImport ("__Internal")]
+        private static extern void _PresentApplePayAuthorization();
+
+        private Cart CurrentCart;
+
+        public iOSNativePayable(Cart cart) {
```
richardmonette: I'd like to see this filled out a bit more before we make up our minds - but still wondering if the way the cart get information is via events dispatched from this class (and then we can avoid having to pass the cart here.)

sleroux:
Can't seem to change my comment to an 'Approve' but this is 👍 from me too.

PR #133, opened 2017-06-29
[Feature] Added the ability to checkout with token to cart

dengjeffrey:
+ These methods are private for now, till we finalize on how we handle talking to the `Cart` from an instance of `INativeCheckout`

### DIFF
--- a/scripts/generator/graphql_generator/csharp/Cart.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/Cart.cs.erb
@@ -120,6 +120,48 @@ namespace <%= namespace %> {
             });
         }
 
+        private void CheckoutWithTokenizedPayment(TokenizedPaymentInput tokenizedPaymentInput, Action<ShopifyError> callback) {
+
+            Action<Payment> pollPayment = (payment) => {
+                PollPaymentReady(payment.id(), (Payment newPayment, ShopifyError pollError) => {
+                    if (pollError == null) {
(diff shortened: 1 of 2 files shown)


scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -110,6 +110,48 @@ namespace <%= namespace %> {
             });
         }
 
+        private void CheckoutWithTokenizedPayment(TokenizedPaymentInput tokenizedPaymentInput, Action<ShopifyError> callback) {
```
sleroux: Something I didn't consider in defining the `INativeCheckout` and `IWebCheckout` interfaces is code that is related to a native checkout that is platform-agnostic. For example, this seems like something that:

1. Only is used for Native Payments
2. Can run for both Android/Apply Pay

If this is a pattern we start to see emerging we can define an abstract base class for native pay (`BaseNativeCheckout`) that implements this code. This way we can keep the checkout logic out of the Cart and delegate to the checkout objects.

dengjeffrey: We might need to consider this later since any queries that return a checkout needs to be set as the `CurrentCheckout` in cart. Adding the extra layer means we need to make `CurrentCheckout` public or move it somewhere else

dengjeffrey:
Had to refactor for the new `UserErrors`, please take another gander 👀

PR #137, opened 2017-06-30
[Feature] INativeCheckout

dengjeffrey:
+ Renames methods in INativeCheckout to fit the IWebCheckout
+ Adds checks before using ICheckouts so it doesn't crash on non iOS platforms

<hr>
+ Will add in exceptions that are thrown in a later PR

### DIFF
--- a/scripts/generator/graphql_generator/csharp/Cart.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/Cart.cs.erb
@@ -97,22 +97,30 @@ namespace <%= namespace %> {
         }
 
         public void CheckoutWithNativeWebView(CheckoutSuccessCallback success, CheckoutFailureCallback failure, CheckoutCancelCallback cancelled) {
-            GetWebCheckoutLink(url => {
-                WebCheckout.Checkout(url, success, failure, cancelled);
-            }, error => {
-                failure(error);
-            });
+            if (WebCheckout != null) {
+                GetWebCheckoutLink(url => {
+                    WebCheckout.Checkout(url, success, failure, cancelled);
+                }, error => {
+                    failure(error);
+                });
+            } else {
+                // TODO throw an exception
+            }
         }
 
         public void CheckoutWithNativePay(string key, CheckoutSuccessCallback success, CheckoutFailureCallback failure, CheckoutCancelCallback cancelled) {
-            CheckoutSave(error => {
-                if (error != null) {
-                    failure(error);
-                }
-                else {
-                    NativeCheckout.Checkout(key, success, failure, cancelled);
-                }
-            });
+            if (NativeCheckout != null) {
+                CheckoutSave(error => {
+                    if (error != null) {
+                        failure(error);
+                    }
+                    else {
+                        NativeCheckout.Checkout(key, success, failure, cancelled);
+                    }
+                });
+            } else {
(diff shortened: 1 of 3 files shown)


scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -95,11 +97,54 @@ namespace <%= namespace %> {
...
-                failure(error);
-            });
+            if (WebCheckout != null) {
+                GetWebCheckoutLink(url => {
+                    WebCheckout.Checkout(url, success, failure, cancelled);
+                }, error => {
+                    failure(error);
+                });
+            } else {
+                // TODO throw an exception,  this platform does not support checkout with web pay
```
richardmonette: Would we not call the failure callback here?

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -95,11 +97,54 @@ namespace <%= namespace %> {
...
+                    else {
+                        NativeCheckout.Checkout(key, success, failure, cancelled);
+                    }
+                });
+            } else {
+                // TODO throw an exception, this platform does not support checkout with native pay
+            }
+        }
+
+        public bool CanShowNativePaySetup() {
```
richardmonette: Are CanShowNativePaySetup and CanCheckoutWithNativePay sort of synonymous, i.e. when could I show the NativePay, but not also be able to Checkout?

dengjeffrey: NativePay Setup shows the iOS `Wallet` app, and instructs the user to add a card to Apple Pay. You won't be able to checkout if the user has no cards added to `Apple Pay`

richardmonette: Ah kk.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
dengjeffrey: Ideally we should have a `CanCheckoutWithNativeWeb` because this feature isn't available on Android. A Failure callback means a Web checkout started but failed for some temporary reason `HTTP Error` etc. An exception in this case means you can't checkout at all, and it was the programmer's fault for misleading the user to do so

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -95,11 +97,54 @@ namespace <%= namespace %> {
...
+                }, error => {
+                    failure(error);
+                });
+            } else {
+                // TODO throw an exception,  this platform does not support checkout with web pay
+            }
+        }
+
+        public void CheckoutWithNativePay(string key, CheckoutSuccessCallback success, CheckoutFailureCallback failure, CheckoutCancelCallback cancelled) {
+            if (NativeCheckout != null) {
```
sleroux: Nit: Super minor but we can flatten this out a bit by early returning:

```csharp
if (NativeCheckout == null) {
    return;
}

CheckoutSave(error => {
    if (error != null) {
        failure(error);
    } else {
        NativeCheckout.Checkout(key, success, failure, cancelled);
    }
});
```

PR #139, opened 2017-07-04
[Feature] Native checkout token deserialization

dengjeffrey:
+ Adds deserialization for token data
+ Refactors serializable classes


scripts/generator/graphql_generator/csharp/SDK/iOS/PaymentDeserializer.cs.erb
```diff
@@ -0,0 +1,52 @@
...
+            ShippingIdentifier,
+            TokenData
+        }
+
+        private enum TokenField {
+            PaymentData,
+            TransactionIdentifier
+        }
+
+        public MailingAddressInput billingAddress { get; }
```
sleroux: Could we use `readonly` here instead of defining a `get` operator? Just curious.

mikkoh: Should these be `BillingAddress`

PR #142, opened 2017-07-04
[Feature] Implements DidFinishCheckoutSession for iOSNativeCheckout

sleroux:
* Parses PaymentStatus from iOS to determine which callback to fire on the C# side.
* Introduces the CheckoutError error type to ShopifyError

### DIFF
--- a/scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
@@ -16,7 +16,7 @@ namespace <%= namespace %>.SDK {
         }
 
         public enum ErrorType {
-            HTTP, GraphQL, UserError
+            HTTP, GraphQL, UserError, CheckoutError
         }
 
         public readonly ErrorType Type;
--- a/scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
@@ -16,7 +16,7 @@ namespace <%= namespace %>.SDK {
         }
 
         public enum ErrorType {
-            HTTP, GraphQL, UserError, CheckoutError
+            HTTP, GraphQL, UserError, NativePaymentProcessingError
         }
 
         public readonly ErrorType Type;
--- a/scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
@@ -58,11 +58,9 @@ namespace <%= namespace %>.SDK.iOS {
                 OnCancelled();
                 return;
             case NativePaymentStatus.Failed:
-                var error = new ShopifyError(ShopifyError.ErrorType.CheckoutError, "Unable to retrieve a payment from the user's payment provider. Fallback to web checkout.");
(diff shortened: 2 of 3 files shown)


scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
```diff
@@ -16,7 +16,7 @@ namespace <%= namespace %>.SDK {
         }
 
         public enum ErrorType {
-            HTTP, GraphQL, UserError
+            HTTP, GraphQL, UserError, CheckoutError
```
sleroux: I'm not sure about calling this `CheckoutError`. I don't really want to use `ApplePayError` either since thats too platform specific. Maybe `PaymentError`?

dengjeffrey: Perhaps: `NativePaymentProcessingError`, since this case happens when a native library fails

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -44,7 +48,22 @@ namespace <%= namespace %>.SDK.iOS {
         }
 
         public void DidFinishCheckoutSession(string serializedMessage) {
-            //TODO
+            var message = NativeMessage.CreateFromJSON(serializedMessage);
+            var paymentStatus = (NativePaymentStatus) Enum.Parse(typeof(NativePaymentStatus), (string) message.Content);
```
dengjeffrey: nice!

scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
mikkoh: I like `NativePaymentProcessingError` since it describes where the error is coming from an in what case.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -44,7 +48,22 @@ namespace <%= namespace %>.SDK.iOS {
...
+                return;
+            case NativePaymentStatus.Cancelled:
+                OnCancelled();
+                return;
+            case NativePaymentStatus.Failed:
+                var error = new ShopifyError(ShopifyError.ErrorType.CheckoutError, "Unable to retrieve a payment from the user's payment provider. Fallback to web checkout.");
+                OnFailure(error);
+                return;
+            default:
+                throw new NotImplementedException(string.Format("Unable to handle PaymentStatus: {0}", message.Content));
```
mikkoh: I think an exception will actually be thrown by `Enum.Parse` and you'll never land in `default`.

sleroux:
Ready for another pass.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
mikkoh: So our assumption here is that `message.Content` will only contain `"Success", "Cancelled", "Failed"` correct?

sleroux: Yes that's correct. If we are sending anything else an exception will be thrown which I think makes sense since we don't want the SDK developer to assume that they can send anything and have it be handled without adding the handling code in C#.

PR #143, opened 2017-07-04
[Docs] Updates examples to include native web view checkout

sleroux:


EXAMPLES.md
```diff
@@ -217,6 +214,50 @@ cart.LineItems.Get(firstProduct, selectedOptions);
...
+
+After creating an instance of `Cart` and adding items to it, you can use the `CheckoutWithNativeWebView` method to 
+start a native modal overlay on top of your game with a web view containing the checkout for the cart.
+
+`CheckoutWithNativeWebView` which takes in 3 callback parameters:
+
+* `CheckoutDidSucceed` is called when the user has completed a checkout successfully.
+
+* `CheckoutDidCancel` is called when the user cancels out of the checkout.
+
```
mikkoh: Remove these empty lines.

dengjeffrey:
Might want to also note the possible errors that are returned from making the native call. And what errors you do not need to handle

EXAMPLES.md
```diff
@@ -217,6 +214,50 @@ cart.LineItems.Get(firstProduct, selectedOptions);
...
+* `CheckoutDidFail` is called when an error was encountered during the web checkout. The callback will be passed an instance of `ShopifyError` describing the issue.
+
+```cs
+// Sample code for adding some product variants to your cart.
+var cart = ShopifyBuy.Client().Cart();
+var secondProductVariants = (List<ProductVariant>) products[1].variants();
```
mikkoh: I wonder if we should do:
```
var secondProduct = products[1];
var secondProductVariants = (List<ProductVariant>) secondProduct.variants();
```

It's always better to be completely obvious in documentation.

EXAMPLES.md
```diff
@@ -217,6 +214,50 @@ cart.LineItems.Get(firstProduct, selectedOptions);
...
+ProductVariant secondProductFirstVariant = secondProductVariants[0];
```
mikkoh: Might be good to change `secondProductFirstVariant` to be `productVariantToCheckout`

EXAMPLES.md
```diff
@@ -217,6 +214,50 @@ cart.LineItems.Get(firstProduct, selectedOptions);
...
+
+cart.LineItems.AddOrUpdate(secondProductFirstVariant.id(), 1);
```
mikkoh: You can just do:
```
cart.LineItems.AddOrUpdate(secondProductFirstVariant, 1);
```

EXAMPLES.md
```diff
@@ -217,6 +214,50 @@ cart.LineItems.Get(firstProduct, selectedOptions);
...
+ );
+```
+
+**Caveats**
+
+There are a few things we're still working out for the web checkout experience:
+
+1. Handling HTTP errors gracefully. When the user loses their internet connection there is currently no way to recover.
+2. Validating the web checkout purchase with results from the API. The completion callback is invoked when we detect that the user has navigated to the
+   thank\_you page. This is the page that is shown when a checkout is completed but we don't validate the completion of the checkout with the server yet. Due to spoofing concern I wouldn't rely on this as
```
mikkoh: Remove this new line and just call it `thank you page`.

EXAMPLES.md
```diff
@@ -217,6 +214,50 @@ cart.LineItems.Get(firstProduct, selectedOptions);
...
+```
+
+**Caveats**
+
+There are a few things we're still working out for the web checkout experience:
+
+1. Handling HTTP errors gracefully. When the user loses their internet connection there is currently no way to recover.
+2. Validating the web checkout purchase with results from the API. The completion callback is invoked when we detect that the user has navigated to the
+   thank\_you page. This is the page that is shown when a checkout is completed but we don't validate the completion of the checkout with the server yet. Due to spoofing concern I wouldn't rely on this as 
+   knowledge of a completely validated purchase _yet_.
```
mikkoh: Do we need a newline here?

mikkoh:
At some point we should add back the documentation how to get a web link. But yes this method to checkout should be more important.

sleroux:
@dengjeffrey I'll probably add more information around errors when I add the HTTP error PR.

EXAMPLES.md
```diff
@@ -217,6 +214,47 @@ cart.LineItems.Get(firstProduct, selectedOptions);
 cart.LineItems.Delete(firstProduct, selectedOptions);
 ```
 
+### Native web view checkout
+
+After creating an instance of `Cart` and adding items to it, you can use the `CheckoutWithNativeWebView` method to
```
mikkoh: Later on we should maybe mention for iOS and Android. Since potentially someone maybe hitting another platform.

PR #146, opened 2017-07-05
[Feature] Native checkout creation

dengjeffrey:
+ Implements Payment session creation
+ Adds currencyCode, and Allow For Partial address to checkout query
+ Splits Email and Shipping query, since email is not returned on a partial address 
+ Changes JSON class to override `ToString` instead of using `ToJsonString`


Assets/Editor/MockLoaderCheckouts.cs
```diff
@@ -1,19 +1,19 @@
 namespace Shopify.Tests {
     public class MockLoaderCheckouts : BaseMockLoader {
         // The following query is for a straightforward checkout create
-        private const string CREATE_QUERY = @"mutation{checkoutCreate (input:{lineItems:[{quantity:33,variantId:""Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0VmFyaWFudC8yMDc1NjEyOTE1NQ==""},{quantity:2,variantId:""Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0VmFyaWFudC8yMDc1NjEyOTM0Nw==""}]}){checkout {id webUrl requiresShipping subtotalPrice totalTax totalPrice ready lineItems (first:250){edges {node {id variant {id }}cursor }pageInfo {hasNextPage }}}userErrors {field message }}}";
+        private const string CREATE_QUERY = @"mutation{checkoutCreate (input:{allowPartialAddresses:true,lineItems:[{quantity:33,variantId:""Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0VmFyaWFudC8yMDc1NjEyOTE1NQ==""},{quantity:2,variantId:""Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0VmFyaWFudC8yMDc1NjEyOTM0Nw==""}]}){checkout {id webUrl currencyCode requiresShipping subtotalPrice totalTax totalPrice ready lineItems (first:250){edges {node {id variant {id }}cursor }pageInfo {hasNextPage }}}userErrors {field message }}}";
```
mikkoh: Even with `allowPartialAddresses` we'll get a user error?

dengjeffrey: No not anymore, we only receive user errors for the required fields that is needed for Partial Addresses 👍

scripts/generator/graphql_generator/csharp/Cart.CheckoutAccessors.cs.erb
```diff
@@ -5,6 +5,19 @@ namespace <%= namespace %> {
     using <%= namespace %>.GraphQL;
 
     public partial class Cart {
+
+        /// <summary>
+        /// Retrieve the currency code for the current checkout.
+        /// </summary>
+        /// <exception cref="NoCheckoutException">Calling this accessor before a checkout was created for the Cart. See Cart.GetWebCheckoutLink</exception>
+        public CurrencyCode GetCurrencyCode() {
```
mikkoh: I wonder if this should be:
```
public CurrencyCode? GetCurrencyCode()
```
Instead of thrown an exception we could return `null` if `CurrentCheckout == null`.

Another option would be to return `CurrencyCode.UNKNOWN` but I feel doing this would require a lot more domain knowledge than just simply returning `null`.

mikkoh: I've also wondered if we should just expose `CurrentCheckout` which would be another way to do this.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -170,17 +170,38 @@ namespace <%= namespace %> {
             }
         }
 
-        public void SetShippingAddressAndEmail(MailingAddressInput mailingAddressInput, string email, CompletionCallback callback) {
+        public void SetEmailAddress(string email, CompletionCallback callback) {
```
mikkoh: I wonder if we should still have a method that does `SetShippingAddressAndEmail`. I don't think this is crucial right now but it's something we might want to do.

Assets/Plugins/iOS/Shopify/Cart+ApplePay.swift
```diff
@@ -79,16 +79,27 @@ import PassKit
...
-            return nil
+            let data = jsonString.data(using: .utf8),
+            let object = try? JSONSerialization.jsonObject(with: data),
+            let stringCollection = object as? [String]
+            else {
+                return nil
+        }
+        
+        var itemCollection = [JSON]()
+        for string in stringCollection {
```
sleroux: Nit: Instead of using for/loop and another list you can use `flatMap`:

```swift
return stringCollection.flatMap { string in
    if let stringData  = string.data(using: .utf8), let itemObject = try? JSONSerialization.jsonObject(with: stringData), let itemJson   = itemObject as? JSON {                 
        itemCollection.append(itemJson)
    } else {
        return nil
    }
}
```

scripts/generator/graphql_generator/csharp/Cart.cs.erb
dengjeffrey: We can re-add it in when we encounter a case where we will need to use it

scripts/generator/graphql_generator/csharp/Cart.CheckoutAccessors.cs.erb
sleroux: +1 an exception seems harsh for this error. `null` is probably the most straightforward.

scripts/generator/graphql_generator/csharp/SDK/iOS/ApplePayEventReceiverBridge.cs.erb
```diff
@@ -0,0 +1,32 @@
+#if UNITY_IPHONE
```
sleroux: Move to using `UNITY_IOS` now that we know its deprecated?

scripts/generator/graphql_generator/csharp/SDK/iOS/ApplePayEventReceiverBridge.cs.erb
```diff
@@ -0,0 +1,32 @@
...
+namespace <%= namespace %>.SDK.iOS {
+    using System.Collections;
+    using System.Collections.Generic;
+    using UnityEngine;
+
+    public class ApplePayEventReceiverBridge : MonoBehaviour, IApplePayEventReceiver {
+
+        public IApplePayEventReceiver Receiver;
+
+        public ApplePayEventReceiverBridge(IApplePayEventReceiver receiver) {
```
sleroux: Since this is a MonoBehaviour that is created using Unity, will we ever hit this constructor? If not then we might as well remove it.

mikkoh: Since this is instantiated via `AddComponent` this constructor is never called. We should 🔥

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.cs.erb
```diff
@@ -17,12 +18,13 @@ namespace <%= namespace %>.SDK.iOS {
...
 
         [DllImport ("__Internal")]
-        private static extern bool _CreateApplePaySession(string merchantID, string countryCode, string currencyCode, string unityDelegateObjectName, string serializedSummaryItems, string serializedShippingMethods, bool requiringShipping);
+        private static extern bool _CreateApplePaySession(string unityDelegateObjectName, string merchantID, string countryCode, string currencyCode, string serializedSummaryItems, string serializedShippingMethods, bool requiringShipping);
 
         [DllImport ("__Internal")]
         private static extern void _PresentApplePayAuthorization();
 
         private Cart CurrentCart;
+        private ApplePayEventReceiverBridge ApplePayReceiver;
```
mikkoh: Since the real receiver is `this` we should name this `ApplePayEventBridge`.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.cs.erb
```diff
@@ -66,11 +68,57 @@ namespace <%= namespace %>.SDK.iOS {
...
+            var shippingMethods = new List<ShippingMethod>();
+
+            try {
+                var availableShippingRates = CurrentCart.GetShippingRates();
+
+                foreach (var shippingRate in availableShippingRates) {
+                    shippingMethods.Add(new ShippingMethod(shippingRate.title(), shippingRate.price().ToString(), shippingRate.handle()));
+                }
+            } catch (Exception e) {
+                throw new Exception("Attempted to access AvailableShippingRates when CurrentCheckout has no property named AvailableShippingRates", e);
```
sleroux: Is there an existing exception that is more specific we can use here or define ourselves instead of throwing a generic exception?

scripts/generator/graphql_generator/csharp/SDK/iOS/ApplePayEventReceiverBridge.cs.erb
mikkoh: 👍

Assets/Editor/MockLoaderCheckouts.cs
mikkoh: Great news.

scripts/generator/graphql_generator/csharp/Cart.CheckoutAccessors.cs.erb
mikkoh: My only worry is that we end up needing to expose more and more of `CurrentCheckout`.

dengjeffrey: Exposing `CurrentCheckout` might mean we also need to expose `CheckoutSave`. Perhaps this is worth revisiting later? 

Making it return `null` might mislead the implementor to use these calls haphazardly, it should the programmer's responsibility to ensure these calls are made after a checkout has been created.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -29,16 +30,35 @@ namespace <%= namespace %>.SDK.iOS {
...
 
-            List<SummaryItem> summaryItems = null;
-            List<ShippingMethod> shippingMethods = null;
+            Action respondWithSuccess = () => {
+                message.Respond(new ApplePayEventResponse(ApplePayAuthorizationStatus.Success, GetSummaryItems(), GetShippingMethods()).ToString());
+            };
+
+            Action<ShopifyError> respondToError = (ShopifyError error) => {
+                switch(error.Type) {
+                case ShopifyError.ErrorType.GraphQL:
```
mikkoh: 🔥 since this method will never receive a `ShopifyError.ErrorType.GraphQL`

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.cs.erb
dengjeffrey: this is a `private` method used internally, but it rethrows `NoQueryException` with a more detailed message. Does it need to be more explicit?

Would this be better: 
```
catch (NoQueryException e) {
    throw new NoQueryException("Attempted to access AvailableShippingRates when CurrentCheckout has no property named AvailableShippingRates", e)
}

File: scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb [hunk 10]
```diff
@@ -29,16 +30,35 @@ namespace <%= namespace %>.SDK.iOS {
...
+                case ShopifyError.ErrorType.HTTP:
+                    message.Respond(new ApplePayEventResponse(ApplePayAuthorizationStatus.Failure).ToString());
+                    break;
+                case ShopifyError.ErrorType.UserError:
+                    var userErrors = CurrentCart.UserErrors;
+
+                    if (userErrors.Count > 0) {
+                        var firstUserError = userErrors[0];
```
mikkoh: I think it's dangerous to expect the first user error to be related shipping.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
dengjeffrey: It will because we might receive a `GraphQL` error from `SetShippingAddress `

mikkoh: 👍

mikkoh: How would we handle this? Would it be similar to how `ShopifyError.ErrorType.HTTP` is handled?

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -64,34 +84,27 @@ namespace <%= namespace %>.SDK.iOS {
...
-            var shippingMethods = new List<ShippingMethod>();
-
-            try {
-                var availableShippingRates = CurrentCart.GetShippingRates();
-
-                foreach (var shippingRate in availableShippingRates) {
-                    shippingMethods.Add(new ShippingMethod(shippingRate.title(), shippingRate.price().ToString(), shippingRate.handle()));
+        // We only receive a partial shipping address before the user has authenticated
+        // City, State, Zip, Country
+        // So we will receive user errors from our GraphQL query, even though it was a success
```
mikkoh: Again just a reminder we should bring this up with #graphql-checkout

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.cs.erb
sleroux: Ah that's fine then 👍

scripts/generator/graphql_generator/csharp/Cart.CheckoutAccessors.cs.erb
mikkoh: I'm looking at the data it's accessing. It's all stuff the end user might want to access. To be honest when I look at it for instance:
```
CurrentCheckout.shippingLine().price()
```
Could be very confusing to the end user to get shipping totals.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
dengjeffrey: This case should fall through to `ShopifyError.ErrorType.HTTP`, which is just a failure to complete `ApplePay`

mikkoh: D'oh. I was thinking you were going to implement something special and this was placeholder 👍

scripts/generator/graphql_generator/csharp/SDK/Serializable.cs.erb
```diff
@@ -6,7 +6,7 @@ namespace <%= namespace %>.SDK {
     using UnityEngine;
 
     public class Serializable {
-        public string ToJsonString() {
+        override public string ToString() {
```
sleroux: As part of the JSON refactor work, I'm moving `Serializable` to be an interface so we won't be able to override `ToString`. It would also be odd for implementors of `Serializable` to need to override `ToString`. Could we change this back to using `toJsonString`?

dengjeffrey: Sure

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.cs.erb
```diff
@@ -66,11 +68,62 @@ namespace <%= namespace %>.SDK.iOS {
...
+            var checkout = CurrentCart.CurrentCheckout;
+            var shippingMethods = new List<ShippingMethod>();
+
+            try {
+                var availableShippingRates = checkout.availableShippingRates().shippingRates();
+
+                foreach (var shippingRate in availableShippingRates) {
+                    shippingMethods.Add(new ShippingMethod(shippingRate.title(), shippingRate.price().ToString(), shippingRate.handle()));
+                }
+            } catch (Exception e) {
```
mikkoh: If `title` `price` or `handle` were not queried then it would be caught here also.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -29,16 +30,44 @@ namespace <%= namespace %>.SDK.iOS {
...
+                            }
+
+                            statusToReturn = status;
+                        }
+                        message.Respond(new ApplePayEventResponse(statusToReturn, GetSummaryItems()).ToString());
+                    } else {
+                        message.Respond(new ApplePayEventResponse(ApplePayAuthorizationStatus.Failure).ToString());
+                    }
+                    break;
+                }
```
mikkoh: Might be good to do at the top of the method:
`ApplePayEventResponse response = null;`

Then at the bottom:
`message.Respond(response.ToString());`

Instead of having multiple:
`message.Respond`

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -29,16 +30,44 @@ namespace <%= namespace %>.SDK.iOS {
...
+                case ShopifyError.ErrorType.HTTP:
+                    message.Respond(new ApplePayEventResponse(ApplePayAuthorizationStatus.Failure).ToJsonString());
+                    break;
+                case ShopifyError.ErrorType.UserError:
+                    var userErrors = CurrentCart.UserErrors;
+
+                    if (userErrors.Count > 0) {
+                        ApplePayAuthorizationStatus statusToReturn = ApplePayAuthorizationStatus.Failure;
+                        foreach (var userError in userErrors) {
+                            var status = AuthorizationStatusForPreliminaryShippingUserError(userError);
```
mikkoh: Why not just do `statusToReturn = AuthorizationStatusForPreliminaryShippingUserError(userError);`

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -29,16 +30,44 @@ namespace <%= namespace %>.SDK.iOS {
...
+            Action<ShopifyError> respondToError = (ShopifyError error) => {
+                switch(error.Type) {
+                case ShopifyError.ErrorType.GraphQL:
+                case ShopifyError.ErrorType.HTTP:
+                    message.Respond(new ApplePayEventResponse(ApplePayAuthorizationStatus.Failure).ToJsonString());
+                    break;
+                case ShopifyError.ErrorType.UserError:
+                    var userErrors = CurrentCart.UserErrors;
+
+                    if (userErrors.Count > 0) {
```
mikkoh: Is there a case where a `ShopifyError.ErrorType.UserError` would be created however `userErrors.Count == 0`

dengjeffrey: no, just being really defensive here

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -29,16 +30,44 @@ namespace <%= namespace %>.SDK.iOS {
...
-            List<SummaryItem> summaryItems = null;
-            List<ShippingMethod> shippingMethods = null;
+            Action respondWithSuccess = () => {
+                message.Respond(new ApplePayEventResponse(ApplePayAuthorizationStatus.Success, GetSummaryItems(), GetShippingMethods()).ToJsonString());
+            };
+
+            Action<ShopifyError> respondToError = (ShopifyError error) => {
+                ApplePayEventResponse response;
+
+                switch(error.Type) {
```
mikkoh: Since we just have one `case` and `default` lets change to a `if else`

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -29,16 +30,44 @@ namespace <%= namespace %>.SDK.iOS {
...
+                case ShopifyError.ErrorType.UserError:
+                    var userErrors = CurrentCart.UserErrors;
+
+                    if (userErrors.Count > 0) {
+                        ApplePayAuthorizationStatus statusToReturn = ApplePayAuthorizationStatus.Failure;
+                        foreach (var userError in userErrors) {
+                            statusToReturn = AuthorizationStatusForPreliminaryShippingUserError(userError);
```
mikkoh: Let's change `AuthorizationStatusForPreliminaryShippingUserError` to be `GetStatusFromUserErrors`.

dengjeffrey:
👍 ??

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.cs.erb
```diff
@@ -17,12 +18,13 @@ namespace <%= namespace %>.SDK.iOS {
         private static extern void _ShowApplePaySetup();
 
         [DllImport ("__Internal")]
-        private static extern bool _CreateApplePaySession(string merchantID, string countryCode, string currencyCode, string unityDelegateObjectName, string serializedSummaryItems, string serializedShippingMethods, bool requiringShipping);
+        private static extern bool _CreateApplePaySession(string unityDelegateObjectName, string merchantID, string countryCode, string currencyCode, string serializedSummaryItems, string serializedShippingMethods, bool requiringShipping);
```
sleroux: Interesting. I noticed that in this PR the method signature of the iOS call didn't change. Does that mean this is a fix to match the iOS line?

Assets/Plugins/iOS/Shopify/Cart+ApplePay.swift
```diff
@@ -79,16 +79,23 @@ import PassKit
...
         else {
             return nil
         }
         
-        return items
+        return stringCollection.flatMap { string in
+            if let stringData  = string.data(using: .utf8),
+                let itemObject = try? JSONSerialization.jsonObject(with: stringData),
+                let itemJson   = itemObject as? JSON {
+
```
sleroux: Nit: Remove extra line?

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.cs.erb
dengjeffrey: yes this is a fix to match the iOS line

PR #147, opened 2017-07-05
[Fix] Refactors JsonUtility to use MiniJSON

sleroux:
* Replaces JsonUtility uses of Deserialize with MiniJSON

We can probably follow this up with a fancier generic-based implementation like JsonUtility does in the future.

### DIFF
--- a/Assets/UnityTestTools/NativeTesting/Native/NativeMessageTester.cs
+++ b/Assets/UnityTestTools/NativeTesting/Native/NativeMessageTester.cs
@@ -1,6 +1,8 @@
 namespace Shopify.Tests {
+    using System.Collections.Generic;
     using UnityEngine;
     using Shopify.Unity.SDK;
+    using Shopify.Unity.MiniJSON;
 
     public class NativeMessageTester: NativeMessage {
 
@@ -8,8 +10,11 @@ namespace Shopify.Tests {
 
         new public static NativeMessage CreateFromJSON(string jsonString)
         {
-            LastMessage = JsonUtility.FromJson<NativeMessage>(jsonString);
+            var dict = (Dictionary<string, object>) Json.Deserialize(jsonString);
+            LastMessage = new NativeMessage((string) dict["Identifier"], (string) dict["Content"]);
             return LastMessage;
         }
+
+        public NativeMessageTester(string identifier, string content) : base(identifier, content) {}
     }
 }
--- a/Assets/UnityTestTools/NativeTesting/iOS/ApplePayEventTester.cs
+++ b/Assets/UnityTestTools/NativeTesting/iOS/ApplePayEventTester.cs
@@ -1,4 +1,4 @@
-﻿#if UNITY_IPHONE 
+﻿#if UNITY_IOS
 namespace Shopify.Tests.iOS {
     using System.Collections;
     using System.Collections.Generic;
--- a/scripts/generator/graphql_generator/csharp/SDK/NativeMessage.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/NativeMessage.cs.erb
@@ -1,10 +1,11 @@
 #if !SHOPIFY_MONO_UNIT_TEST
 namespace <%= namespace %>.SDK {
     using System;
-    using UnityEngine;
+    using System.Collections.Generic;
     using System.Runtime.InteropServices;
+    using UnityEngine;
+    using <%= namespace %>.MiniJSON;
 
-    [Serializable]
     public class NativeMessage {
 
         [DllImport ("__Internal")]
@@ -13,9 +14,14 @@ namespace <%= namespace %>.SDK {
         public string Identifier;
         public string Content;
 
-        public static NativeMessage CreateFromJSON(string jsonString)
-        {
-            return JsonUtility.FromJson<NativeMessage>(jsonString);
+        public static NativeMessage CreateFromJSON(string jsonString) {
+            var dict = (Dictionary<string, object>) Json.Deserialize(jsonString);
+            return new NativeMessage((string) dict["Identifier"], (string) dict["Content"]);
+        }
+
+        public NativeMessage(string identifier, string content) {
+            Identifier = identifier;
+            Content = content;
         }
 
         public void Respond(string message) {
--- a/scripts/generator/graphql_generator/csharp/SDK/Serializable.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/Serializable.cs.erb
@@ -1,14 +1,14 @@
-#if !SHOPIFY_MONO_UNIT_TEST
 namespace <%= namespace %>.SDK {
     using System;
     using System.Collections;
     using System.Collections.Generic;
-    using UnityEngine;
+    using <%= namespace %>.MiniJSON;
 
-    public class Serializable {
+    public abstract class Serializable {
         public string ToJsonString() {
-            return JsonUtility.ToJson(this);
+            return Json.Serialize(ToJson());
         }
+
+        public abstract object ToJson();
     }
 }
-#endif
--- a/scripts/generator/graphql_generator/csharp/SDK/iOS/ApplePayEventResponse.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/iOS/ApplePayEventResponse.cs.erb
@@ -1,9 +1,8 @@
 #if UNITY_IPHONE
 namespace <%= namespace %>.SDK.iOS {
+    using System;
     using System.Collections;
     using System.Collections.Generic;
-    using System;
-    using UnityEngine;
 
     public class ApplePayEventResponse: Serializable {
         public string AuthorizationStatus;
@@ -14,6 +13,25 @@ namespace <%= namespace %>.SDK.iOS {
             AuthorizationStatus = authorizationStatus.ToString("G");
             SummaryItems = summaryItems ?? new List<SummaryItem>();
             ShippingMethods = shippingMethods ?? new List<ShippingMethod>();
+        }
+
+        public override object ToJson() {
+            var dict = new Dictionary<string, object>();
+            dict["AuthorizationStatus"] = AuthorizationStatus;
+
+            var summaryItemsJSON = new List<Dictionary<string, object>>();
+            foreach (var item in SummaryItems) {
+                summaryItemsJSON.Add((Dictionary<string, object>) item.ToJson());
+            }
+            dict["SummaryItems"] = summaryItemsJSON;
+
+            var shippingMethodsJSON = new List<Dictionary<string, object>>();
+            foreach (var method in ShippingMethods) {
+                shippingMethodsJSON.Add((Dictionary<string, object>) method.ToJson());
+            }
+            dict["ShippingMethods"] = shippingMethodsJSON;
+
+            return dict;
         }
     }
 }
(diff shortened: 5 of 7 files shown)


Assets/UnityTestTools/NativeTesting/Native/NativeMessageTester.cs
```diff
@@ -1,15 +1,19 @@
...
 
     public class NativeMessageTester: NativeMessage {
 
         public static NativeMessage LastMessage;
 
         new public static NativeMessage CreateFromJSON(string jsonString)
         {
-            LastMessage = JsonUtility.FromJson<NativeMessage>(jsonString);
-            return LastMessage;
+            var dict = (Dictionary<string, object>) Json.Deserialize(jsonString);
```
mikkoh: Shouldn't this still update `LastMessage`. Something like:

```
LastMessage = NativeMessage. CreateFromJSON(jsonString);
return LastMessage;
```

scripts/generator/graphql_generator/csharp/SDK/Serializable.cs.erb
```diff
@@ -4,10 +4,11 @@ namespace <%= namespace %>.SDK {
     using System.Collections;
     using System.Collections.Generic;
     using UnityEngine;
+    using <%= namespace %>.MiniJSON;
 
     public class Serializable {
         public string ToJsonString() {
-            return JsonUtility.ToJson(this);
+            return Json.Serialize(this);
```
mikkoh: `MiniJSON.Serialize` expects the passed in param to implement `IDictionary` or `IList`.

mikkoh: `MiniJSON` really should have overloaded `Serialize`

scripts/generator/graphql_generator/csharp/SDK/NativeMessage.cs.erb
```diff
@@ -1,8 +1,10 @@
 #if !SHOPIFY_MONO_UNIT_TEST
 namespace <%= namespace %>.SDK {
     using System;
-    using UnityEngine;
+    using System.Collections.Generic;
     using System.Runtime.InteropServices;
+    using UnityEngine;
+    using <%= namespace %>.MiniJSON;
 
     [Serializable]
```
mikkoh: 🔥

scripts/generator/graphql_generator/csharp/SDK/Serializable.cs.erb
sleroux: If that's the case then it doesn't make sense for there to be a `Serializable` subclass anymore since we don't want to implement `IDictionary`/`IList` here. It makes more sense to 🔥 this class and have `ApplePayEventResponse` implement `IDictionary` and replace the `ToJsonString()` calls to just `Json.Serialize(applePayEventResponseInstance)`.

mikkoh: I think we need to make `Serializable` an interface.

scripts/generator/graphql_generator/csharp/SDK/NativeMessage.cs.erb
```diff
@@ -13,14 +15,19 @@ namespace <%= namespace %>.SDK {
...
+        public static NativeMessage CreateFromJSON(string jsonString) {
+            var dict = (Dictionary<string, object>) Json.Deserialize(jsonString);
+            return new NativeMessage((string) dict["Identifier"], (string) dict["Content"]);
         }
 
         public void Respond(string message) {
             _RespondToNativeMessage(Identifier, message);
         }
+
+        public NativeMessage(string identifier, string content) {
```
richardmonette: By convention, would we put the constructor first?

mikkoh: 👍

scripts/generator/graphql_generator/csharp/SDK/Serializable.cs.erb
richardmonette: Is there  a situation where `Json.Serialize(applePayEventResponseInstance)` doesn't fit? Seems more explicit and clearer than having to have an interface around this essentially one line of code.

sleroux:
@richardmonette @mikkoh Updated

Assets/UnityTestTools/NativeTesting/Native/NativeMessageTester.cs
```diff
@@ -1,15 +1,20 @@
...
 
         new public static NativeMessage CreateFromJSON(string jsonString)
         {
-            LastMessage = JsonUtility.FromJson<NativeMessage>(jsonString);
+            var dict = (Dictionary<string, object>) Json.Deserialize(jsonString);
+            LastMessage = new NativeMessage((string) dict["Identifier"], (string) dict["Content"]);
             return LastMessage;
         }
+
+        public NativeMessageTester(string identifier, string content) : base(identifier, content) {}
```
richardmonette: Sorry, this is the same nitpick as the other constructor

mikkoh: Constructors should follow statics.

scripts/generator/graphql_generator/csharp/SDK/iOS/ApplePayEventResponse.cs.erb
```diff
@@ -15,6 +15,14 @@ namespace <%= namespace %>.SDK.iOS {
             SummaryItems = summaryItems ?? new List<SummaryItem>();
             ShippingMethods = shippingMethods ?? new List<ShippingMethod>();
         }
+
+        public string ToJsonString() {
+            var dict = new Dictionary<string, object>();
+            dict["SummaryItems"] = SummaryItems;
```
mikkoh: It's unfortunate but it's not as easy as this. `SummaryItems` would need to be turned into a `List<Dictionary<string, object>`.

scripts/generator/graphql_generator/csharp/SDK/iOS/SummaryItem.cs.erb
```diff
@@ -1,7 +1,9 @@
 #if UNITY_IPHONE
 namespace <%= namespace %>.SDK.iOS {
+    using System.Collections.Generic;
     using System.Collections;
     using System;
+    using <%= namespace %>.MiniJSON;
 
     [Serializable]
```
mikkoh: 🔥

scripts/generator/graphql_generator/csharp/SDK/iOS/ApplePayEventResponse.cs.erb
```diff
@@ -15,6 +15,14 @@ namespace <%= namespace %>.SDK.iOS {
...
+            dict["AuthorizationStatus"] = AuthorizationStatus;
+            dict["ShippingMethods"] = ShippingMethods;
```
mikkoh: Same thing with `ShippingMethods` it needs to be a list of `List<Dictionary<string, object>>`.

scripts/generator/graphql_generator/csharp/SDK/Serializable.cs.erb
```diff
@@ -1,14 +1,9 @@
...
     using System.Collections;
     using System.Collections.Generic;
-    using UnityEngine;
 
-    public class Serializable {
-        public string ToJsonString() {
-            return JsonUtility.ToJson(this);
-        }
+    public interface Serializable {
+        string ToJsonString();
```
mikkoh: I wonder if we should add in something like:
`object ToJson();`
`ToJsonString` could use the result of `ToJson`

Assets/UnityTestTools/NativeTesting/iOS/ApplePayEventTester.cs
```diff
@@ -1,5 +1,4 @@
-﻿#if UNITY_IPHONE
```
sleroux: @dengjeffrey Seems that with this #ifdef check the tests won't work but without it works.

mikkoh: What if it's `UNITY_IOS` since `UNITY_IPHONE` is deprecated?

scripts/generator/graphql_generator/csharp/SDK/iOS/ApplePayEventResponse.cs.erb
```diff
@@ -15,6 +15,29 @@ namespace <%= namespace %>.SDK.iOS {
...
+
+            var summaryItemsJSON = new List<Dictionary<string, object>>();
+            foreach (var item in SummaryItems) {
+                summaryItemsJSON.Add((Dictionary<string, object>) item.ToJson());
+            }
+            dict["SummaryItems"] = summaryItemsJSON;
+
+            var shippingMethodsJSON = new List<Dictionary<string, object>>();
+            foreach (var method in ShippingMethods) {
+                shippingMethodsJSON.Add((Dictionary<string, object>) method.ToJson());
```
mikkoh: `ShippingMethod.cs.erb` also needs to implement `ToJson`.

sleroux: Heh yup just ran into this when running the tests 🙉

scripts/generator/graphql_generator/csharp/SDK/Serializable.cs.erb
mikkoh: Since all implementations of `Serializable` simply do:
```
public string ToJsonString() {
    return Json.Serialize(ToJson());
}
```
We should just make `Serializable` be abstract and add the above here.

sleroux: Good call 👍

Assets/UnityTestTools/NativeTesting/iOS/ApplePayEventTester.cs
sleroux: Yup - `UNITY_IOS` makes it work 👍

sleroux:
@mikkoh @richardmonette OK - should be good for another pass.

scripts/generator/graphql_generator/csharp/SDK/NativeMessage.cs.erb
```diff
@@ -13,9 +14,14 @@ namespace <%= namespace %>.SDK {
...
 
-        public static NativeMessage CreateFromJSON(string jsonString)
-        {
-            return JsonUtility.FromJson<NativeMessage>(jsonString);
+        public NativeMessage(string identifier, string content) {
+            Identifier = identifier;
+            Content = content;
+        }
+
+        public static NativeMessage CreateFromJSON(string jsonString) {
```
mikkoh: Nit: `CreateFromJSON` should be above constructor.

scripts/generator/graphql_generator/csharp/SDK/NativeMessage.cs.erb
```diff
@@ -1,10 +1,11 @@
 #if !SHOPIFY_MONO_UNIT_TEST
 namespace <%= namespace %>.SDK {
     using System;
-    using UnityEngine;
+    using System.Collections.Generic;
     using System.Runtime.InteropServices;
+    using UnityEngine;
```
dengjeffrey: Do we still need this? now that we don't use `JsonUtility`

sleroux: I think we still need it for the DllImport statement below.

dengjeffrey: `System.Runtime.InteropServices` is for DLLImport

PR #151, opened 2017-07-05
[FIX] Publish Mobile Tests

mikkoh:
Currently when you run `scripts/publish.sh` it copies over files required for testing the iOS Native Extension. This PR basically copies and deletes these files when they're needed.

Another thing we could do is delete the files required for testing and then `git checkout` them after the publish is complete. Thoughts?


scripts/publish.sh
```diff
@@ -14,6 +15,9 @@ check() {
     fi
 }
 
+# delete all files required for native testing
+delete_native_tests
```
sleroux: So I guess before the .unitypackage would include all of our tests because Unity will pull in everything from the Assets/Plugins folder right?

mikkoh: Correct. @richardmonette looked into adding files to ignore.

scripts/native_tests.sh
```diff
@@ -0,0 +1,13 @@
...
+# will copy all files needed for testing native extenstions
+copy_native_tests() {
+    mkdir -p $SCRIPTS_ROOT/native_tests/Assets/Plugins/iOS/Shopify/BuyTests/
+    cp -r $SCRIPTS_ROOT/native_tests/Assets/Plugins/iOS/Shopify/ $PROJECT_ROOT/Assets/Plugins/iOS/Shopify/
+}
+
+# will delete all files needed for testing native extenstions
+delete_native_tests() {
+    rm -rf $PROJECT_ROOT/Assets/Plugins/iOS/Shopify/BuyTests/
+    rm -f $PROJECT_ROOT/Assets/Plugins/iOS/Shopify/Unity-iPhone-Tests-Bridging-Header.h
```
dengjeffrey: Nit: Should probably quote where we use `$` incase `PROJECT_ROOT` has spaces
`"$PROJECT_ROOT/Assets/Plugins/iOS/Shopify/Unity-iPhone-Tests-Bridging-Header.h"`

mikkoh:
I just changed to reset files by using checkout instead cause it's less aggressive.

scripts/native_tests.sh
mikkoh: 👍

dengjeffrey:
Should the cleanup only happen before calling `publish.sh`? What if we wanted to write tests and test them on the go?

PR #153, opened 2017-07-06
[Feature] Combines SetEmailUpdate/SetShippingAddress/SetShippingLine into SetFinalCheckoutFields method

sleroux:
* Adds `SetFinalCheckoutFields` method for combining the 3 update query calls into one.
* Adds `ShippingFields` to enclose the shipping address and identifier since the update query either has both or none of those properties.

I'll follow this PR up with tests on TestCart.cs

### DIFF
--- a/scripts/generator/graphql_generator/csharp/Cart.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/Cart.cs.erb
@@ -23,6 +23,11 @@ namespace <%= namespace %> {
     public struct ShippingFields {
         public MailingAddressInput ShippingAddress;
         public string ShippingIdentifier; 
+
+        public ShippingFields(MailingAddressInput shippingAddress, string shippingIdentifier) {
+            ShippingIdentifier = shippingIdentifier;
+            ShippingAddress = shippingAddress;
+        }
     }
     
     /// <summary>
--- a/scripts/generator/graphql_generator/csharp/Cart.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/Cart.cs.erb
@@ -246,7 +246,7 @@ namespace <%= namespace %> {
                     }
 
                     if (UpdateState(response.checkoutEmailUpdate().checkout(), userErrors)) {
-                        PollCheckoutAndUpdate(PollCheckoutAvailableShippingRatesReady, callback);
+                        PollCheckoutAndUpdate(PollCheckoutReady, callback);
                     } else {
                         HandleUserError(callback);
                     }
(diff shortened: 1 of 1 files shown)


scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -211,6 +225,35 @@ namespace <%= namespace %> {
...
+                    callback(error);
+                } else {
+                    var userErrors = response.checkoutShippingAddressUpdate().userErrors();
+                    if (shippingFields.HasValue) {
+                        userErrors.AddRange(response.checkoutShippingLineUpdate().userErrors());
+                        userErrors.AddRange(response.checkoutEmailUpdate().userErrors());
+                    }
+
+                    if (UpdateState(response.checkoutEmailUpdate().checkout(), userErrors)) {
+                        PollCheckoutAndUpdate(PollCheckoutAvailableShippingRatesReady, callback);
```
dengjeffrey: `PollCheckoutAndUpdate(PollCheckoutReady, callback)` instead because we want to wait for `Checkout.ready()`

sleroux:
@dengjeffrey Updated

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -211,6 +225,35 @@ namespace <%= namespace %> {
             });
         }
 
+        public void SetFinalCheckoutFields(string email, ShippingFields? shippingFields, CompletionCallback callback) {
```
mikkoh: I'm a little bit worried about the amount of public methods we're adding on `Cart`. Like an end user would probably not use this method but `Cart` by design is something meant for the end user to interactive with.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -16,6 +16,20 @@ namespace <%= namespace %> {
     public class NoMatchingVariantException : Exception {
         public NoMatchingVariantException(string message) : base(message) {}
     }
+
+    /// <summary>
+    /// Wraps around a shipping address and shipping identifier. Used for sending the final checkout fields in one GraphQL query.
+    /// </summary>
+    public struct ShippingFields {
+        public MailingAddressInput ShippingAddress;
```
mikkoh: Because `MailingAddressInput` is used here it'll be very hard for the end user to instantiate a `ShippingFields`

dengjeffrey: This does seem messy but it's due to the bigger problem of having the cart handle the state of the checkout. This struct is public because it needs to be used in `iOSNativeCheckout`

mikkoh: I wonder if it should exist in `iOSNativeCheckout` and `SetFinalCheckoutFields` should also live there also.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
mikkoh: I wonder if things like this should exist on `iOSNativeCheckout`

dengjeffrey:
Yep, I'm fine with this now. We need to spend some time thinking of a better abstraction

dengjeffrey:
:shipit:

Issue #136, opened 2017-06-30
Handle HTTP errors more gracefully when performing a web view checkout


sleroux:
At the moment, if the user encounters a loss of internet or some other kind of HTTP error they have no way to recover in the checkout process. A couple of options include:

1. Whenever the web view fails to load a web page, we can bubble that up to the Unity developer through the `CheckoutDidFail` callback. 

2. Handle the HTTP error internally on the web view and build in a recovery mechanism.

Option (1) would give the Unity developer more control over what is happening in the checkout flow but would add more work on their side to implement the recovery mechanism. Option (2) would isolate any HTTP errors into the web flow and have them be self handled (maybe through a locally-injected HTML page with a refresh button and text). We could also combine both options.

At the moment I'm leaning towards option (2) so the Unity developer doesn't have to worry about it and keeps the API more simple.

sleroux:
Here's an example of handling the HTTP errors within the web view (Option 2):

<img width="372" alt="screen shot 2017-07-06 at 9 41 53 am" src="https://user-images.githubusercontent.com/434322/27913868-709f7352-622f-11e7-875d-f156f35705bf.png">

This is implemented by loading a local error HTML page into the web view when we fail to load the page. Tapping Reload will perform a browser refresh on the URL they were trying to access. I've also pulled in the Polaris styles to make it feel more Shopify-y.

@mikkoh Do you think this approach will suffice for handling network failures? We could also expose the HTML template to developers in the future if they want to customize it in some way.

I'm also assuming that failing to load a page is the extent of HTTP errors we would receive. Is there any other HTTP errors you can see us running into that isn't failing to load?

mikkoh:
So here are some of my thoughts.

I like the approach of showing this page when there is an error. It would be cool to call the failure callback also so that the developer could handle it. But if we do that we also have to have the ability to kill the webview. Currently we don't have that capability.

I wonder about the copy. Shopify has a specific style of writing copy. Let me try to find someone who could give a look at it.

sleroux:
> It would be cool to call the failure callback also so that the developer could handle it.

Yea and even for logging purposes so the developer knows that the network was down.

> I wonder about the copy. Shopify has a specific style of writing copy. Let me try to find someone who could give a look at it.

I was wondering the same thing. The message seems a bit too 'developer' 😉

mikkoh:
Chatted with @biggeidea about the copy.

Here's what he said:
```
Hey. We usually need the following for an error message:

[...]

PR #156, opened 2017-07-06
[Feature] Handle WebView loading failures during Web Checkout

sleroux:
Adds errorPage.html/polaris-1.0.2.css files which are loaded into the web view when we've encountered a web view error such as loss of connection.

### DIFF
--- a/Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebViewController.swift
+++ b/Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebViewController.swift
@@ -34,7 +34,6 @@ class WebViewController: UIViewController {
     weak var delegate: WebCheckoutDelegate?
     
     fileprivate(set) var startURL: URL
-    fileprivate var errorHTMLString: String
     fileprivate var lastGoodURL: URL
     
     fileprivate var contentView: WebContentView {
@@ -49,14 +48,6 @@ class WebViewController: UIViewController {
         self.startURL = url
         self.lastGoodURL = url
         self.delegate = delegate
-        
-        let errorPageFileURL = Bundle.main.url(forResource: "errorPage", withExtension: "html")!
-        let polarisCSSFileURL = Bundle.main.url(forResource: "polaris-1.0.2", withExtension: "css")!
-        
-        let errorPageString = try! String(contentsOf: errorPageFileURL)
-        let polarisCSSString = try! String(contentsOf: polarisCSSFileURL)
-        
-        errorHTMLString = errorPageString.replacingOccurrences(of: "<%= polaris_css %>", with: polarisCSSString)
         
         super.init(nibName: nil, bundle: nil)
     }
@@ -88,6 +79,28 @@ extension Checkout.WebViewController {
         webView.load(request)
     }
     
+    func generateErrorPage(message: String, retryURL: URL) -> URL {
+        // Instead of using webView.loadHTMLString, we generate a new HTML file and store on disk so 
+        // we can use webView.loadFileURL which allows us differentiate between an error page and a 
+        // web page. Using webView.loadHTMLString will set the url to about:blank which isn't strong enough
+        // to tell us that it's not an error page.
+        let errorPageFileURL = Bundle.main.url(forResource: "errorPage", withExtension: "html")!
+        let polarisCSSFileURL = Bundle.main.url(forResource: "polaris-1.0.2", withExtension: "css")!
+        
+        let errorPageString = try! String(contentsOf: errorPageFileURL)
+        let polarisCSSString = try! String(contentsOf: polarisCSSFileURL)
+        
+        var errorHTMLString = errorPageString.replacingOccurrences(of: "<%= polaris_css %>", with: polarisCSSString)
+        errorHTMLString = errorHTMLString.replacingOccurrences(of: "<%= retry_url %>", with: lastGoodURL.absoluteString)
+        errorHTMLString = errorHTMLString.replacingOccurrences(of: "<%= message %>", with: message)
+        
+        // Write the .html file to the temporary directory.
+        let tempDir = NSTemporaryDirectory()
+        let errorPageURL = URL(string: "file:///\(tempDir)/errorPage.html")!
+        try! errorHTMLString.write(to: errorPageURL, atomically: true, encoding: .utf8)
+        return errorPageURL
+    }
+    
     func didPressCancel() {
         delegate?.willDismiss { _ in
             self.dismiss(animated: true)
@@ -102,9 +115,7 @@ extension Checkout.WebViewController: WKNavigationDelegate {
     }
     
     public func webView(_ webView: WKWebView, didFailProvisionalNavigation navigation: WKNavigation!, withError error: Error) {
-        if (error as NSError).code == NSURLErrorNotConnectedToInternet {
-            loadErrorPage()
-        }
+        handleLoadError(error as NSError)
     }
     
     public func webView(_ webView: WKWebView, didCommit navigation: WKNavigation!) {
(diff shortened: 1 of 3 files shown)


sleroux:
FYI copy is still in the works: https://github.com/Shopify/unity-buy-sdk/issues/136

sleroux:
Here's what it looks like 👀

screen shot 2017-07-06 at 9 41 53 am

Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebViewController.swift
```diff
@@ -73,10 +75,32 @@ class WebViewController: UIViewController {
...
+    func generateErrorPage(message: String, retryURL: URL) -> URL {
+        // Instead of using webView.loadHTMLString, we generate a new HTML file and store on disk so 
+        // we can use webView.loadFileURL which allows us differentiate between an error page and a 
+        // web page. Using webView.loadHTMLString will set the url to about:blank which isn't strong enough
+        // to tell us that it's not an error page.
+        let errorPageFileURL = Bundle.main.url(forResource: "errorPage", withExtension: "html")!
+        let polarisCSSFileURL = Bundle.main.url(forResource: "polaris-1.0.2", withExtension: "css")!
+        
+        let errorPageString = try! String(contentsOf: errorPageFileURL)
+        let polarisCSSString = try! String(contentsOf: polarisCSSFileURL)
```
dengjeffrey: Can we store these resources, so that they are not read and loaded every single time we `generateErrorPage`?

Assets/Plugins/iOS/Shopify/errorPage.html
```diff
@@ -0,0 +1,46 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8" /> 
+<meta name="viewport" content="width=device-width,initial-scale=1.0,maximum-scale=1.0,user-scalable=no" />
+<link rel="stylesheet" href="https://sdks.shopifycdn.com/polaris/1.0.2/polaris.css" />
```
dengjeffrey: We don't need this right? If we are injecting the `CSS`

dengjeffrey: I wonder if instead of injecting the `CSS` we can link to the `CSS` using the bundle path

Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebViewController.swift
```diff
@@ -116,4 +149,18 @@ extension Checkout.WebViewController: WKNavigationDelegate {
...
+    private func handleLoadError(_ error: NSError) {
+        let errorPageURL: URL
+        switch error.code {
+        case NSURLErrorNotConnectedToInternet:
+            errorPageURL = generateErrorPage(message: "Looks like you don't have an internet connection. You won't be able to continue checking out until you do. Tap 'Reload' to try loading the page again.",
+                                             retryURL: lastGoodURL)
+        default:
+            errorPageURL =  generateErrorPage(message: "We've run into an unknown issue. Tap 'Reload' to try loading the page again.",
+                                              retryURL: lastGoodURL)
+        }
```
dengjeffrey: Looks like the switch statement is deciding which `message` to use. Would it be cleaner if `generateErrorPage` was used after the `switch`, while the `switch` just picks a `message`

Assets/Plugins/iOS/Shopify/errorPage.html
sleroux: I know with `loadHTMLString` you can pass in a `baseURL` but you can't with `loadURL`. You might be able to find the directory the resource is stored in someway but feels unsafe without being able to pass in the path.

sleroux:
Ill need to rebase this with the updated UI changes from https://github.com/Shopify/unity-buy-sdk/pull/164

Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebViewController.swift
```diff
@@ -44,9 +48,15 @@ class WebViewController: UIViewController {
...
     
     init(url: URL, delegate: WebCheckoutDelegate) {
-        self.url = url
+        self.startURL = url
+        self.lastGoodURL = url
         self.delegate = delegate
-        
+
+        let errorPageFileURL = Bundle.main.url(forResource: "errorPage", withExtension: "html")!
+        let polarisCSSFileURL = Bundle.main.url(forResource: "polaris-1.0.2", withExtension: "css")!
```
dbart01: The `main` bundle what you want here? I'm not sure what context this runs in but if this bundle is a dependency of an application bundle, your resources will come back `nil`. It's entirely possible that this will never be the case but it may be safer to use `Bundle(for: WebViewController.self)` instead. Your call.

Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebViewController.swift
```diff
@@ -73,10 +83,26 @@ class WebViewController: UIViewController {
...
+    func generateErrorPage(message: String, retryURL: URL) -> URL {
+        // Instead of using webView.loadHTMLString, we generate a new HTML file and store on disk so 
+        // we can use webView.loadFileURL which allows us differentiate between an error page and a 
+        // web page. Using webView.loadHTMLString will set the url to about:blank which isn't strong enough
+        // to tell us that it's not an error page.
+        var errorHTMLString = errorPageString.replacingOccurrences(of: "<%= polaris_css %>", with: polarisCSSString)
+        errorHTMLString = errorHTMLString.replacingOccurrences(of: "<%= retry_url %>", with: lastGoodURL.absoluteString)
+        errorHTMLString = errorHTMLString.replacingOccurrences(of: "<%= message %>", with: message)
+        
+        // Write the .html file to the temporary directory.
```
dbart01: I generally find that blocks of code requiring explicit comments to describe what they're doing, can greatly benefit from being abstracted into a dedicated, single-responsibility method that also provides self-documentation. The above can be written as:
```swift
func generateErrorPage(message: String, retryURL: URL) -> URL {
    let errorHTML = self.generateErrorHTML()
    return self.logErrorHTML(errorHTML)
}
```

Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebViewController.swift
```diff
@@ -73,10 +83,26 @@ class WebViewController: UIViewController {
...
+        let tempDir = NSTemporaryDirectory()
+        let errorPageURL = URL(string: "file:///\(tempDir)/errorPage.html")!
```
dbart01: There's already a `URL` initializer to do this for you:
```swift
let url = URL(fileURLWithPath: "/path/to/your/file")
```

Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebViewController.swift
```diff
@@ -73,10 +83,26 @@ class WebViewController: UIViewController {
...
+        try! errorHTMLString.write(to: errorPageURL, atomically: true, encoding: .utf8)
```
dbart01: How confident are you that this will *always* succeed?

Assets/Plugins/iOS/Shopify/Checkout/Checkout.WebViewController.swift
```diff
@@ -116,4 +151,17 @@ extension Checkout.WebViewController: WKNavigationDelegate {
     private func isThankYouPage(_ url: URL) -> Bool {
         return url.absoluteString.hasSuffix("/thank_you")
     }
+    
+    private func handleLoadError(_ error: NSError) {
+        let message: String
+        switch error.code {
```
dbart01: This `switch` might be better suited as an `extension` on `NSError`. Something like:
```swift
extension NSError {
    var networkErrorDescription: String {
        // swift statement
    }
}
```

sleroux:
If we're going with the `SFSafariViewController` route, we won't be able or need to show a custom error page for when the user has no internet connection. Closing this for now.

PR #157, opened 2017-07-06
[Feature] Apple Pay checkout token

dengjeffrey:
Requires #153, so I set to temporary merge into that branch

+ Implements checkout with token and error handling
+ Adds missed deserializations for Payment
+ Adds Simulator compile time flag

### Decisions
When handling Apple Pay errors, if more than one error is present such as `InvalidShippingContact` or `InvalidShippingAddress`, it doesn't matter which one we return, we can only return one at a time, so it is random to which one is returned

### DIFF
--- a/scripts/generator/graphql_generator/csharp/Cart.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/Cart.cs.erb
@@ -289,7 +289,7 @@ namespace <%= namespace %> {
             });
         }
 
-        private void CheckoutWithTokenizedPayment(TokenizedPaymentInput tokenizedPaymentInput, CompletionCallback callback) {
+        public void CheckoutWithTokenizedPayment(TokenizedPaymentInput tokenizedPaymentInput, CompletionCallback callback) {
             Action<Payment> pollPayment = (payment) => {
                 PollPaymentReady(payment.id(), (Payment newPayment, ShopifyError error) => {
                     if (error != null) {
--- a/scripts/generator/graphql_generator/csharp/SDK/iOS/PaymentDeserializer.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/iOS/PaymentDeserializer.cs.erb
@@ -54,6 +54,8 @@ namespace <%= namespace %>.SDK.iOS {
             if (tokenDictionary.ContainsKey(TokenField.PaymentData.ToString("G"))) {
                 var paymentDictionary = tokenDictionary[TokenField.PaymentData.ToString("G")];
                 PaymentData = Json.Serialize(paymentDictionary);
+            } else {
+                PaymentData = "";
             }
         }
     }
--- a/scripts/generator/graphql_generator/csharp/SDK/iOS/PaymentDeserializer.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/iOS/PaymentDeserializer.cs.erb
@@ -1,4 +1,4 @@
-#if UNITY_iPHONE
+#if UNITY_IOS
 namespace <%= namespace %>.SDK.iOS {
     using System.Collections;
     using System.Collections.Generic;
@@ -21,6 +21,7 @@ namespace <%= namespace %>.SDK.iOS {
 
         readonly public MailingAddressInput BillingAddress;
         readonly public MailingAddressInput ShippingAddress;
+        readonly public string Email;
         readonly public string ShippingIdentifier;
         readonly public string TransactionIdentifier;
         readonly public string PaymentData;
@@ -35,6 +36,9 @@ namespace <%= namespace %>.SDK.iOS {
             // Shipping Address
             var shippingContactDictionary = (Dictionary<string, object>)dictionary[PaymentField.ShippingContact.ToString("G")];
             ShippingAddress = new MailingAddressInput(shippingContactDictionary);
+
+            // Email
+            Email = (string)shippingContactDictionary["email"];
 
             // Shipping Identifier
             ShippingIdentifier = (string)dictionary[PaymentField.ShippingIdentifier.ToString("G")];
--- a/scripts/generator/graphql_generator/csharp/SDK/iOS/PaymentDeserializer.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/iOS/PaymentDeserializer.cs.erb
@@ -41,15 +41,20 @@ namespace <%= namespace %>.SDK.iOS {
             Email = (string)shippingContactDictionary["email"];
 
             // Shipping Identifier
-            ShippingIdentifier = (string)dictionary[PaymentField.ShippingIdentifier.ToString("G")];
+            if (dictionary.ContainsKey(PaymentField.ShippingIdentifier.ToString("G"))) {
+                ShippingIdentifier = (string)dictionary[PaymentField.ShippingIdentifier.ToString("G")];
+            }
 
             // Transaction Identifier
             var tokenDictionary = (Dictionary<string, object>)dictionary[PaymentField.TokenData.ToString("G")];
             TransactionIdentifier = (string)tokenDictionary[TokenField.TransactionIdentifier.ToString("G")];
 
             // Payment Data
-            var paymentDictionary = tokenDictionary[TokenField.PaymentData.ToString("G")];
-            PaymentData = Json.Serialize(paymentDictionary);
+            // Payment Data can be null when running on a simulator
+            if (tokenDictionary.ContainsKey(TokenField.PaymentData.ToString("G"))) {
+                var paymentDictionary = tokenDictionary[TokenField.PaymentData.ToString("G")];
+                PaymentData = Json.Serialize(paymentDictionary);
+            }
         }
     }
 }
--- a/scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
@@ -71,7 +71,7 @@ namespace <%= namespace %>.SDK.iOS {
             if (payment.PaymentData == null) {
                 message.Respond(new ApplePayEventResponse(ApplePayAuthorizationStatus.Failure).ToJsonString());
             } else {
-                var tokenizedPaymentInput = new TokenizedPaymentInput(amount, payment.BillingAddress, payment.TransactionIdentifier, payment.PaymentData, "apple_pay");
+                var tokenizedPaymentInput = new TokenizedPaymentInput(amount: amount, billingAddress: payment.BillingAddress, idempotencyKey: payment.TransactionIdentifier, paymentData: payment.PaymentData, type: "apple_pay");
 
                 Action performCheckout = () => {
                     CheckoutWithTokenizedPayment(tokenizedPaymentInput, checkout, (ApplePayEventResponse errorResponse) => {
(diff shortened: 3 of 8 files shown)


scripts/generator/graphql_generator/csharp/SDK/iOS/PaymentDeserializer.cs.erb
```diff
@@ -36,16 +37,26 @@ namespace <%= namespace %>.SDK.iOS {
...
 
             // Transaction Identifier
             var tokenDictionary = (Dictionary<string, object>)dictionary[PaymentField.TokenData.ToString("G")];
             TransactionIdentifier = (string)tokenDictionary[TokenField.TransactionIdentifier.ToString("G")];
 
             // Payment Data
-            var paymentDictionary = tokenDictionary[TokenField.PaymentData.ToString("G")];
-            PaymentData = Json.Serialize(paymentDictionary);
+            // Payment Data can be null when running on a simulator
+            if (tokenDictionary.ContainsKey(TokenField.PaymentData.ToString("G"))) {
```
sleroux: Would it make sense to use `#if IOS_SIMULATOR_SDK` here to blank out the `PaymentData`? It would allow us to differentiate between running on the simulator and an issue with the tokenDictionary. This also might not even be an issue 🤷‍♂️

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -60,7 +60,76 @@ namespace <%= namespace %>.SDK.iOS {
         }
 
         public void FetchApplePayCheckoutStatusForToken(string serializedMessage) {
```
sleroux: I'm not sure about the name of this method. It's named `Fetch..` but doesn't return anything. It looks like it finishes by sending a the response back to the native side. Is it possible to name it to reflect that?

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -80,16 +149,67 @@ namespace <%= namespace %>.SDK.iOS {
...
         }
 
+         private ApplePayAuthorizationStatus GetAuthorizationStatusFromCheckoutUserErrors(List<UserError> errors) {
+
+            ApplePayAuthorizationStatus statusToReturn = ApplePayAuthorizationStatus.Failure;
+
+            foreach (var error in errors) {
+                var fields = error.field();
+
+                // If there exists a UserErrorPath that is atleast 2 Fields long, and the second last field is `billingAddress`
```
sleroux: Is it safe to assume this in the future? I don't see the form for a checkout changing anytime soon. Just curious if this is going to be solid down the road.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -80,16 +149,67 @@ namespace <%= namespace %>.SDK.iOS {
...
+            var shippingAddressFieldsSet = new HashSet<string>(shippingAddressFields);
+
+            string[] shippingContactFields = {"firstName", "lastName", "phone", "email"};
+            var shippingContactFieldsSet = new HashSet<string>(shippingContactFields);
+
+            ApplePayAuthorizationStatus statusToReturn = ApplePayAuthorizationStatus.Failure;
+
+            foreach (var error in errors) {
+                var fields = error.field();
+                var lastField = fields[fields.Count - 1];
```
sleroux: I'm assuming that if `.field()` returns something it means it always has at least 1 entry but I would add an `if` check just to be safe and `break;` if we have no fields.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -80,16 +149,67 @@ namespace <%= namespace %>.SDK.iOS {
...
+
+                var isInvalidShippingAddressError =  shippingAddressFieldsSet.Contains(lastField);
+                var isInvalidShippingContactError =  shippingContactFieldsSet.Contains(lastField);
```
sleroux: Nice use of `HashSet` 👍

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -80,16 +149,67 @@ namespace <%= namespace %>.SDK.iOS {
...
+
+                if (isInvalidShippingAddressError) {
+                    statusToReturn = ApplePayAuthorizationStatus.InvalidShippingPostalAddress;
+                } else if (isInvalidShippingContactError) {
+                    statusToReturn = ApplePayAuthorizationStatus.InvalidShippingContact;
+                } else {
+                    statusToReturn = ApplePayAuthorizationStatus.Failure;
+                    break;
```
sleroux: Is this break needed?

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -80,16 +149,67 @@ namespace <%= namespace %>.SDK.iOS {
...
 
+            string[] shippingAddressFields = {"city", "country", "province", "zip"};
+            var shippingAddressFieldsSet = new HashSet<string>(shippingAddressFields);
+
             foreach (var error in errors) {
                 var fields = error.field();
                 var lastField = fields[fields.Count - 1];
-                var isInvalidShippingAddressError =
-                    lastField.Equals("city") || lastField.Equals("country") || lastField.Equals("province") || lastField.Equals("zip");
+                var isInvalidShippingAddressError = shippingAddressFieldsSet.Contains(lastField);
```
sleroux: 👍

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -60,7 +60,76 @@ namespace <%= namespace %>.SDK.iOS {
...
+                shippingFields = new ShippingFields(payment.ShippingAddress, payment.ShippingIdentifier);
+            }
+
+            CurrentCart.SetFinalCheckoutFields(payment.Email, shippingFields, (ShopifyError error) => {
+                if (error == null) {
+                    performCheckout();
+                } else {
+                    respondToSetFinalCheckoutFieldsError(error);
+                }
+            });
```
sleroux: The length of this method and the number of callbacks we're inlining makes me a bit uneasy but I'm not sure splitting it up will make sense since it's mostly code related to flow. Does it make sense to split this up into two helper methods for setting the final checkout fields and doing the checkout?

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -289,7 +289,7 @@ namespace <%= namespace %> {
             });
         }
 
-        private void CheckoutWithTokenizedPayment(TokenizedPaymentInput tokenizedPaymentInput, CompletionCallback callback) {
+        public void CheckoutWithTokenizedPayment(TokenizedPaymentInput tokenizedPaymentInput, CompletionCallback callback) {
```
sleroux: We should an audit at some point of what we make public on Cart because it might be confusing for developer to see a method like this and use it. If I see correctly this is only made public so it can be accessed by `iOSNativeCheckout` right?

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
dengjeffrey: I suppose this could break if graphQL decides to add another branching field to `billingAddress`. To be more solid I suppose we can check all the fields for a matching string

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
dengjeffrey: I talked with Mikko, these checks should happen higher up, when we construct a `ShopifyError` because this would be an issue with the SDK

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
dengjeffrey: I suppose reading this from the `Swift` end it might make more sense, to me this method is similar to an HTTP request. I'm not sure which keyword we could use to make this sound more clear, `Get`?.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
dengjeffrey: yes, if we encounter an unrecoverable user error we can stop looking

scripts/generator/graphql_generator/csharp/Cart.cs.erb
dengjeffrey: Yep

mikkoh: Agreed.

scripts/generator/graphql_generator/csharp/SDK/iOS/PaymentDeserializer.cs.erb
dengjeffrey: I removed the `IOS_SIMULATOR_SDK`, turns out to test payments you can set `PaymentData` to an empty string.

dengjeffrey: Yes good point, I'll add some checks

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
sleroux: Fetch sounds better than Get. Let's keep it the way it is.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -60,7 +62,34 @@ namespace <%= namespace %>.SDK.iOS {
...
-            //TODO
+            var checkout = CurrentCart.CurrentCheckout;
+            var message = NativeMessage.CreateFromJSON(serializedMessage);
+            var amount  = checkout.totalPrice();
+            var payment = new ApplePayPayment(message.Content);
+
+            if (payment.PaymentData == null) {
+                message.Respond(new ApplePayEventResponse(ApplePayAuthorizationStatus.Success).ToJsonString());
+            } else {
+                var tokenizedPaymentInput = new TokenizedPaymentInput(amount, payment.BillingAddress, payment.TransactionIdentifier, payment.PaymentData, "apple_pay");
```
mikkoh: Can we use named params here.

PR #161, opened 2017-07-07
[Feature] Adds checkout validation after navigating to end of a web checkout.

sleroux:
* Adds DidCheckoutCompleteSuccessfully method that checks the CurrentCart's completedAt date to see if it's not null.
* Updates DateTime customScalar in the generator to return a nullable type.

### DIFF
--- a/scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
@@ -16,7 +16,7 @@ namespace <%= namespace %>.SDK {
         }
 
         public enum ErrorType {
-            HTTP, GraphQL, UserError, NativePaymentProcessingError, WebCheckoutVerificationError
+            HTTP, GraphQL, UserError, NativePaymentProcessingError
         }
 
         public readonly ErrorType Type;
--- a/scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
@@ -16,7 +16,7 @@ namespace <%= namespace %>.SDK {
         }
 
         public enum ErrorType {
-            HTTP, GraphQL, UserError, NativePaymentProcessingError
+            HTTP, GraphQL, UserError, NativePaymentProcessingError, WebCheckoutVerificationError
         }
 
         public readonly ErrorType Type;
--- a/scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
@@ -3,9 +3,18 @@ namespace <%= namespace %>.SDK.iOS {
     using UnityEngine;
     using System.Runtime.InteropServices;
     using <%= namespace %>.SDK;
+    using <%= namespace %>.GraphQL;
 
     class iOSWebCheckout: IWebCheckout {
         private static iOSWebViewMessageReceiver _messageReceiver;
+
+        private ShopifyClient Client;
+        private Cart Cart;
+
+        public iOSWebCheckout(Cart cart, ShopifyClient client) {
+            Client = client;
+            Cart = cart;
+        }
 
         [DllImport ("__Internal")]
         private static extern void _CheckoutWithWebView(string unityDelegateObjectName, string checkoutURL);
@@ -15,11 +24,14 @@ namespace <%= namespace %>.SDK.iOS {
             _CheckoutWithWebView(GlobalGameObject.Name, checkoutURL);
         }
 
-        private static void SetupNativeMessageReceiver(CheckoutSuccessCallback success, CheckoutCancelCallback cancelled, CheckoutFailureCallback failure) {
+        private void SetupNativeMessageReceiver(CheckoutSuccessCallback success, CheckoutCancelCallback cancelled, CheckoutFailureCallback failure) {
             if (_messageReceiver == null) {
                 _messageReceiver = GlobalGameObject.AddComponent<iOSWebViewMessageReceiver>();
             }
         
+            _messageReceiver.Client = Client;
+            _messageReceiver.CurrentCheckout = Cart.CurrentCheckout;
+
             _messageReceiver.OnSuccess = success;
             _messageReceiver.OnCancelled = cancelled;
             _messageReceiver.OnFailure = failure;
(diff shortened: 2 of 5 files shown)


scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
```diff
@@ -36,12 +51,42 @@ namespace <%= namespace %>.SDK.iOS {
...
+                    }
+                });
             } else if (message.Content == "cancelled") {
                 OnCancelled();
                 message.Respond("");
             }
         }
+
+        private delegate void CheckoutCheckCallback(bool didComplete);
+        private void DidCheckoutCompleteSuccessfully(string checkoutId, CheckoutCheckCallback callback) {
```
dengjeffrey: Does this sound a bit like a delegate callback? Would `VerifyCheckoutCompleteSuccessfully` be a better name?

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
```diff
@@ -36,12 +51,42 @@ namespace <%= namespace %>.SDK.iOS {
...
+            query.node(
+                buildQuery: n => n
+                    .onCheckout(c => c.completedAt()),
+                id: CurrentCheckout.id()
+            );
+            
+            Client.Query(query, (response, error) => {
+                var checkout = (Checkout) response.node();
+                if (error != null) {
+                    OnFailure(error);
```
dengjeffrey: 🔥 callback is left hanging

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
```diff
@@ -36,12 +51,42 @@ namespace <%= namespace %>.SDK.iOS {
...
+                        OnFailure(error);
+                    }
+                });
             } else if (message.Content == "cancelled") {
                 OnCancelled();
                 message.Respond("");
             }
         }
+
+        private delegate void CheckoutCheckCallback(bool didComplete);
```
dengjeffrey: prefer that `didComplete` be renamed to `didSucceed`, `complete` is synonymous with having the function complete its check

scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
```diff
@@ -16,7 +16,7 @@ namespace <%= namespace %>.SDK {
         }
 
         public enum ErrorType {
-            HTTP, GraphQL, UserError, NativePaymentProcessingError
+            HTTP, GraphQL, UserError, NativePaymentProcessingError, CheckoutError
```
dengjeffrey: Is `CheckoutError` too broad? One might wonder if they need to handle a `CheckoutError` on all checkouts.  Currently it is only thrown when verification fails, perhaps `WebCheckoutVerificationError`

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
sleroux: Yup that sounds a lot better

sleroux: Actually maybe just `VerifyCheckoutComplete` since the success is determined by the bool it gets passed in.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
sleroux: I've added an error to the completion callback so the callback is always called.

dengjeffrey: I'm not sure we should be calling `OnFailure` here. As a function `DidCheckoutCompleteSuccessfully` should return if checkout completed successfully or not via the `callback`. If it didn't complete successfully then you might want to attach the error to the callback. 

Not calling `callback` here puts more burden on the implementor when they wonder why their callback was never called, on success or failure.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb
```diff
@@ -36,12 +51,46 @@ namespace <%= namespace %>.SDK.iOS {
...
                 // Invoke callbacks on cart to pass to Unity
-                OnSuccess();
+                DidCheckoutCompleteSuccessfully(CurrentCheckout.id(), (didSucceed, error) => {
+                    if (didSucceed) {
+                        OnSuccess();
+                        return;
+                    } 
+
+                    if (error != null) {
+                        OnFailure(error);
```
dengjeffrey: Not exactly sure if we should be sending the `HTTP` error to the implementor as an `OnFailure` since the transaction could of be completed but our verification failed to complete. Maybe this should be a `WebCheckoutVerificationError` as well. Thoughts on this @mikkoh?

sleroux: I like the idea of wrapping both these in the `WebCheckoutVerificationError` since technically if we run into an HTTP error there was an error encountered while trying to verify.

scripts/generator/graphql_generator/csharp/SDK/ShopifyError.cs.erb
```diff
@@ -16,7 +16,7 @@ namespace <%= namespace %>.SDK {
         }
 
         public enum ErrorType {
-            HTTP, GraphQL, UserError, NativePaymentProcessingError
+            HTTP, GraphQL, UserError, NativePaymentProcessingError, WebCheckoutVerificationError
```
mikkoh: Is there a reason why we need `WebCheckoutVerificationError`. Could we not use a `NativePaymentProcessingError`?

sleroux: @dengjeffrey Is NativePaymentProcessingError something we use for Apple Pay? I think the errors are different in that one is related to Apple Pay processing and the other is related to validating a checkout. I might be better from a game dev perspective to just use one of these though.

sleroux: I'd be down for getting rid of WebCheckoutValidationError. It seems too specific to web checkout. If we find we need something more specific later we can add something.

dengjeffrey: Yes this might be better, they both technically are validating that the checkout is valid. We would have to rename `NativePaymentProcessingError` though

PR #162, opened 2017-07-07
[Fix] Adds finished check before calling the cancelled callback

sleroux:
* Checks to see if we've finished before sending the 'cancelled' message to Unity/C#.


sleroux:
Hold on on this PR. Needs some more detail.

sleroux:
No longer needed since we moved away from the custom view controller and using SFSafariViewController

PR #168, opened 2017-07-10
[Feature] Use SFSafariViewController for native web checkouts on iOS

sleroux:
* Replaces our own custom web view controller with the SFSafariViewController.
* Updates the message interface between Unity/iOS with new dismissed message

https://github.com/Shopify/unity-buy-sdk/pull/161 will need to land first for this all to work properly.


sleroux:
Hey @dbart01, just for some context around all of these PRs:

We've been getting some initial feedback on the web checkout stuff and one of the downsides to using our own web experience is the lack of autofill. Since `SFSafariViewController` gives us that autofill we're thinking of moving in that direction instead. On the plus side this makes the web checkout implementation much simpler but removes a lot of the work already done.

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -68,53 +53,17 @@ protocol WebCheckoutDelegate: class {
...
-    }
-    
-    func willDismiss(completionHandler: UnityMessage.MessageCompletion?) {
-        hideOverlay()
-        let message = UnityMessage(content: "cancelled", object: unityDelegateObjectName, method: "OnNativeMessage")
-        MessageCenter.send(message, completionHandler: completionHandler)
+extension WebCheckoutSession: SFSafariViewControllerDelegate {
+    public func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
+        let message = UnityMessage(content: "dismissed", object: unityDelegateObjectName, method: "OnNativeMessage")
+        MessageCenter.send(message, completionHandler: nil)
```
dengjeffrey: Nit: `completionHandler` not needed when `nil`

PR #171, opened 2017-07-11
[Refactor] Split up C# iOS Plugin code for testability

sleroux:
It would be nice if there was a way to unit test the C# code behind our iOS plugin for Apple Pay and Web Checkout. Unfortunately as it stands a lot of the code is excluded from our mono test suite because it is either wrapped in UNITY_IOS or !SHOPIFY_MONO_UNIT_TEST flags. This PR tries to split this up to make us able to mono test this code by:

* Including a `-define UNITY_IOS` flag to our `mcs` command to include iOS code for testing
* For iOSWebCheckout, #ifdef out Unity-specific code to allow us to run it on Mono test
* For iOSNativeCheckout, move it to an abstract class and create a subclass named `iOSUnityNativeCheckout` that contains all of the UnityEngine code. This gives us the benefit of isolating the Unity code for mono'testing but also gives us a way hook for non-Unity C# engines.
* I also moved the old partial `iOSNativeCheckout` stuff into `iOSUnityNativeCheckout`. Not sure fi we want to keep this as a partial?

I'm not sure if this is the best approach. Let me know!


sleroux:
Tagging @JakeCataford in case he has some experience/opinions about this stuff!

scripts/generator/graphql_generator/csharp/SDK/NativeMessage.cs.erb
```diff
@@ -25,8 +23,9 @@ namespace <%= namespace %>.SDK {
         }
 
         public void Respond(string message) {
+            #if UNITY_IOS
```
dengjeffrey: Should this be `#if !SHOPIFY_MONO_UNIT_TEST`? 
Wouldn't this 💥 , now that you've added `UNITY_IOS` to the defined flags in mcs

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSUnityNativeCheckout.cs.erb
```diff
@@ -0,0 +1,83 @@
...
+        /// </summary>
+        /// <remarks>
+        ///  Displays a payment interface to the user based on the contents of the Cart
+        /// </remarks>
+        /// <param name="key">Merchant ID for Apple Pay from the Apple Developer Portal</param>
+        /// <param name="success">Delegate method that will be notified upon a successful payment</param>
+        /// <param name="failure">Delegate method that will be notified upon a failure during the checkout process</param>
+        /// <param name="cancelled">Delegate method that will be notified upon a cancellation during the checkout process</param>
+        public void Checkout(string key, CheckoutSuccessCallback success, CheckoutCancelCallback cancelled, CheckoutFailureCallback failure) {
+            StoreCallbacks(success, cancelled, failure);
```
dengjeffrey: Would this be more fitting:`SetCallbacks`

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.cs.erb
```diff
@@ -67,32 +46,105 @@ namespace <%= namespace %>.SDK.iOS {
...
-            OnFailure = failure;
+        public void UpdateSummaryItemsForShippingIdentifier(string serializedMessage) {
+            var message = NativeMessage.CreateFromJSON(serializedMessage);
+
+            CurrentCart.SetShippingLine(message.Content, (ShopifyError error) => {
+                if (error == null) {
+                    var summaryItems = GetSummaryItems();
+                    message.Respond(new ApplePayEventResponse(ApplePayAuthorizationStatus.Success, summaryItems).ToJsonString());
+                } else {
+                    message.Respond(new ApplePayEventResponse(ApplePayAuthorizationStatus.Failure).ToJsonString());
```
dengjeffrey: These methods aren't Mono testable because of `message.Respond`. These need to be abstract methods as well

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSUnityNativeCheckout.cs.erb
sleroux: I like that better. Updating.

sleroux:
Thanks for the feedback. I'm going to close this PR for now and reopen one that is smaller.

sleroux:
Cleaned up PR https://github.com/Shopify/unity-buy-sdk/pull/177.

PR #173, opened 2017-07-12
[Fix] Fixes presentation and status bar issues with safari view controller for web checkouts

sleroux:
To resolve both the status bar and push animation issues, I've wrapped the SFSafariViewController with a custom UINavigationController. The custom UINavigationController will also constrain the rotation allowed to whatever the game supports. If the game is landscape only, the controller will only be landscape!

### DIFF
--- a/Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
+++ b/Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
@@ -55,7 +55,11 @@ private var activeWebPaySession: WebCheckoutSession?
         
         let webViewController = SFSafariViewController(url: url)
         webViewController.delegate = self
-        UnityAppController.root.present(webViewController, animated: true, completion: nil)
+
+        let navController = WebCheckoutNavigationController(rootViewController: webViewController)
+        navController.isNavigationBarHidden = true
+        UnityAppController.root.present(navController, animated: true, completion: nil)
+
         return true
     }
 }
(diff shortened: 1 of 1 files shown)


Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -74,3 +78,12 @@ private extension UnityAppController {
         return unityController.rootViewController
     }
 }
+
+// Small helper class for presenting the SFSafariViewController
+private class WebCheckoutNavigationController: UINavigationController {
+        
+    // Only allow the same rotations the underlying game allows.
+    override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
+        return UnityAppController.root.supportedInterfaceOrientations
```
dengjeffrey: We can also use `GetAppController().rootViewController` from `UnityAppController.h` since we include that class in the bridging header

sleroux: Oh I wasn't aware of this! I'll update.

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -74,3 +78,12 @@ private extension UnityAppController {
...
+    }
+}
```
sleroux: We should be using this `UINavigationController` for any screens we want to modally present on top of the Unity controller so I think renaming this to something like `UnityModalNavigationController` makes more sense.

Wouldn't hurt to also override those two options!

sleroux:
@dengjeffrey Updated with comment fixes

Assets/Plugins/iOS/Shopify/WebCheckoutSession.swift
```diff
@@ -74,3 +78,18 @@ private extension UnityAppController {
         return unityController.rootViewController
     }
 }
+
+// Small helper class for presenting the SFSafariViewController
```
dengjeffrey: nit: need a new description for this class

PR #176, opened 2017-07-12
[Feature] Partial apple pay tests

dengjeffrey:
+ Added an `ApplePayEventReceiverBridge` to `Tester`
    + Depending on which apple pay tests we are doing the bridge's receiver will be set to the correct test
+ Added one test for Apple Pay
    + Will add more in later PRs to keep this one short


sleroux:
When I 🎩 using `./script/test_iOS.sh` I'm getting:
```
Assets/UnityTestTools/NativeTesting/iOS/ApplePayEventTesterReceiver.cs(8,48): error CS0246: The type or namespace name `IApplePayEventReceiver' could not be found. Are you missing an assembly reference?
Assets/UnityTestTools/NativeTesting/iOS/ApplePayEventTesterReceiver.cs(32,22): error CS0246: The type or namespace name `SummaryItem' could not be found. Are you missing an assembly reference?
Assets/UnityTestTools/NativeTesting/iOS/ApplePayEventTesterReceiver.cs(38,22): error CS0246: The type or namespace name `ShippingMethod' could not be found. Are you missing an assembly reference?
```

Are you seeing this too? I think that's what causing the CI failure.

Assets/UnityTestTools/NativeTesting/iOS/ApplePayEventTesterReceiver.cs
```diff
@@ -0,0 +1,44 @@
+
+namespace Shopify.Tests.iOS {
+    using System.Collections;
+    using System.Collections.Generic;
+    using UnityEngine;
```
sleroux: I don't think you need to import UnityEngine here

sleroux:
Ah it looks like its because they are hidden behind the UNITY_IOS flag

PR #177, opened 2017-07-13
[Refactor] Refactors classes/methods that rely on UNITY_IOS to support monotesting.

sleroux:
* Defines UNITY_IOS preproccessor in test script mcs command to include all iOS dependent classes.
* Moves iOSWebViewMessageReceiver into it's own class for easier #ifdef exclusion for SHOPIFY_MONO_UNIT_TEST/UNITY_IOS.
* Removed not needed UnityEngine import for NativeMessage.cs.erb and wrapped extern call with !SHOPIFY_MONO_UNIT_TEST.
* Split ApplePayEventReceiverBridge.cs.erb into 2 partial classes to isolate MonoBehaviour subclassing.

### DIFF
--- a/scripts/generator/graphql_generator/csharp.rb
+++ b/scripts/generator/graphql_generator/csharp.rb
@@ -127,6 +127,7 @@ module GraphQLGenerator
         SDK/iOS/ApplePayEventResponse
         SDK/iOS/ShippingMethod
         SDK/iOS/iOSWebCheckout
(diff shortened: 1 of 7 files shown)


sleroux:
For now, there are some unused warnings being thrown for the ApplePay but I plan on addressing those as part of the public API audit since a lot of that code will probably need to move anyways.

PR #178, opened 2017-07-13
[Fix] Adds cancelled state handling for web checkouts and tests

sleroux:
Before, we were only ever invoking success when a checkout has a completedAt date or failure when it doesn't. In the case of cancelling the web checkout, the failure callback was being invoked instead of cancelled. It doesn't make much sense to 'fail' when the user just 'cancelled' so this patch will only call the failure callback if we ran into some kind of failure communicating with our API and calls cancelled if we don't see the completedAt date. This means that in cases of spoofing the web view that it will come back as cancelled but that seems like an edge case.


sleroux:
Requires https://github.com/Shopify/unity-buy-sdk/pull/177 to get the tests actually running so I would hold off merging this until that lands.

PR #181, opened 2017-07-14
[Feature] iOS CI Tests

dengjeffrey:
+ Adds scripts to activate and deactivate license
+ Changed the iOS Test Scene to be Unity 5.4.2 compatible 
+ Adds iOS test script to Circle's commands

#### Caveat
+ We need to use Unity 5.4.2 locally when running `test_ios.sh` because the test scene will become corrupted otherwise, since it is not backwards compatible between the latest version of Unity and 5.4.2

#### Decisions
I generate `ApplePayEventReceiverBridge` via scripts because every time the directory was cleaned the component in iOS Test Scene with `ApplePayEventReceiverBridge` becomes corrupted .


circle.yml
```diff
@@ -3,28 +3,32 @@ machine:
...
-    IOS_PKG_URL: http://netstorage.unity3d.com/unity/fdbb5133b820/MacEditorTargetInstaller/UnitySetup-iOS-Support-for-Editor-5.3.4f1.pkg
+    IOS_PKG_URL: http://netstorage.unity3d.com/unity/b7e030c65c9b/MacEditorTargetInstaller/UnitySetup-iOS-Support-for-Editor-5.4.2f2.pkg
     UNITY_CIRCLE_XML_DIR: $CIRCLE_TEST_REPORTS/Unity
     UNITY_CIRCLE_XML_OUT_PATH: $UNITY_CIRCLE_XML_DIR/junit.xml
     # These are actually not needed but since Circle expects all OSX projects to be
     # XCode / iOS projects we need to drop these in
-    XCODE_SCHEME: symbolicator
-    XCODE_WORKSPACE: symbolicator
-  # Also not needed but Circle needs this
+    #XCODE_SCHEME: symbolicator
```
richardmonette: Do we need to keep these comments?

dengjeffrey: No we absolutely don't, nice catch!

richardmonette:
yes that is awesome

PR #182, opened 2017-07-14
[Cleanup] Refactors Cart.Poll into CheckoutPoller/PaymentPoller

sleroux:
First part of moving out the Apple Pay setters of Cart.cs. This patch removes the partial we had on Cart.Poll.cs and moves the polling logic into their own classes CheckoutPoller and PaymentPoller. It doesn't make that much sense in the context of this patch but is important for moving the Apple Pay methods out since we won't be passing in a cart instance into iOSNativeCheckout.


scripts/generator/graphql_generator/csharp/SDK/PaymentPoller.cs.erb
```diff
@@ -0,0 +1,38 @@
...
+        public void PollPaymentReady(string paymentId, PaymentPollFinishedHandler callback) {
+            PollUpdatedHandler isReady = (updatedQueryRoot) => {
+                var payment = (Payment)updatedQueryRoot.node();
+                return payment.ready();
+            };
+
+            PollPaymentNode(paymentId, isReady, callback);
+        }
+    }
+}
```
dengjeffrey: nit: New line

PR #183, opened 2017-07-14
[Cleanup] Refactors Checkout mutations on Cart into CheckoutMutator

sleroux:
PR is based off of https://github.com/Shopify/unity-buy-sdk/pull/182

**Problem**

One of the challenges we had with implementing the logic for Apple Pay was that `iOSNativeCheckout` required a reference to the Cart to make updates to the current checkout. In turn this required us to expose methods on Cart that would have no use to the SDK user and are only used for the Apple Pay implementation (iOSNativeCheckout). 

**Solution**

This refactoring addresses the core of the issue by extracting all of the mutation logic on a Checkout out of `Cart` and into a new `CheckoutMutator` class. By doing so, `iOSNativeCheckout` no longer has a dependency on the Cart and can make mutations to it on it's own. This also gives us the ability to move out the other mutations we were doing on Cart (such as Create/Update/Save) into the mutator to keep the Cart class lean and single purposed. One wrinkle though is that we still want the checkout to live in the Cart object. To allow the state to be stored in Cart but have mutations run in CheckoutMutator, a `CheckoutMutationCallback` is passed in (UpdateState) so Cart can update the state whenever we make changes.

**Issues**
One remaining issue is that `ApplePayFlowTester` uses the old `SetShippingAddress` method on Cart and now with that removed there isn't a way to perform this operation. The test fails at the moment.


sleroux:
Looking for feedback on the approach before I write tests around the mutation methods.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -191,218 +152,60 @@ namespace <%= namespace %> {
...
+
+            List<CheckoutLineItemInput> toAdd = CartLineItems.ConvertToCheckoutLineItemInput(LineItems.All());
+
+            CheckoutMutator.UpdateItems(
+                checkout: CurrentCheckout,
+                lineItemsToAdd: toAdd,
+                lineItemsToRemove: toRemove,
+                callback: (e) => {
+                    DeletedLineItems.Clear();
+                    callback(e);
```
dengjeffrey: If `UpdateItems` result in a failure, we still clear the deleted line items, do we want to do that?

scripts/generator/graphql_generator/csharp/SDK/CheckoutMutator.cs.erb
```diff
@@ -0,0 +1,268 @@
...
+
+        private void HandleUserError(CompletionCallback callback) {
+            ShopifyError error = new ShopifyError(
+                ShopifyError.ErrorType.UserError,
+                "There were issues with some of the fields sent. See `cart.UserErrors`"
+            );
+            callback(error);
+        }
+    }
+}
```
dengjeffrey: Nit: new line

scripts/generator/graphql_generator/csharp/SDK/CheckoutMutator.cs.erb
```diff
@@ -0,0 +1,268 @@
...
+            Client.Mutation(query, (Mutation response, ShopifyError error) => {
+                if (error != null) {
+                    callback(error);
+                    return;
+                }
+
+                HandleCheckoutMutation(
+                    response.checkoutEmailUpdate().checkout(),
+                    response.checkoutEmailUpdate().userErrors(),
+                    CheckoutPoller.PollCheckoutAvailableShippingRatesReady,
```
dengjeffrey: I just realized this line should be `CheckoutPoller.PollCheckoutReady` since Email doesn't affect available shipping rates

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -63,14 +50,13 @@ namespace <%= namespace %> {
...
-        private const float POLL_DELAY_SECONDS = 0.5f;
         private CartLineItems _LineItems;
-        private List<UserError> _UserErrors = null;
+        private List<UserError> _UserErrors = new List<UserError>();
         private List<String> DeletedLineItems = new List<string>();
-        private ShopifyClient Client;
 
         private IWebCheckout WebCheckout;
         private INativeCheckout NativeCheckout;
+        private CheckoutMutator CheckoutMutator;
```
dengjeffrey: Will this naming clash, if `CheckoutMutator` has class/static methods?

scripts/generator/graphql_generator/csharp/SDK/CheckoutMutator.cs.erb
```diff
@@ -0,0 +1,268 @@
...
+                callback(null);
+            } else {
+                CheckoutPoller.PollCheckoutAndUpdate(newCheckout.id(), pollFinishedHandler, callback);
+            }
+        }
+
+        private void HandleUserError(CompletionCallback callback) {
+            ShopifyError error = new ShopifyError(
+                ShopifyError.ErrorType.UserError,
+                "There were issues with some of the fields sent. See `cart.UserErrors`"
```
dengjeffrey: We're no longer in `cart` so this description needs to change.  I wonder if this would be better if it lived in `Cart`, maybe a delegate methods: `ErrorForMutateCheckoutUserErrors(UserErrors) -> ShopifyError` for error  and `DidMutateCheckout(newCheckout)` for success

scripts/generator/graphql_generator/csharp/SDK/CheckoutPoller.cs.erb
```diff
@@ -0,0 +1,77 @@
...
+        public void PollCheckoutAndUpdate(string checkoutId, CheckoutPoll poll, CompletionCallback callback) {
+            poll(checkoutId, (Checkout checkout, ShopifyError error) => {
+                if (error == null && checkout != null) {
+                    DidMutateCheckout(checkout, new List<UserError>());
+                }
+                callback(error);
+            });
+        }
+    }
+}
```
dengjeffrey: nit: new line

scripts/generator/graphql_generator/csharp/SDK/PaymentPoller.cs.erb
```diff
@@ -0,0 +1,38 @@
...
+        public void PollPaymentReady(string paymentId, PaymentPollFinishedHandler callback) {
+            PollUpdatedHandler isReady = (updatedQueryRoot) => {
+                var payment = (Payment)updatedQueryRoot.node();
+                return payment.ready();
+            };
+
+            PollPaymentNode(paymentId, isReady, callback);
+        }
+    }
+}
```
dengjeffrey: nit: new line

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -134,18 +136,17 @@ namespace <%= namespace %>.SDK.iOS {
             });
         }
 
-        private void CheckoutWithTokenizedPayment(TokenizedPaymentInput tokenizedPaymentInput, Checkout checkout, ApplePayEventHandlerCompletion callback) {
+        private void CheckoutWithTokenizedPayment(Checkout checkout, TokenizedPaymentInput tokenizedPaymentInput, ApplePayEventHandlerCompletion callback) {
```
dengjeffrey: I noticed that `checkout` here wasn't used, we can probably remove

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -191,218 +152,60 @@ namespace <%= namespace %> {
...
-                        PollCheckoutAndUpdate(PollCheckoutReady, callback);
-                    } else {
-                        HandleUserError(callback);
-                    }
-                }
-            });
-        }
-
-        private void CheckoutSave(CompletionCallback callback) {
+        public void CheckoutSave(CompletionCallback callback) {
```
sleroux: It might make more sense - I was just wary about changing the public facing API name especially if people are using it.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
sleroux: Hmm probably not! Good catch.

sleroux:
Thanks for all the feedback on this PR. I've opened a new one that accomplishes the same goal as the PR but is significantly de-risked here https://github.com/Shopify/unity-buy-sdk/pull/195. I think there is still value in building out abstractions like `CheckoutMutator` but the current implementation is a bit too leaky with the delegates it uses.

Closing

PR #184, opened 2017-07-14
[Feature] Complete Shipping Apple Pay Tests

dengjeffrey:
+ Adds tests that test apple pay checkout for items with Shipping required

#### Not included
+ tests for items that don't require shipping

### DIFF
--- a/Assets/Plugins/iOS/Shopify/BuyTests/Models/Models.swift
+++ b/Assets/Plugins/iOS/Shopify/BuyTests/Models/Models.swift
@@ -119,6 +119,10 @@ struct Models {
         return MockPaymentToken(paymentMethod: Models.createPaymentMethod() as! MockPaymentMethod)
     }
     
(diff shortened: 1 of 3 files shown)


Assets/Plugins/iOS/Shopify/BuyTests/Models/Models.swift
```diff
@@ -41,7 +41,7 @@ struct Models {
     static let city         = "Toronto"
     static let country      = "Canada"
     static let province     = "ON"
-    static let zip          = "A1B 2C3"
+    static let zip          = "M5V 2J4"
```
sleroux: Nit: Technically a postal code not a zip. Does it matter?

dengjeffrey: I just named it after the GraphQL fields for `MailingAddressInput`

PR #186, opened 2017-07-14
[Fix] Added failure callback

dengjeffrey:


sleroux:
Sorry I was a bit too hasty in approving this. How come there is a compilation failure in CI?

dengjeffrey:
When you see missing reference for the namespace `iOS`it means it didn't get a license.

PR #187, opened 2017-07-14
[Fix] Cleanup CI activation

dengjeffrey:


dengjeffrey:
Awesome, will look at that!

richardmonette:
Probably overkill, but if we changed from if else, to an if with early return, then you could put the check for `-n $UNITY_USERNAME && -n $UNITY_PASSWORD && -n $UNITY_SERIAL` in a separate script, and then `source` it into both files.

dengjeffrey:
Let's look at that at a later time https://github.com/Shopify/unity-buy-sdk/issues/190

PR #193, opened 2017-07-17
[Feature] Adds Reset method to Cart for clearing out state

sleroux:
Adds a `Reset` method on `Cart` that allows the cart to have it's state be reset and be reused.

### DIFF
--- a/Assets/Editor/TestCart.cs
+++ b/Assets/Editor/TestCart.cs
@@ -27,6 +27,41 @@ namespace Shopify.Tests
             Assert.IsNotNull(ShopifyBuy.Client().Cart(cartWithId), "created a cart with id");
             Assert.AreNotEqual(ShopifyBuy.Client().Cart(), ShopifyBuy.Client().Cart(cartWithId));
         }
+
+        [Test]
+        public void TestReset() {
+            ShopifyBuy.Init(new MockLoader());
+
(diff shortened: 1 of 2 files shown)


Assets/Editor/TestCart.cs
```diff
@@ -27,6 +27,41 @@ public class TestCart {
...
+
+            Assert.IsNotNull(responseURL, "Was able to generate a web checkout url");
+            Assert.IsNull(error, "No error from getting web checkout");
+            Assert.IsNotNull(cart.CurrentCheckout, "Cart has a checkout");
+
+            cart.Reset();
+
+            // Verify that we cleared out the line items and checkout.
+            Assert.AreEqual(0, cart.LineItems.All().Count, "has 0 items in cart");
+            Assert.IsNull(cart.CurrentCheckout, "cart has a null Checkout");
```
dengjeffrey: This might not be necessary, but should we also test that we can construct a different checkout again? Just to fully ensure that the Cart purged all data.

sleroux: Wouldn't hurt!

sleroux: Actually checking for null should be good enough to verify I think. The only way we'd get a new checkout is if we create one from the server anyways.

PR #195, opened 2017-07-18
[Refactor] Trim Cart's public API to only be methods we want to expose

sleroux:
This patch moves unwanted public methods, private state/methods, and poller methods into CartState. Cart now wraps around an instance of CartState. Previous clients of Cart like iOSNativeCheckout now accept a CartState instance instead of Cart. This is a lower risk way of cleanng up our public API compared to https://github.com/Shopify/unity-buy-sdk/pull/183 since it just moves code around and make very little changes. In the future it would make sense to break apart CartState into more specific interfaces/classes if needed.


scripts/generator/graphql_generator/csharp/CartState.cs.erb
```diff
@@ -0,0 +1,389 @@
...
+
+            Client.Mutation(query, (Mutation response, ShopifyError error) => {
+                if (error != null) {
+                    callback(error);
+                } else {
+                    if (UpdateState(response.checkoutEmailUpdate().checkout(), response.checkoutEmailUpdate().userErrors())) {
+                        if (CurrentCheckout.ready()) {
+                            callback(null);
+                        } else {
+                            PollCheckoutAndUpdate(PollCheckoutAvailableShippingRatesReady, callback);
```
dengjeffrey: Should be `PollCheckoutAndUpdate (PollCheckoutReady`

PR #197, opened 2017-07-19
[Feature] Adds iOS Post Processor for Build & Run from Unity

sleroux:
* Moved non-test processor functionality into `iOSPostProcessor.cs.erb` and moved `ExtendedPBXProject.cs.erb` into Shopify/Editor/BuildPipeline for developer use.
* Renamed `ShopifyIoSPostProcessor` to `iOSTestPostProcessor` and refactored calls to common build settings setup to `iOSPostProcessor`


Assets/Editor/BuildPipeline/iOSTestPostProcessor.cs
```diff
@@ -0,0 +1,45 @@
+#if UNITY_EDITOR
+namespace Shopify.Unity.Editor.BuildPipeline {
+    using UnityEngine;
+    using UnityEditor;
+    using UnityEditor.iOS.Xcode;
+    using System.IO;
+    using System;
+
+    public class iOSTestPostProcessor {
```
dengjeffrey: Can we subclass `iOSPostProcessor` here?

sleroux: Since all these methods are static we can't really use base class calls so we don't gain much unless we move everything to non static and instantiate an instance in the build pipeline. Thoughts?

dengjeffrey: True, this is fine then

PR #199, opened 2017-07-19
[Fix] Use store country code for payment session

dengjeffrey:
###### Temporarily merging in to #195 to reduce files changed

+ Adds queries for PaymentSettings of a Store
+ Uses the country code from payment settings

#### Decisions made:
Since we need to fetch the payment settings of a store before we can perform a checkout, we fetch it only when it is needed instead of when the `Cart` is created. 

##### Cons
+ By doing this we have to pass in `PaymentSettings` through the methods in the INativePayment interface

##### Pros

+ Users who are already using the SDK don't have to make any changes
+ Users who don't need to support Apple Pay don't need to make the PaymentSettings query

### DIFF
--- a/scripts/generator/graphql_generator/csharp/Cart.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/Cart.cs.erb
@@ -71,7 +71,7 @@ namespace <%= namespace %> {
         private INativeCheckout NativeCheckout;
         private PaymentSettings StorePaymentSettings;
 
-        private Client CurrentClient;
+        private ShopifyClient CurrentClient;
 
         #if (SHOPIFY_TEST)
         public CartState State;
@@ -95,6 +95,7 @@ namespace <%= namespace %> {
         /// \endcode
         public Cart(ShopifyClient client) {
             State = new CartState(client);
+            CurrentClient = client;
 
             #if UNITY_IOS
             WebCheckout = new iOSWebCheckout(this, client);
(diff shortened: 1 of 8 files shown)


Assets/Plugins/iOS/Shopify/BuyTests/PaymentSessionTests.swift
```diff
@@ -68,7 +68,7 @@ class PaymentSessionTests: XCTestCase {
     //  MARK: - Supported Networks -
     //
     func testSupportedNetworks() {
-        XCTAssertEqual(PaymentSession.supportedNetworks, [.amex, .masterCard, .visa, .discover])
```
richardmonette: Is changing the supported networks actually related to this change? (Also, won't this be a dynamic list soon enough anyways?)

dengjeffrey: No I suppose it could of went into a different PR, but it's kinda small, just shoved it in here.
Yes it will be dynamic soon.

richardmonette: I'd propose we take it out for now, the PRs are quicker to review when its just one concise change per PR.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -69,6 +69,9 @@ namespace <%= namespace %> {
 
         private IWebCheckout WebCheckout;
         private INativeCheckout NativeCheckout;
+        private PaymentSettings StorePaymentSettings;
+
+        private ShopifyClient CurrentClient;
```
sleroux: Nit: No need to call this `Current` since we're note re-assigning it or anything in the future.

PR #200, opened 2017-07-19
[Fix] Moves reset logic into CartState

sleroux:
Git didn't pick up the changes between CartState and the Reset patch so this resolves the compilation issue and moves the Reset method into CartState.


dengjeffrey:
Just ignore the buildkite CI for now

PR #207, opened 2017-07-20
[Fix] Added handling for single shipping rates

dengjeffrey:
+ Added check to see if there is one shipping method 
    + When there is one, update shipping lines immediately

##### Why?
+ When there is one shipping method, Apple does not call the delegate method for selecting a shipping line item.
+ Shopify Admin does not update the checkout query `Total`when there is only one shipping method, this seems like a bug.
    + You can checkout without selecting a shipping line item, and the total Shopify expects includes the single shipping cost when trying to complete a checkout


dengjeffrey:
It is not called when  it selects the only one by default.

It is called when it selects one from a list by default. For sanity I'm going to confirm this, this morning on a device and not on simulator

PR #210, opened 2017-07-20
Beta -> Master 0.2.0

sleroux:
Preparing 0.2.0 release


sleroux:
@dengjeffrey / @richardmonette How do we want to merge this?

1. Rebase commits ontop of master?
2. Create a merge commit on master?

Typically what I've done in the past is have all the commits land as is and use tags to cut releases. Alternatively I've had it where master was the 'development' branch and a release branch is cut from master to support additional hot fixes but we've been working on beta this entire time so I'm not sure what we want to do here.

dengjeffrey:
I don't mind this way, although I prefer that we don't squash, so that it shows nicely if/when we revert a commit. Making a tag that cooresponds with the version, would be nice

sleroux:
I think there is definitely way too many changes to do a squash and having the history will be really helpful. 👍

PR #212, opened 2017-07-20
[Fix] Squelch warnings

dengjeffrey:


scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -100,6 +100,9 @@ namespace <%= namespace %> {
             #if UNITY_IOS
             WebCheckout = new iOSWebCheckout(this, client);
             NativeCheckout = new iOSNativeCheckout(State);
+            #else
+            WebCheckout = null;
+            NativeCheckout = null;
```
JakeCataford: Not in this pr, but it would be cool to use the null object pattern here and have `NullWebCheckout` which responds to the same interface as the iOS/other web/native checkout objects and just no-ops or throws errors

scripts/generator/graphql_generator/csharp/CartState.cs.erb
```diff
@@ -300,7 +300,9 @@ namespace <%= namespace %> {
                 List<CheckoutLineItem> lineItems = (List<CheckoutLineItem>) checkout.lineItems();
 
                 LineItems.UpdateLineItemsFromCheckoutLineItems(lineItems);
+            #pragma warning disable 0168
             } catch(NoQueryException exception) {}
```
JakeCataford: This is because NoQueryException is defined in a generated file? can you explain why this warning occurs and why we need to disable it rather than provide a fix?

dengjeffrey: This warning occurs when we try to access a property we didn't query for, in this case it is `lineItems`. When we poll for checkout ready, we do not query for line items.  Since `UpdateLineItemFromCheckout` is used in `UpdateState` and `UpdateState` is used whenever we have an updated `Checkout` there are cases when `lineItems` has not been queried.

dengjeffrey:
@sleroux Feel free to merge this in, if you approve of this PR

PR #217, opened 2017-07-24
[Feature] Adds Android AAR project for SDK

sleroux:
* Adds an Android project with an AAR library target under a root new folder named `unity_buy_android`
* Adds `./scripts/build_android.sh` which compiles and exports the AAR library into the `Assets/Plugins/Android` folder for use as a Unity Plugin.
* Contains two temp test .java files for running unit tests.

I'm not sure what the best place for the Android source to live is. It shouldn't be included in the Unity project under Assets/ and it doesn't really make sense under scripts so I've added it as new root folder. Any thoughts?

I was originally going to use Buck (https://buckbuild.com/) but the documentation is slim and it's support for Android Studio is shaky at best.


sav007:
I would not commit *.idea folder under VC, put it to .ignore

unity_buy_android/unity_buy_plugin/build.gradle
```diff
@@ -0,0 +1,55 @@
+apply plugin: 'com.android.library'
+
+android {
+    compileSdkVersion 26
+    buildToolsVersion "26.0.0"
+
+    defaultConfig {
+        minSdkVersion 16
```
sav007: Min version is 16? You sure you want to support this old OS:
https://developer.android.com/about/dashboards/index.html

I believe you can easliy bump it up to 19. But it's up to you

sleroux: I think moving  to 19 makes sense. I was only targeting 16 because that is what Unity exports by default by more than 90% of users are on >= 19 it looks like. 

Also, it looks like the MBSDK's pay module is set to a minimum of 19 which would be nice to use for the Unity Android Pay implementation.

PR #222, opened 2017-07-26
Added gradle dependency for Unity classses, and classes.jar

dengjeffrey:


unity_buy_android/settings.gradle
```diff
@@ -1 +1 @@
-include ':unity_buy_plugin'
+include ':unity_buy_plugin', ':unity_classes'
```
sav007: is it new gradle sub module?

sav007: Oh I see, interesting
```
configurations.maybeCreate("default")
artifacts.add("default", file('classes.jar')) 
```

PR #223, opened 2017-07-27
[Quality] Adds checkstyle task to Android project

sleroux:
Copies over the checkstyle rules and task we use for other Shopify Android projects:

https://github.com/Shopify/mobile-shopify/wiki/Checkstyle

It's also worth adding this to Android Studio directly so you can run the rules as you develop:

https://github.com/Shopify/mobile-shopify/wiki/Checkstyle-in-Android-Studio

### DIFF
--- a/scripts/test_android.sh
+++ b/scripts/test_android.sh
@@ -1,0 +1,6 @@
+. $(dirname $0)/common.sh
+
+ANDROID_PLUGIN_ROOT="$PROJECT_ROOT/unity_buy_android"
+
+"$ANDROID_PLUGIN_ROOT"/gradlew unity_buy_plugin:check -p "$ANDROID_PLUGIN_ROOT" || die "Failed to check Android plugin"
+"$ANDROID_PLUGIN_ROOT"/gradlew unity_buy_plugin:test -p "$ANDROID_PLUGIN_ROOT" || die "Failed to test Android plugin"
--- a/scripts/test_android.sh
+++ b/scripts/test_android.sh
@@ -3,4 +3,3 @@
 ANDROID_PLUGIN_ROOT="$PROJECT_ROOT/unity_buy_android"
 
 "$ANDROID_PLUGIN_ROOT"/gradlew unity_buy_plugin:check -p "$ANDROID_PLUGIN_ROOT" || die "Failed to check Android plugin"
-"$ANDROID_PLUGIN_ROOT"/gradlew unity_buy_plugin:test -p "$ANDROID_PLUGIN_ROOT" || die "Failed to test Android plugin"
--- a/scripts/test_android_unit.sh
+++ b/scripts/test_android_unit.sh
@@ -1,6 +1,0 @@
-. $(dirname $0)/common.sh
-
-ANDROID_PLUGIN_ROOT="$PROJECT_ROOT/unity_buy_android"
-
-# Run Unity assembly task
-"$ANDROID_PLUGIN_ROOT"/gradlew unity_buy_plugin:testDebug -p "$ANDROID_PLUGIN_ROOT" || die "Failed to test Android plugin"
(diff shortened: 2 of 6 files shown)


unity_buy_android/checkstyle/config.xml
```diff
@@ -0,0 +1,223 @@
...
+                <property name="tokens" value="BAND, BOR, BSR, BXOR, DIV, EQUAL, GE, GT, LAND, LE, LITERAL_INSTANCEOF, LOR, LT, MINUS, MOD, NOT_EQUAL, PLUS, QUESTION, SL, SR, STAR "/>
+            </module>
+                    <module name="EmptyCatchBlock">
+                <property name="exceptionVariableName" value="expected"/>
+            </module>
+
+                <module name="LocalFinalVariableName"/>
+        <module name="StaticVariableName"/>
+        <module name="ConstantName"/>
+        -->
```
dengjeffrey: There's a lot of commented out lines here, will we need these lines? If not should they be removed?

scripts/test_android_unit.sh
```diff
@@ -2,5 +2,8 @@
 
 ANDROID_PLUGIN_ROOT="$PROJECT_ROOT/unity_buy_android"
 
+# Run checkstyle before running tests
+"$ANDROID_PLUGIN_ROOT"/gradlew unity_buy_plugin:checkstyle -p "$ANDROID_PLUGIN_ROOT" || die "Failed to run checkstyle"
+
```
sav007: You can do it inside of gradle, so that you don't need this line.

```
afterEvaluate {
      if (project.tasks.findByName('check')) {
        check.dependsOn('checkstyle')
      }
    }
```

sleroux: Should I also be running the `check` task instead of `testDebug` for running tests? Will that run unit and instrument tests?

unity_buy_android/checkstyle/config.xml
sleroux: I'm just going to leave it as is since this is just a copy of the Shopify Android is using.

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/UnityBuyPlugin.java
```diff
@@ -1,6 +1,8 @@
 package com.shopify.unity.buy;
 
 public class UnityBuyPlugin {
+    private UnityBuyPlugin() { }
```
dengjeffrey: Quick question, why is this needed?

sleroux: Checkstyle complains that utility classes shouldn't have a default or public constructors. Makes sense considering a class that only has static methods should never be instantiated. 

This class it just a dummy but figured I'd fix it so the test script doesn't fail.

sav007: Even more it must be final. Any Util class must be final, with private constructor.

scripts/test_android.sh
```diff
@@ -0,0 +1,6 @@
+. $(dirname $0)/common.sh
+
+ANDROID_PLUGIN_ROOT="$PROJECT_ROOT/unity_buy_android"
+
+"$ANDROID_PLUGIN_ROOT"/gradlew unity_buy_plugin:check -p "$ANDROID_PLUGIN_ROOT" || die "Failed to check Android plugin"
```
sav007: If I'm not wrong but check do testing as well

sleroux: Ah yup looks like it does! I'll remove the redundancy

sleroux:
@sav007 Removed the redundant gradle call. 👍?

PR #230, opened 2017-07-27
[Refactor] Payment Settings

mikkoh:
Should be noted this PR is merging onto `master`.

This PR removes `paymentSettings` from the client and just adds the query functionality to `Cart`.

### DIFF
--- a/scripts/generator/graphql_generator/csharp/Cart.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/Cart.cs.erb
@@ -235,11 +235,17 @@ namespace <%= namespace %> {
             if (StorePaymentSettings != null) {
                 callback(StorePaymentSettings, null);
             } else {
-                Client.paymentSettings((paymentSettings, error) => {
-                    if (error == null) {
-                        StorePaymentSettings = paymentSettings;
+                var query = new QueryRootQuery();
+                DefaultQueries.shop.PaymentSettings(query);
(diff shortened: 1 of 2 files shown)


sleroux:
How come we're merging it straight to master? Would it make sense to land this on beta then onto master?

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -235,11 +235,17 @@ namespace <%= namespace %> {
...
+                var query = new QueryRootQuery();
+                DefaultQueries.shop.PaymentSettings(query);
+
+                Client.Query(query, (response, error) => {
+                    if (error != null) {
+                        callback(null, error);
+                    } else {
+                        StorePaymentSettings = response.shop().paymentSettings();
+
+                        callback(response.shop().paymentSettings(), null);
```
dengjeffrey: Nit: could probably use: `callback(StorePaymentSettings, null)` to save a few dictionary lookups

mikkoh:
@sleroux `beta` currently contains stuff for Android. This really should have been in `master` from the start.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
mikkoh: 👍

PR #231, opened 2017-07-27
[Refactor] Native web checkout

dengjeffrey:
+ Moves non iOS specific functionality out of `iOSWebCheckout` and into `NativeWebCheckout`, so that we can re-use some of that functionality in Android


scripts/generator/graphql_generator/csharp/SDK/NativeWebCheckout.cs.erb
```diff
@@ -0,0 +1,31 @@
+namespace <%= namespace %>.SDK {
+    using System;
+
+    #if !SHOPIFY_MONO_UNIT_TEST
+    using UnityEngine;
+    #endif
+
+    class NativeWebCheckout: IWebCheckout {
```
sleroux: Would it make more sense to turn `IWebCheckout` into an abstract base class instead of an interface? At least that way we could flatten the class hierarchy a bit.

dengjeffrey:
@sleroux I refactored into abstract class

scripts/generator/graphql_generator/csharp/SDK/NativeWebViewMessageReceiver.cs.erb
```diff
@@ -17,6 +17,20 @@ namespace <%= namespace %>.SDK {
         public CheckoutCancelCallback OnCancelled;
         public CheckoutFailureCallback OnFailure;
 
+        public void Init(
+            ShopifyClient client,
+            Checkout currentCheckout,
+            CheckoutSuccessCallback onSuccess,
+            CheckoutCancelCallback onCancelled,
+            CheckoutFailureCallback onFailure) {
```
sleroux: Nit: Move `) {` to next line flush with public

Issue #213, opened 2017-07-20
Rename Apple Pay Total to Shop name


dengjeffrey:
Currently we use `TOTAL` as the label for the last summary item, resulting in `PAY TOTAL`. 
According to guidelines it should be the shop name, such that it reads `PAY SHOP`. So when it errors out, the user receives a message as such: `SHOP` was not able to complete payment. 

Although this sounds trivial, we need to rethink how we create a Cart or Shop Client. Ideally `Shop` information like the name should be available once you create a client with the access token and shop url.

PR #232, opened 2017-07-28
[Android Pay] Create AndroidPayCheckoutSession and load stub fragment from Unity

sleroux:
First part of implementing Android Pay:

* Creates an AndroidPayCheckoutSession Java object that takes in a Cart and other information from Unity and adds a UnityAndroidPayFragment to the root Activity
* Adds API to C# for starting the Android pay session
* Pulls in the buy3 Android Pay helper library


scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
```diff
@@ -0,0 +1,78 @@
...
+        ) {
+            // TODO: Store callbacks and extract items we need from the cart to pass to Android Pay.
+            OnSuccess = success;
+            OnCancelled = cancelled;
+            OnFailure = failure;
+
+            var checkout = CartState.CurrentCheckout;
+
+            // TODO: Replace with Shop name
+            var merchantName = "mables-cables";
```
sleroux: I can't leave this empty for now since `PayCart` on the android side will throw an exception. Temporary until we can get the `shop` query stuff in.

scripts/generator/graphql_generator/csharp/SDK/CheckoutUtils.cs.erb
```diff
@@ -0,0 +1,22 @@
+namespace <%= namespace %>.SDK {
+    using System.Collections.Generic;
+    using <%= namespace %>.MiniJSON;
+    
+    public class CheckoutUtils {
```
sleroux: I only needed this method from `iOSNativeCheckout` so I pulled it into a utility class but if I end up needed more I'll probably end up abstract base classing `INativeCheckout` down the road.

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -0,0 +1,102 @@
...
+import java.math.BigDecimal;
+
+public class AndroidPayCheckoutSession {
+
+    public boolean checkoutWithAndroidPay(
+            String unityDelegateObjectName,
+            String merchantName,
+            String summaryItemsString,
+            String currencyCode,
+            String countryCode,
```
sleroux: It's unclear how the `countryCode` maps to the buy3's `PayCart`. Any ideas?

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -0,0 +1,102 @@
...
+                    items.shippingPrice = amountFromJson(summaryItemObj);
+                } else if (label.equals("TAXES")) {
+                    items.taxPrice = amountFromJson(summaryItemObj);
+                } else if (label.equals("TOTAL")) {
+                    items.totalPrice = amountFromJson(summaryItemObj);
+                }
+            }
+
+            // Verify that we've set all of the required parameters that we need.
+            if (items.subtotal == null) {
```
sleroux: I added this validation to catch any developer errors when trying to pass missing information. Does this pattern make sense in Java?

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
sleroux: See https://github.com/Shopify/unity-buy-sdk/pull/233 to get the actual shop name

dengjeffrey: But these changes include `GetPaymentSettingsAndShopName`, so shouldn't these changes also reflect the use of shop name?

scripts/generator/graphql_generator/csharp/SDK/CheckoutUtils.cs.erb
dengjeffrey: I don't think we should be abstracting this out of `iOSNativeCheckout`, reason being is https://github.com/Shopify/unity-buy-sdk/issues/213 where we need to change `TOTAL` to shop name. 

Since both platforms are expecting different formats of `SummaryItem` I think we should have a subclass of `SummaryItem` for each platforms that returns a `SerializedSummaryItem` in the format that each platform uses. A list for iOS and a dictionary for Android

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -76,6 +80,7 @@ namespace <%= namespace %> {
         private IWebCheckout WebCheckout;
         private INativeCheckout NativeCheckout;
         private PaymentSettings StorePaymentSettings;
+        private String ShopName;
```
dengjeffrey: I think calling this `StoreName` would be more fitting. Reason being we want to stay away from `Shopify Shop`

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
dengjeffrey: I believe `countryCode` is not needed for Android Pay

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
```diff
@@ -0,0 +1,39 @@
...
+import android.os.Bundle;
+import android.support.annotation.Nullable;
+
+import com.shopify.buy3.pay.PayCart;
+
+public class UnityAndroidPayFragment extends Fragment {
+    private PayCart cart;
+    private String unityDelegateObjectName;
+
+    private static String EXTRA_UNITY_DELEGATE_OBJECT_NAME = "unityDelegateObjectName";
```
dengjeffrey: Is `EXTRA` an Android convention? Also maybe `private static final`?

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -0,0 +1,102 @@
...
+            String currencyCode,
+            String countryCode,
+            boolean requiresShipping) {
+
+        SummaryItems items = null;
+        try {
+            items = SummaryItems.fromJsonString(summaryItemsString);
+        } catch (JSONException e) {
+            Log.e("ShopifyBuyPlugin", "Failed to parse summary items from Unity");
+            e.printStackTrace();
```
sav007: you should use: `Log.e(tag, message, error)`

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
sav007: Hm, good question. Where do you get `countryCode`?

sav007: From docs of Android Pay:
```
MaskedWalletRequest.Builder | setCountryCode(String countryCode)Sets the ISO 3166-1 alpha-2 country code based on where the transaction was completed or processed.
-- | --
```

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -0,0 +1,102 @@
...
+                .merchantName(merchantName)
+                .currencyCode(currencyCode)
+                .shippingAddressRequired(requiresShipping)
+                .subtotal(items.subtotal)
+                .shippingPrice(items.shippingPrice)
+                .taxPrice(items.taxPrice)
+                .totalPrice(items.totalPrice)
+                .build();
+
+        addAndroidPayFragment(unityDelegateObjectName, cart);
```
sav007: nit: probably a good idea to call it `showAndroidPayFragment`

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -0,0 +1,102 @@
...
+
+        static SummaryItems fromJsonString(String jsonString) throws JSONException,
+                IllegalArgumentException  {
+            SummaryItems items = new SummaryItems();
+            JSONArray jsonArray = new JSONArray(jsonString);
+            for (int i = 0; i < jsonArray.length(); i++) {
+                JSONObject summaryItemObj = new JSONObject(jsonArray.getString(i));
+                String label = summaryItemObj.getString("Label");
+
+                if (label.equals("SUBTOTAL")) {
```
sav007: nit: usually such const like 'SUBTOTAL' extracted as real const `private static final String KEY_SUBTOTAL = "SUBTOTAL"`

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
sav007: Yes, makes sense. But maybe it should be `JSONException`?

sav007: And you can make it more cleaner:
create util function like `checkNotNull`
so your code will look like:
```
checkNotNull(items.subtotal, "...")
checkNotNull(items. taxPrice, "...")
checkNotNull(items. totalPrice, "...")
```

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -0,0 +1,102 @@
...
+import com.shopify.buy3.pay.PayCart;
+import com.unity3d.player.UnityPlayer;
+
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.math.BigDecimal;
+
+public class AndroidPayCheckoutSession {
```
sav007: nit: usually I make by default everything final (just to indicate that class wasn't designed for extension, and only when it need to be extended then think carefully how to do it right, as shared state is an evil)

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -0,0 +1,102 @@
...
+                UnityAndroidPayFragment.newInstance(unityDelegateObjectName, cart);
+        UnityPlayer.currentActivity.getFragmentManager()
+                .beginTransaction()
+                .add(payFragment, "payFragment")
+                .commit();
+    }
+
+    // Wrapper around SummaryItems from Unity that performs validation.
+    private static class SummaryItems {
+        private BigDecimal subtotal;
```
sav007: nit: omit `private` scope as it's redundant here the class itself private

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
sav007: Yeah, usually we call this in this way to indicate that this key belongs to parsing extras.

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
```diff
@@ -0,0 +1,39 @@
...
+    private static String EXTRA_PAY_CART = "payCart";
+
+    public static final UnityAndroidPayFragment newInstance(
+            String unityDelegateObjectName,
+            PayCart cart) {
+
+        UnityAndroidPayFragment fragment = new UnityAndroidPayFragment();
+        Bundle bundle = new Bundle(2);
+        bundle.putString(EXTRA_UNITY_DELEGATE_OBJECT_NAME, unityDelegateObjectName);
+        bundle.putParcelable(EXTRA_UNITY_DELEGATE_OBJECT_NAME, cart);
```
sav007: Where this bundle goes? Should it be `fragment .setArg(bundle)`

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
```diff
@@ -0,0 +1,39 @@
...
+        return fragment;
+    }
+
+    @Override
+    public void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        if (savedInstanceState != null) {
+            unityDelegateObjectName = savedInstanceState.getString(EXTRA_UNITY_DELEGATE_OBJECT_NAME);
```
sav007: savedInstanceState -> getArguments() ?

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
sleroux: Looks like I cherry picked that commit over by accident and shouldn't live in this branch. I'll keep this as is for now and rebase with the proper commit when the shop name branch lands.

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
sleroux: Thanks for the info. I'll pass this through to the fragment then since it's not needed on `PayCart`.

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
sleroux: Class is public so Unity can talk to it

sleroux:
While this was being reviewed, I pushed up some unit testing code and some small refactoring of the AndroidPayCheckoutSession class.

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
sav007: What do you mean?
`private static class SummaryItems`

My comment was about redundant private access modifiers for fields `private BigDecimal subtotal`

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -0,0 +1,133 @@
...
+
+        private static BigDecimal amountFromJson(JSONObject obj) throws JSONException {
+            return new BigDecimal(obj.getString("Amount"));
+        }
+
+        private static void checkNotNull(BigDecimal value, String missingField)
+                throws JSONException {
+            if (value == null) {
+                String msg = String.format("Missing '%s' field from summary items.", missingField);
+                throw new IllegalArgumentException(msg);
```
sav007: JSONException

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -0,0 +1,133 @@
...
+        private static final String SHIPPING = "SHIPPING";
+        private static final String TAXES = "TAXES";
+        private static final String TOTAL = "TOTAL";
+
+        static SummaryItems fromJsonString(String jsonString) throws JSONException {
+            SummaryItems items = new SummaryItems();
+            JSONArray jsonArray = new JSONArray(jsonString);
+            for (int i = 0; i < jsonArray.length(); i++) {
+                JSONObject summaryItemObj = new JSONObject(jsonArray.getString(i));
+                String label = summaryItemObj.getString("Label");
```
sav007: nit: `Label` should be as well defined as const

unity_buy_android/unity_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
sav007: But we need to figure out this, maybe we have an issue on MBSDK side do not accept country code.
So where this `countryCode ` comes from?

dengjeffrey: Messaged @sav007

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -0,0 +1,128 @@
...
+
+        private static BigDecimal amountFromJson(JSONObject obj) throws JSONException {
+            return new BigDecimal(obj.getString("Amount"));
+        }
+
+        private static void checkNotNull(BigDecimal value, String missingField)
+                throws JSONException {
+            if (value == null) {
+                String msg = String.format("Missing '%s' field from summary items.", missingField);
+                throw new IllegalArgumentException(msg);
```
sav007: `throw new JSONException(msg)`

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -0,0 +1,128 @@
...
+        private static final String SHIPPING = "SHIPPING";
+        private static final String TAXES = "TAXES";
+        private static final String TOTAL = "TOTAL";
+
+        static SummaryItems fromJsonString(String jsonString) throws JSONException {
+            SummaryItems items = new SummaryItems();
+            JSONArray jsonArray = new JSONArray(jsonString);
+            for (int i = 0; i < jsonArray.length(); i++) {
+                JSONObject summaryItemObj = new JSONObject(jsonArray.getString(i));
+                String label = summaryItemObj.getString("Label");
```
sav007: "Label" as const

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -0,0 +1,128 @@
...
+            e.printStackTrace();
+            return false;
+        }
+    }
+
+    public PayCart cartFromUnity(
+        String merchantName,
+        String summaryItemsString,
+        String currencyCode,
+        String countryCode,
```
sav007: Pls add TODO to fix it later with new version of `PayCart`. I'm going to add this and make a new release of pay support

sleroux:
@dengjeffrey 👍?

dengjeffrey:
I'll re-review once https://github.com/Shopify/unity-buy-sdk/pull/233 lands

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
```diff
@@ -0,0 +1,94 @@
...
+            if (checkout.requiresShipping()) {
+                try {
+                    summaryItems.Add(new SummaryItem("SHIPPING", checkout.shippingLine().price().ToString()));
+                } catch {}
+            }
+
+            summaryItems.Add(new SummaryItem("TAXES", checkout.totalTax().ToString()));
+            summaryItems.Add(new SummaryItem("TOTAL", checkout.totalPrice().ToString()));
+
+            return summaryItems;
```
dengjeffrey: This works fine, but what was the motivation to keep this a `List` and not a `Dictionary`?

sleroux: Not sure - this was mostly copied from the iOS function.

I agree though that it seems more natural to use a map data structure to hold this stuff like: 

```json
{
   "taxes": 3.21,
   "shippingPrice": 2.11
}
```
So I feel like this was built this way for a reason? I think it's fine for now though.

dengjeffrey: It was a List on iOS because Apple Pay takes an ordered list of summary items, but in this case Android Pay takes in individual values without the need for ordering

sleroux: Gotcha - I'll update to make it a map instead then since it's more natural in this context 👍

sleroux:
@dengjeffrey I ended up refactoring SummaryItem to be PricingLineItems and leverage Moshi to do the JSON instead of doing it myself. By doing that I don't really need to test the JSON conversion since Moshi just 'does' it.

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
```diff
@@ -0,0 +1,94 @@
...
+            }
+            #endif
+        }
+
+        private Dictionary<string, string> GetPricingLineItemsFromCheckout(Checkout checkout) {
+            var dict = new Dictionary<string, string>()
+            {
+                { "TAXES",      checkout.totalTax().ToString() },
+                { "SUBTOTAL",   checkout.subtotalPrice().ToString() },
+                { "TOTAL",      checkout.totalPrice().ToString() }
```
dengjeffrey: Quick question, do these keys need to match the naming of the instance variables when we deserialize using Moshi? `subtotal` `taxPrice` `shippingPrice`

sleroux: Yes! I'll update.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/models/PricingLineItems.java
```diff
@@ -0,0 +1,38 @@
...
+import java.math.BigDecimal;
+
+public final class PricingLineItems implements JsonSerializable {
+    @NonNull public final BigDecimal subtotal;
+    @NonNull public final BigDecimal taxPrice;
+    @NonNull public final BigDecimal totalPrice;
+
+    @Nullable public final BigDecimal shippingPrice;
+
+    private static JsonAdapter<PricingLineItems> jsonAdapter =
```
dengjeffrey: Should this be `final`?

PR #233, opened 2017-07-28
[Feature] Adds caching for Shop's name whenever we fetch PaymentSettings

sleroux:
Android Pay requires a merchant name so this patch appends to the existing GraphQL query we're making to get the payment settings to get the shop name.

### DIFF
--- a/scripts/generator/graphql_generator/csharp/Cart.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/Cart.cs.erb
@@ -76,6 +76,7 @@ namespace <%= namespace %> {
         private IWebCheckout WebCheckout;
         private INativeCheckout NativeCheckout;
         private PaymentSettings StorePaymentSettings;
+        private String ShopName;
 
         private ShopifyClient Client;
 
@@ -175,11 +176,11 @@ namespace <%= namespace %> {
                 if (error != null) {
                     failure(error);
                 } else {
-                    GetPaymentSettings((paymentSettings, paymentSettingsError) => {
+                    GetPaymentSettingsAndShopName((paymentSettings, shopName, paymentSettingsError) => {
                         if (paymentSettingsError != null) {
                             failure(paymentSettingsError);
                         } else {
-                            NativeCheckout.Checkout(key, paymentSettings, success, cancelled, failure);
+                            NativeCheckout.Checkout(key, shopName, paymentSettings, success, cancelled, failure);
                         }
                     });
                 }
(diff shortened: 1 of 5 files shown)


sleroux:
I wasn't sure if we should be caching the entire Shop object (which includes the payment settings) or just pull out the shop name into the existing GraphQL query. I added just the shop name to the query for now but if we think we'll need more than that from shop it might make sense to take a harder look to refactor`GetPaymentSettings` beyond just adding another parameter.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.cs.erb
```diff
@@ -67,7 +67,7 @@ namespace <%= namespace %>.SDK.iOS {
         /// <param name="success">Delegate method that will be notified upon a successful payment</param>
         /// <param name="failure">Delegate method that will be notified upon a failure during the checkout process</param>
         /// <param name="cancelled">Delegate method that will be notified upon a cancellation during the checkout process</param>
-        public void Checkout(string key, PaymentSettings paymentSettings, CheckoutSuccessCallback success, CheckoutCancelCallback cancelled, CheckoutFailureCallback failure) {
+        public void Checkout(string key, string shopName, PaymentSettings paymentSettings, CheckoutSuccessCallback success, CheckoutCancelCallback cancelled, CheckoutFailureCallback failure) {
```
dengjeffrey: I think it might be worthwhile to put `shopName` and `paymentSettings` into a struct `ShopMeta`

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -175,11 +183,11 @@ namespace <%= namespace %> {
...
                 } else {
-                    GetPaymentSettings((paymentSettings, paymentSettingsError) => {
-                        if (paymentSettingsError != null) {
-                            failure(paymentSettingsError);
+                    GetShopMetadata((metadata, metadataError) => {
+                        if (metadataError != null) {
+                            failure(metadataError);
                         } else {
-                            NativeCheckout.Checkout(key, paymentSettings, success, cancelled, failure);
+                            NativeCheckout.Checkout(key, metadata.Value.Name, metadata.Value.PaymentSettings, success, cancelled, failure);
```
dengjeffrey: Can we not do `NativeCheckout.Checkout(key, metaData, ...)`?

sleroux:
@dengjeffrey Pushed up with refactor to the Checkout API

PR #234, opened 2017-07-28
[Feature] Added AndroidWebCheckout to Unity SDK

dengjeffrey:
+ Adds implementation of Android Web Checkout to the managed code side

### DIFF
--- a/scripts/generator/graphql_generator/csharp/SDK/Android/AndroidWebCheckout.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/Android/AndroidWebCheckout.cs.erb
@@ -1,0 +1,45 @@
+namespace <%= namespace %>.SDK.Android {
+    #if UNITY_ANDROID
+    using System;
+    using <%= namespace %>.SDK;
+
+    #if !SHOPIFY_MONO_UNIT_TEST
+    using UnityEngine;
+    #endif
+
+    class AndroidWebCheckout: WebCheckout {
+
+        private ShopifyClient _client;
+        private Cart _cart;
+
+        protected override ShopifyClient Client {
+            get {
+                return _client;
+            }
+        }
+
+        protected override Cart Cart {
+            get {
+                return _cart;
+            }
+        }
+
+        private const string UnityPlayerClassName = "com.unity3d.player.UnityPlayer";
+        private const string WebCheckoutSessionClassName = "com.shopify.unity.buy.WebCheckoutSession";
+
+        public AndroidWebCheckout(Cart cart, ShopifyClient client) {
+            _cart = cart;
+            _client = client;
+        }
+
+        public override void Checkout(string checkoutURL, CheckoutSuccessCallback success, CheckoutCancelCallback cancelled, CheckoutFailureCallback failure) {
+            #if !SHOPIFY_MONO_UNIT_TEST
+            using(var webCheckoutSession = new AndroidJavaObject(WebCheckoutSessionClassName)) {
+                SetupNativeMessageReceiver(success, cancelled, failure);
+                webCheckoutSession.Call("checkout", checkoutURL);
+            }
+            #endif
+        }
+    }
+    #endif
+}
--- a/scripts/generator/graphql_generator/csharp.rb
+++ b/scripts/generator/graphql_generator/csharp.rb
@@ -127,6 +127,7 @@ module GraphQLGenerator
         SDK/INativeCheckout
         SDK/WebCheckout
         SDK/NativeWebViewMessageReceiver
+        SDK/Android/AndroidWebCheckout
         SDK/iOS/ApplePayAuthorizationStatus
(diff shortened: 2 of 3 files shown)


scripts/generator/graphql_generator/csharp/SDK/Android/AndroidWebCheckout.cs.erb
```diff
@@ -0,0 +1,41 @@
+namespace <%= namespace %>.SDK.Android {
+    #if UNITY_ANDROID
```
dengjeffrey: I put `#if UNITY_ANDROID` below the namespace because this is the only file defining the namespace

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidWebCheckout.cs.erb
```diff
@@ -0,0 +1,41 @@
...
+        private const string WebCheckoutSessionClassName = "com.shopify.unity.buy.WebCheckoutSession";
+
+        public AndroidWebCheckout(Cart cart, ShopifyClient client) {
+            _cart = cart;
+            _client = client;
+        }
+
+        public void Checkout(string checkoutURL, CheckoutSuccessCallback success, CheckoutCancelCallback cancelled, CheckoutFailureCallback failure) {
+            #if !SHOPIFY_MONO_UNIT_TEST
+            using(var webCheckoutSession = new AndroidJavaObject(WebCheckoutSessionClassName)) {
```
sleroux: One thing I would look into is exactly when this object will be destroyed. Does it get release when we fall out of the `Checkout` scope? Do we want to maintain a handle to the Java object while `AndroidWebCheckout` is still around?

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidWebCheckout.cs.erb
```diff
@@ -0,0 +1,45 @@
...
+    #if !SHOPIFY_MONO_UNIT_TEST
+    using UnityEngine;
+    #endif
+
+    class AndroidWebCheckout: WebCheckout {
+
+        private ShopifyClient _client;
+        private Cart _cart;
+
+        protected override ShopifyClient Client {
```
richardmonette: In https://github.com/Shopify/unity-buy-sdk/blob/77c1288da163ab499637d02a96ae2c5869719f25/scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb#L15 it looks like ShopifyClient and Cart are just sitting public, whereas here they are private with accessors. Is there a reason we have differing access for these between Android and iOS?

dengjeffrey: We've changed `IWebCheckout` to an abstract class, here is the most up to date file of `iOSWebCheckout`: https://github.com/Shopify/unity-buy-sdk/blob/beta/scripts/generator/graphql_generator/csharp/SDK/iOS/iOSWebCheckout.cs.erb

richardmonette: Would it make sense to make the base class not abstract, so the actual definitions could live there? (just wondering if we can reduce the duplication of the code for those functions)

dengjeffrey: Reason being for an abstract class was to reduce the class hierarchy, since we're not making too many of these classes I don't see duplication too much of an issue down the road, but flattening class hierarchy might make it more clear for readers 🤷‍♂️ 

> Would it make more sense to turn IWebCheckout into an abstract base class instead of an interface? At least that way we could flatten the class hierarchy a bit.

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidWebCheckout.cs.erb
dengjeffrey: `AndroidWebCheckout` will always be around as soon as `Cart` is instantiated, so we can't necessarily keep the reference till `AndroidWebCheckout` is gone. And since we don't provide any guarantees that only one `Checkout` is called at a time, I don't see the benefit of holding a reference to an unmanaged object. Are there any other concerns?

sleroux: If we're creating a session object in Unity to start the checkout, how do we communicate back to the Android side of things after we've done the validation check from Unity? You'd need to create a new instance or call a static method, no?

It might make sense to tie the lifecycle of the Android object with that of the checkout game object that we create to communicate between Android/Unity.

sleroux: Nevermind! We don't communicate back to the native side after validation so ignore this. Good to go.

PR #236, opened 2017-07-28
[Fix] Check publish.sh build's return value

dengjeffrey:
+ Adds a variable that records the build return value
+ Changes the if statement to check for that value instead of the return value of `restore_native_tests`


dengjeffrey:
Yep I noticed that too with most of our scripts. I'll make a patch to fix it

PR #240, opened 2017-07-31
[Feature] Android unity message

dengjeffrey:
+ Adds a container class for Unity Messages that serializes content for Unity to consume

### DIFF
--- a/android/shopify_buy_plugin/build.gradle
+++ b/android/shopify_buy_plugin/build.gradle
@@ -35,6 +35,7 @@ dependencies {
     compile 'com.android.support:appcompat-v7:26.0.0'
     compile 'com.android.support:customtabs:26.0.0'
     testCompile 'junit:junit:4.12'
+    testCompile 'org.json:json:20080701'
 }
 
 task copyAARToPlugins {
--- a/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityMessage.java
+++ b/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityMessage.java
@@ -27,7 +27,7 @@ public class UnityMessage implements JsonSerializable {
             json.put(IDENTIFIER_KEY, identifier);
             json.put(CONTENT_KEY, content);
         } catch (JSONException e) {
-            throw new IllegalStateException("Unity Message is not JsonSerializable");
+            throw new IllegalStateException("UnityMessage is not JsonSerializable");
         }
         return json.toString();
     }
--- a/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityMessage.java
+++ b/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityMessage.java
@@ -1,0 +1,22 @@
+package com.shopify.unity.buy;
+
+import com.google.gson.Gson;
+import java.util.UUID;
+
+public class UnityMessage implements JsonSerializable {
+
+    // Using capital case here, so when it is serialized the keys are capitalized
+    public String Content;
+    private String Identifier;
+
+    public UnityMessage(String content) {
+        Identifier = UUID.randomUUID().toString();
+        this.Content = content;
+    }
+
+    @Override
+    public String toJsonString() {
+        Gson gson = new Gson();
+        return gson.toJson(this);
+    }
+}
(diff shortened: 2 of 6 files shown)


android/shopify_buy_plugin/src/test/java/com/shopify/unity/buy/UnityMessageTest.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Map;
+
+import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.Assert.assertTrue;
+
+public class UnityMessageTest {
+
+    @Test
+    public void test_serialize() {
```
sleroux: Nit: Doesn't really matter but we can follow captial casing as a convention for naming test cases. Just going based off what we've done for the Mobile Buy SDK https://github.com/Shopify/mobile-buy-sdk-android/blob/master/MobileBuy/buy3-pay-support/src/androidTest/java/com/shopify/buy3/pay/PayCartTest.java

android/shopify_buy_plugin/build.gradle
```diff
@@ -28,6 +28,7 @@ android {
 
 dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
+    compile group: 'com.google.code.gson', name: 'gson', version: '2.3.1'
```
sleroux: Any reason in particular you're pulling in Gson? Android has a `JSONObject` class that you can use to do simple JSON serializing/deserializing.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityMessage.java
```diff
@@ -0,0 +1,22 @@
+package com.shopify.unity.buy;
+
+import com.google.gson.Gson;
+import java.util.UUID;
+
+public class UnityMessage implements JsonSerializable {
+
+    // Using capital case here, so when it is serialized the keys are capitalized
+    public String Content;
```
sleroux: Related to the checkstyle errors, we can probably make these lowercased and set the JSON object keys in the `toJsonString` function to be captialized - similar to how we've done in Unity.

android/shopify_buy_plugin/build.gradle
sleroux: Hmm I just realized that I don't think JSONObject is mocked when running unit tests.

dengjeffrey:
I made the requested changes, please 👀  again @sleroux

dengjeffrey:
Since we've discussed possibly using https://github.com/square/moshi instead of `gson` and JSONObject. I prefer that we come back and refactor later if we do choose to use one of those, since it's a minute implementation detail

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityMessage.java
```diff
@@ -0,0 +1,34 @@
+package com.shopify.unity.buy;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.util.UUID;
+
+public class UnityMessage implements JsonSerializable {
```
sav007: make it final to not accidentally extend it

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityMessage.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public static String CONTENT_KEY = "Content";
```
sav007: final

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityMessage.java
```diff
@@ -0,0 +1,34 @@
...
+    public static String IDENTIFIER_KEY = "Identifier";
+
+    public String content;
```
sav007: I think you can make it immutable as they are initialized only in constructor (make them final)

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityMessage.java
```diff
@@ -0,0 +1,34 @@
...
+    private String identifier;
```
sav007: should it be public as well?

dengjeffrey: No, this is a property that won't be used by any interface on Android

PR #244, opened 2017-08-01
[Feature] Added Android Web Intent Fragment

dengjeffrey:
+ Added a fragment that launches a `CustomTabs` intent for a URL

### DIFF
--- a/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/WebIntentFragment.java
+++ b/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/WebIntentFragment.java
@@ -23,11 +23,13 @@ public class WebIntentFragment extends Fragment {
     }
 
     public void launchUrl(String url) {
-        CustomTabsIntent.Builder builder = new CustomTabsIntent.Builder();
-        CustomTabsIntent customTabsIntent = builder.build();
+        if (isAdded() && !isRemoving()) {
+            CustomTabsIntent.Builder builder = new CustomTabsIntent.Builder();
+            CustomTabsIntent customTabsIntent = builder.build();
 
-        didShowIntent = true;
-        customTabsIntent.launchUrl(getActivity().getApplicationContext(), Uri.parse(url));
+            didShowIntent = true;
+            customTabsIntent.launchUrl(getActivity().getApplicationContext(), Uri.parse(url));
+        }
     }
 
     private void notifyListenerOnClose() {
--- a/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/WebIntentFragment.java
+++ b/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/WebIntentFragment.java
@@ -9,6 +9,7 @@ import android.support.customtabs.CustomTabsIntent;
 public final class WebIntentFragment extends Fragment {
 
     private static final String ARGS_URL = "url";
+    private boolean onPauseIssued = false;
     private WebIntentListener listener;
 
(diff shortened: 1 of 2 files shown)


android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/WebIntentFragment.java
```diff
@@ -0,0 +1,41 @@
+package com.shopify.unity.buy;
+
+import android.app.Fragment;
+import android.net.Uri;
+import android.support.customtabs.CustomTabsIntent;
+
+public class WebIntentFragment extends Fragment {
+
+    private boolean didShowIntent = false;
```
sav007: Did you try https://developer.android.com/reference/android/support/customtabs/CustomTabsCallback.html#onNavigationEvent(int, android.os.Bundle)
```
int TAB_HIDDEN
Sent when the tab becomes hidden.
```

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/WebIntentFragment.java
```diff
@@ -0,0 +1,41 @@
...
+    public void setListener(WebIntentListener listener) {
+        this.listener = listener;
+    }
+
+    public void launchUrl(String url) {
+        CustomTabsIntent.Builder builder = new CustomTabsIntent.Builder();
+        CustomTabsIntent customTabsIntent = builder.build();
+
+        didShowIntent = true;
+        customTabsIntent.launchUrl(getActivity().getApplicationContext(), Uri.parse(url));
```
sleroux: @sav007 Does it matter which context is used here (Activity vs Application)? Just curious

sav007: Depends, but if it's related to UI stuff (like Android Pay requires Activity context is it shows UI). So I guess here it's fine to get activity as context.

Agh one note here, as this is a public method make sure about lifecycle, add verification that fragment is not detached from the activity

sav007: Don't remember the exact life cycle, pls check docs for these:
- Fragment.isAdded() 
- Fragment.isDetached() 
- Fragment.isRemoving()

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/WebIntentFragment.java
dengjeffrey: After doing some exploration with `CustomTabsCallback`, `TAB_HIDDEN` is an event that is invoked when the user leaves the web view. This is something we don't want because we want to know when the user returns to the application, so we can make a check to see if the checkout has been completed.

For example, if the user presses the home button while the web view was shown, we will be receiving this event whilst the application is in the background.

dengjeffrey:
@sav007 Made changes as suggested

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/WebIntentFragment.java
sav007: Oh so you want to track when user actually returns back to application?

dengjeffrey: Yes, when they return to the application we check to see if they have completed the checkout via the web intent. Checking whether they finished the web checkout on `TAB_HIDDEN` may be misleading, because they might end up navigating back to the web view.

PR #246, opened 2017-08-02
[Fix] Adds UNITY_IPHONE checks to build pipeline scripts

sleroux:
On machines that don't support the iOS player (like Windows), the package won't compile because of errors with the iOS-only build pipeline scripts.


sleroux:
@ThereExistsX Since you're on a Windows machine, would I be able to drop you a .unitypackage from this PR to 🎩?

ThereExistsX:
@sleroux Send it on over! =)

ThereExistsX:
LGTM

dengjeffrey:
🚢

sleroux:
At some point this will pass the activate_unity script and pass CI :/

PR #248, opened 2017-08-02
[Refactor] Check setup script

dengjeffrey:
+ Fixes an issue where the setup on machine is always False
+ changed `echo` to `printf` to use escape characters and fix lint-ing issues

### DIFF
--- a/scripts/check_setup.sh
+++ b/scripts/check_setup.sh
@@ -35,45 +35,66 @@ else
     HAS_OPTIONAL=false
 fi
 
+print_bold_text() {
+    # arg 1: String
+    # arg 2: Color
+    printf "%b$1%b" "$2$BOLD" "$RESET_STYLE"
+}
+
+print_point_form_prefix() {
+    # arg 1: String
+    printf "◦ %s : " "$(print_bold_text "$1")"
+}
+
 # Output required dependencies
 if [[ $HAS_REQUIRED = false ]] ; then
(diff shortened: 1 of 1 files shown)


scripts/check_setup.sh
```diff
@@ -33,46 +36,50 @@ else
 fi
 
 # Output required dependencies
-if [ $HAS_REQUIRED = false ] ; then
-    echo "\n\n${COLOR_RED}${BOLD}The following required required dependencies are missing:${RESET_STYLE}\n"
+if [[ $HAS_REQUIRED = false ]] ; then
+    printf "\n\n $COLOR_RED $BOLD The following required required dependencies are missing: $RESET_STYLE \n"
```
sleroux: Nit: I'm getting a lint warning on all these `printf` calls to avoid using the variables directly in the string and use a formatter string instead like:

```bash
printf "\n\n %s %s The following required required dependencies are missing: %s \n" "$COLOR_GREEN" "$BOLD" "$RESET_STYLE"
```

Does that make more sense or do we want to leave it?

dengjeffrey: Looking into it more, didnt know we could use `%b` to evaluate the variables we are substituting

dengjeffrey:
@sleroux I made a lot of changes, please 👀 again 😬

sleroux:
👍

PR #249, opened 2017-08-02
[Feature] Android web checkout session

dengjeffrey:
+ Adds the layer that Unity talks to, to launch a web intent with a checkout URL

### DIFF
--- a/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/WebCheckoutSession.java
+++ b/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/WebCheckoutSession.java
@@ -1,0 +1,58 @@
+package com.shopify.unity.buy;
+
+import android.app.FragmentManager;
+import android.app.FragmentTransaction;
+
+import com.unity3d.player.UnityPlayer;
+
+public class WebCheckoutSession implements WebIntentListener {
+
+    private static final String WEB_FRAGMENT_TAG = "webFragment";
+    private static final String NATIVE_WEB_DELEGATE_METHOD_NAME = "OnNativeMessage";
+    private static final String NATIVE_WEB_DELEGATE_METHOD_CONTENT = "dismissed";
+    private String unityDelegateObjectName;
+
+    public void checkout(String unityDelegateObjectName, String url) {
+        this.unityDelegateObjectName = unityDelegateObjectName;
+        launchWebIntent(url);
+    }
+
+    @Override
+    public void onWebIntentClose() {
+        FragmentManager manager = getFragmentManager();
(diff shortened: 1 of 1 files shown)


android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/WebCheckoutSession.java
```diff
@@ -0,0 +1,58 @@
...
+        UnityPlayer.UnitySendMessage(unityDelegateObjectName, NATIVE_WEB_DELEGATE_METHOD_NAME, dismissedMessage.toJsonString());
+    }
+
+    private FragmentManager getFragmentManager() {
+        return UnityPlayer.currentActivity.getFragmentManager();
+    }
+
+    private void launchWebIntent(final String checkoutUrl) {
+        final FragmentManager manager = getFragmentManager();
+        final FragmentTransaction transaction = getFragmentManager().beginTransaction();
```
sleroux: Nit: You can just use the local variable manager her instead of calling `getFragmentManager()`

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/WebCheckoutSession.java
```diff
@@ -0,0 +1,58 @@
...
+
+        WebIntentFragment fragment = WebIntentFragment.newInstance(checkoutUrl);
+        fragment.setListener(this);
+        transaction.add(fragment, WEB_FRAGMENT_TAG);
+
+        final Runnable launchWebView = new Runnable() {
+            public void run() {
+                if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.N) {
```
sleroux: Interesting - is this the best practice for adding a fragment? I should probably use this pattern for the fragment I'm using for Android Pay.

dengjeffrey: I'm not exactly sure if this is the best practice, but `manager.executePendingTransactions()` must be invoked from the main thread

sav007: not sure that you need this, I think calling `commit` should be fine. The difference is that `commitNow` is sync and `commit` as async. But can't see what the benefit of that for your case

dengjeffrey: You're right, we don't need this now that `WebIntentFragment` explicitly launches the itnent`onCreate`

dengjeffrey:
@sav007 I removed the `commitNow`

PR #250, opened 2017-08-03
[Android Pay] Launches wallet chooser UI when checking out using Native Pay from Unity

sleroux:
* Adds Google API client and wallet listeners to UnityAndroidPayFragment to launch a masked wallet request.

### DIFF
--- a/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
+++ b/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
@@ -58,8 +58,7 @@ public final class AndroidPayCheckoutSession {
 
             return true;
         } catch (IOException e) {
-            logger.error("ShopifyBuyPlugin", "Failed to parse summary items from Unity");
-            e.printStackTrace();
+            logger.error("ShopifyBuyPlugin", "Failed to parse summary items from Unity!");
             return false;
         }
     }
--- a/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
+++ b/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
@@ -3,6 +3,7 @@ package com.shopify.unity.buy;
 import com.google.android.gms.wallet.WalletConstants;
 import com.shopify.buy3.pay.PayCart;
 import com.shopify.buy3.pay.PayHelper;
+import com.shopify.unity.buy.models.MailingAddressInput;
 import com.shopify.unity.buy.models.PricingLineItems;
 import com.shopify.unity.buy.utils.AndroidLogger;
 import com.shopify.unity.buy.utils.ILogger;
@@ -10,7 +11,7 @@ import com.unity3d.player.UnityPlayer;
 
 import java.io.IOException;
 
-public final class AndroidPayCheckoutSession {
+public final class AndroidPayCheckoutSession implements AndroidPaySessionCallbacks {
     private ILogger logger;
 
     public AndroidPayCheckoutSession() {
@@ -20,6 +21,9 @@ public final class AndroidPayCheckoutSession {
     AndroidPayCheckoutSession(ILogger logger) {
         this.logger = logger;
     }
+
+    private UnityAndroidPayFragment payFragment;
+    private String unityDelegateObjectName;
 
     //CHECKSTYLE:OFF
     public boolean checkoutWithAndroidPay(
@@ -42,19 +46,9 @@ public final class AndroidPayCheckoutSession {
             PayCart cart = cartFromUnity(merchantName, pricingLineItemsString, currencyCode, countryCode,
                     requiresShipping);
 
-            UnityAndroidPayFragment payFragment = UnityAndroidPayFragment.builder()
-                    .setUnityDelegateObjectName(unityDelegateObjectName)
-                    .setPayCart(cart)
-                    .setCountryCode(countryCode)
-                    .setEnvironment(testing ?
-                            WalletConstants.ENVIRONMENT_TEST : WalletConstants.ENVIRONMENT_PRODUCTION)
-                    .setPublicKey(publicKey)
-                    .build();
+            this.unityDelegateObjectName = unityDelegateObjectName;
 
-            UnityPlayer.currentActivity.getFragmentManager()
-                    .beginTransaction()
-                    .add(payFragment, "payFragment")
-                    .commit();
+            addPayFragment(cart, countryCode, publicKey, testing);
 
             return true;
         } catch (IOException e) {
@@ -83,5 +77,61 @@ public final class AndroidPayCheckoutSession {
                 .taxPrice(items.taxPrice)
                 .totalPrice(items.totalPrice)
                 .build();
+    }
+
+    private void addPayFragment(PayCart cart, String countryCode, String publicKey, boolean testing) {
+        if (payFragment != null) {
+            removePayFragment();
+        }
+
+        payFragment = UnityAndroidPayFragment.builder()
+            .setPayCart(cart)
+            .setCountryCode(countryCode)
+            .setEnvironment(testing ?
+                WalletConstants.ENVIRONMENT_TEST : WalletConstants.ENVIRONMENT_PRODUCTION)
+            .setPublicKey(publicKey)
+            .build();
+        payFragment.setSessionCallbacks(this);
+
+        UnityPlayer.currentActivity.getFragmentManager()
+            .beginTransaction()
+            .add(payFragment, "payFragment")
+            .commit();
+    }
+
+    private void removePayFragment() {
+        UnityPlayer.currentActivity.getFragmentManager()
+            .beginTransaction()
+            .remove(payFragment)
+            .commit();
+    }
+
+    public void onUpdateShippingAddress(MailingAddressInput address, MessageCenter.MessageCallbacks messageCallbacks) {
+        UnityMessage msg = UnityMessage.fromAndroid(address.toJsonString());
+        MessageCenter.UnityMessageReceiver receiver = new MessageCenter.UnityMessageReceiver(
+            unityDelegateObjectName,
+            MessageCenter.Method.ON_UPDATE_SHIPPING_ADDRESS
+        );
+        MessageCenter.sendMessageTo(msg, receiver, messageCallbacks);
+    }
+
+    public void onError(String error) {
+        removePayFragment();
+        UnityMessage msg = UnityMessage.fromAndroid(error);
+        MessageCenter.UnityMessageReceiver receiver = new MessageCenter.UnityMessageReceiver(
+            unityDelegateObjectName,
+            MessageCenter.Method.ON_ERROR
+        );
+        MessageCenter.sendMessageTo(msg, receiver);
+    }
+
+    public void onCancel() {
+        removePayFragment();
+        UnityMessage msg = UnityMessage.fromAndroid("");
+        MessageCenter.UnityMessageReceiver receiver = new MessageCenter.UnityMessageReceiver(
+            unityDelegateObjectName,
+            MessageCenter.Method.ON_CANCEL
+        );
+        MessageCenter.sendMessageTo(msg, receiver);
     }
 }
--- a/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
+++ b/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
@@ -1,6 +1,8 @@
 package com.shopify.unity.buy;
 
+import com.google.android.gms.wallet.WalletConstants;
 import com.shopify.buy3.pay.PayCart;
+import com.shopify.buy3.pay.PayHelper;
 import com.shopify.unity.buy.models.PricingLineItems;
 import com.shopify.unity.buy.utils.AndroidLogger;
 import com.shopify.unity.buy.utils.ILogger;
@@ -15,22 +17,45 @@ public final class AndroidPayCheckoutSession {
         this.logger = new AndroidLogger();
     }
 
-    public AndroidPayCheckoutSession(ILogger logger) {
+    AndroidPayCheckoutSession(ILogger logger) {
         this.logger = logger;
     }
 
+    //CHECKSTYLE:OFF
     public boolean checkoutWithAndroidPay(
             String unityDelegateObjectName,
             String merchantName,
+            String publicKey,
             String pricingLineItemsString,
             String currencyCode,
             String countryCode,
-            boolean requiresShipping
+            boolean requiresShipping,
+            boolean testing
     ) {
+        //CHECKSTYLE:ON
+        if (!PayHelper.isAndroidPayEnabledInManifest(UnityPlayer.currentActivity)) {
+            // TODO: Send unsupported error to Unity
+            return false;
+        }
+
         try {
             PayCart cart = cartFromUnity(merchantName, pricingLineItemsString, currencyCode, countryCode,
                     requiresShipping);
-            addAndroidPayFragment(unityDelegateObjectName, cart, countryCode);
+
+            UnityAndroidPayFragment payFragment = UnityAndroidPayFragment.builder()
+                    .setUnityDelegateObjectName(unityDelegateObjectName)
+                    .setPayCart(cart)
+                    .setCountryCode(countryCode)
+                    .setEnvironment(testing ?
+                            WalletConstants.ENVIRONMENT_TEST : WalletConstants.ENVIRONMENT_PRODUCTION)
+                    .setPublicKey(publicKey)
+                    .build();
+
+            UnityPlayer.currentActivity.getFragmentManager()
+                    .beginTransaction()
+                    .add(payFragment, "payFragment")
+                    .commit();
+
             return true;
         } catch (IOException e) {
             logger.error("ShopifyBuyPlugin", "Failed to parse summary items from Unity");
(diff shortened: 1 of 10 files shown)


android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -15,22 +17,45 @@ public AndroidPayCheckoutSession() {
         this.logger = new AndroidLogger();
     }
 
-    public AndroidPayCheckoutSession(ILogger logger) {
+    AndroidPayCheckoutSession(ILogger logger) {
         this.logger = logger;
     }
 
+    //CHECKSTYLE:OFF
```
sleroux: Checkstyle doesn't like that we have more than 7 method parameters but turning this into a JSON object instead just to get parameters from Unity doesn't seem necessary.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -15,22 +17,45 @@ public AndroidPayCheckoutSession() {
...
+            // TODO: Send unsupported error to Unity
+            return false;
+        }
+
         try {
             PayCart cart = cartFromUnity(merchantName, pricingLineItemsString, currencyCode, countryCode,
                     requiresShipping);
-            addAndroidPayFragment(unityDelegateObjectName, cart, countryCode);
+
+            UnityAndroidPayFragment payFragment = UnityAndroidPayFragment.builder()
```
sleroux: Since we add the fragment in one place I moved this inline with this method since this method isn't long anyways.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
```diff
@@ -1,33 +1,74 @@
...
+
+    private MaskedWallet maskedWallet;
 
     private static final String EXTRA_UNITY_DELEGATE_OBJECT_NAME = "unityDelegateObjectName";
     private static final String EXTRA_PAY_CART = "payCart";
     private static final String EXTRA_COUNTRY_CODE = "countryCode";
+    private static final String EXTRA_ANDROID_PAY_ENVIRONMENT = "androidPayEnvironment";
+    private static final String EXTRA_PUBLIC_KEY = "publicKey";
+
+    static final class UnityAndroidPayFragmentBuilder {
```
sleroux: Refactored `newInstance` into a builder object now that we're passing along more information.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
```diff
@@ -39,9 +80,68 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
+            androidPublicKey = bundle.getString(EXTRA_PUBLIC_KEY);
         }
 
         // Don't recreate this fragment during configuration change.
         setRetainInstance(true);
+
+        googleApiClient = new GoogleApiClient.Builder(this.getActivity())
+                .addApi(Wallet.API, new Wallet.WalletOptions.Builder()
+                        .setEnvironment(androidPayEnvironment)
+                        .setTheme(WalletConstants.THEME_DARK)
```
sleroux: Coded to be dark theme for now. If developers ask to skin this in the future we can make that change then.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
dengjeffrey: I love the builder pattern, but it feels icky since all the parameters are required. Would a `Step Builder` pattern make this more robust?

dengjeffrey: Alternatively I think you might be able to get away with checking that `bundle` contains all the keys in a predefined set, since all these parameters and new ones will be required

sleroux: Step builder would probably add too much bloat since you need to define differing return types for each setter and I'm not concerned with explicit ordering here. Adding the run down list of all properties prior to building and throwing an exception if everything isn't set sounds pretty solid 👍

sleroux:
@dengjeffrey Updated with bundle validation check

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
```diff
@@ -1,33 +1,99 @@
...
+            UnityAndroidPayFragment fragment = new UnityAndroidPayFragment();
+            fragment.setArguments(bundle);
+            return fragment;
+        }
+
+        private void checkBundleContainsRequiredExtras() {
+            List<String> missingExtras = new ArrayList<>();
+            for (String extra : requiredExtras) {
+                if (bundle.get(extra) == null) {
+                    missingExtras.add(extra);
```
dengjeffrey: nice!

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -0,0 +1,88 @@
...
+
+            UnityPlayer.currentActivity.getFragmentManager()
+                    .beginTransaction()
+                    .add(payFragment, "payFragment")
+                    .commit();
+
+            return true;
+        } catch (IOException e) {
+            logger.error("ShopifyBuyPlugin", "Failed to parse summary items from Unity");
+            e.printStackTrace();
```
sav007: shouldn't it be part of `logger.error` ?

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -0,0 +1,88 @@
...
+
+    PayCart cartFromUnity(
+        String merchantName,
+        String pricingLineItemsString,
+        String currencyCode,
+
+        // TODO: Pull in updated buy3 SDK which will use the countryCode.
+        String countryCode,
+        boolean requiresShipping
+    ) throws IOException {
```
sav007: is it true, that it throws IOException?
seems JsonExcepton would be more appropriate here instead

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
```diff
@@ -0,0 +1,172 @@
...
+    private int androidPayEnvironment;
+    private GoogleApiClient googleApiClient;
+
+    private MaskedWallet maskedWallet;
+
+    private static final String EXTRA_UNITY_DELEGATE_OBJECT_NAME = "unityDelegateObjectName";
+    private static final String EXTRA_PAY_CART = "payCart";
+    private static final String EXTRA_COUNTRY_CODE = "countryCode";
+    private static final String EXTRA_ANDROID_PAY_ENVIRONMENT = "androidPayEnvironment";
+    private static final String EXTRA_PUBLIC_KEY = "publicKey";
```
sav007: Not sure how this passed the checkstyle, but the order usually first consts and then variables

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
```diff
@@ -0,0 +1,172 @@
...
+
+    static final class UnityAndroidPayFragmentBuilder {
+        private Bundle bundle;
+        private String[] requiredExtras = {
```
sav007: final static ?
And the order should be before `private Bundle bundle;`

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
```diff
@@ -0,0 +1,172 @@
...
+    @Override
+    public void onStop() {
+        googleApiClient.disconnect();
+        super.onStop();
+    }
+
+    @Override
+    public void onConnected(@Nullable Bundle bundle) {
+        // On connection, request for a masked wallet.
+        PayHelper.requestMaskedWallet(googleApiClient, cart, androidPublicKey);
```
sav007: Is this right behaviour, every time when we connected to play services we will request masked wallet. Keep in mind that this will happen I guess even in case when user presses home button.
Activity will go to background `onStop` will be called, `googleApiClient.disconnect` will be called as well. Then user brings app back, `onStart()` -> `googleApiClient.connect()` -> `PayHelper.requestMaskedWallet(googleApiClient, cart, androidPublicKey)`.

Just asking if this flow is expected.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityMessage.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.UUID;
+
+public final class UnityMessage implements JsonSerializable {
+
+    public static final String CONTENT_KEY = "Content";
+    public static final String IDENTIFIER_KEY = "Identifier";
+
+    public final String content;
+    private final String identifier;
```
sav007: nit: just for symmetry should it be public as well?

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
sleroux: I've refactored this in another PR to get rid of the custom logger wrapper class and just use Android's Log class.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
sleroux: Moshi throws an `IOSException` when it fails to do the JSON conversion 🤷🏼‍♂️

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
sleroux: That doesn't sound good. I think the best way would be to make the request masked wallet call be explicit from the `AndroidPayCheckoutSession` so we're not leaning on the Google client/Fragment's lifecycle for when we show the wallet chooser. I'll investigate this.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -1,45 +1,63 @@
...
-public final class AndroidPayCheckoutSession {
+public final class AndroidPayCheckoutSession implements AndroidPaySessionCallbacks {
     private ILogger logger;
 
     public AndroidPayCheckoutSession() {
         this.logger = new AndroidLogger();
     }
 
-    public AndroidPayCheckoutSession(ILogger logger) {
+    AndroidPayCheckoutSession(ILogger logger) {
```
sleroux: This will be removed in the unit testing patch I have up. Since I'm using instrument tests for this I'm changing this back to using Android's `Log` class.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/MessageCenter.java
```diff
@@ -0,0 +1,81 @@
...
+
+import android.os.Handler;
+import android.os.HandlerThread;
+
+import com.unity3d.player.UnityPlayer;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class MessageCenter {
```
sleroux: Note: This code is from https://github.com/Shopify/unity-buy-sdk/pull/251

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
```diff
@@ -36,12 +105,89 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
+
+    @Override
+    public void onStop() {
+        googleApiClient.disconnect();
+        super.onStop();
+    }
+
+    @Override
+    public void onConnected(@Nullable Bundle bundle) {
+        if (currentCheckoutState == CheckoutState.READY) {
```
sleroux: Added a state enum to check to see if we've already called for a request to get a wallet. In cass of a reconnecting google client in the case of backgrounding the app I hope to use this state to continue where we left off instead of making a new masked wallet request.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityMessage.java
```diff
@@ -11,13 +11,30 @@
...
+    private UnityMessage(String content) {
+        this(UUID.randomUUID().toString(), content);
+    }
+
+    private UnityMessage(String identifier, String content) {
+        this.identifier = identifier;
         this.content = content;
     }
 
+    public static UnityMessage fromUnity(String message) {
```
sleroux: https://github.com/Shopify/unity-buy-sdk/pull/251

sleroux:
@sav007 I've landed https://github.com/Shopify/unity-buy-sdk/pull/251 onto this branch and did some small refactoring in how we start making requests and how we send requests to Unity.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -61,12 +79,59 @@ public PayCart cartFromUnity(
                 .build();
     }
 
-    private void addAndroidPayFragment(String unityDelegateObjectName, PayCart cart, String countryCode) {
-        UnityAndroidPayFragment payFragment =
-                UnityAndroidPayFragment.newInstance(unityDelegateObjectName, cart, countryCode);
+    private void addPayFragment(PayCart cart, String countryCode, String publicKey, boolean testing) {
+        if (payFragment != null) {
```
sav007: would be much safer if you use this:
```
Fragment payFragment = UnityPlayer.currentActivity.getFragmentManager().findFragrmentByTag("payFragment");
if (fragment != null) {
  removePayFragment(payFragment);
}
```

So in this way you don't need to rely on local field `payFragment`. There will be only one source of truth, FragmentManager from UnityPlayer.currentActivity

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -61,12 +79,59 @@ public PayCart cartFromUnity(
...
+        UnityMessage msg = UnityMessage.fromAndroid(address.toJsonString());
+        MessageCenter.UnityMessageReceiver receiver = new MessageCenter.UnityMessageReceiver(
+            unityDelegateObjectName,
+            MessageCenter.Method.ON_UPDATE_SHIPPING_ADDRESS
+        );
+        MessageCenter.sendMessageTo(msg, receiver, messageCallbacks);
+    }
+
+    public void onError(String error) {
+        removePayFragment();
```
sav007: won't it be NPE if payFragment is null? Can it be such case?

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -61,12 +79,59 @@ public PayCart cartFromUnity(
...
+    }
+
+    private void removePayFragment() {
+        UnityPlayer.currentActivity.getFragmentManager()
+            .beginTransaction()
+            .remove(payFragment)
+            .commit();
+    }
+
+    public void onUpdateShippingAddress(MailingAddressInput address, MessageCenter.MessageCallbacks messageCallbacks) {
```
sav007: is `public` here is intentional?

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -61,12 +79,59 @@ public PayCart cartFromUnity(
...
+        UnityMessage msg = UnityMessage.fromAndroid(address.toJsonString());
+        MessageCenter.UnityMessageReceiver receiver = new MessageCenter.UnityMessageReceiver(
+            unityDelegateObjectName,
+            MessageCenter.Method.ON_UPDATE_SHIPPING_ADDRESS
+        );
+        MessageCenter.sendMessageTo(msg, receiver, messageCallbacks);
+    }
+
+    public void onError(String error) {
```
sav007: is `public` here is intentional?

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPayCheckoutSession.java
```diff
@@ -61,12 +79,59 @@ public PayCart cartFromUnity(
...
+        removePayFragment();
+        UnityMessage msg = UnityMessage.fromAndroid(error);
+        MessageCenter.UnityMessageReceiver receiver = new MessageCenter.UnityMessageReceiver(
+            unityDelegateObjectName,
+            MessageCenter.Method.ON_ERROR
+        );
+        MessageCenter.sendMessageTo(msg, receiver);
+    }
+
+    public void onCancel() {
```
sav007: is `public` here is intentional?

sleroux: Yea because these implement the `AndroidPaySessionCallbacks` interface.

PR #251, opened 2017-08-03
[Android Pay] Adds plumbing for sending and receiving messages to and from Unity

sleroux:
* Adds `MessageCenter` class that handles sending messages back and forth with Unity. 
* Adds a wrapper class around `PayAddress` that allows it to be turned into a JSON string for use in Unity.
* Adds error formatter for converting error codes into user-readable strings in Unityland.

Currently branched off https://github.com/Shopify/unity-buy-sdk/pull/250 but has no real dependency on it.


android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/models/MailingAddressInput.java
```diff
@@ -0,0 +1,22 @@
...
+import com.shopify.buy3.pay.PayAddress;
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.Moshi;
+
+// A wrapper around PayAddress to convert it to Json.
+public class MailingAddressInput {
+    private final PayAddress payAddress;
+
+    private static final JsonAdapter<PayAddress> jsonAdapter =
+            new Moshi.Builder().build().adapter(PayAddress.class);
```
dengjeffrey: nit: Extra spaces 🤔 ?

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/MessageCenter.java
```diff
@@ -0,0 +1,77 @@
...
+
+import android.os.Handler;
+import android.os.HandlerThread;
+
+import com.unity3d.player.UnityPlayer;
+
+import java.util.HashMap;
+import java.util.Map;
+
+class MessageCenter {
```
sav007: should it be public?

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityMessage.java
```diff
@@ -11,13 +11,30 @@
...
+        this.identifier = identifier;
         this.content = content;
     }
 
+    public static UnityMessage fromUnity(String message) {
+        try {
+            JSONObject json = new JSONObject(message);
+            return new UnityMessage(json.getString(IDENTIFIER_KEY), json.getString(CONTENT_KEY));
+        } catch (JSONException e) {
+            throw new IllegalStateException("UnityMessage from Unity is invalid -- Message: " + message);
```
sav007: I guess it would more suitable to throw `IllegalArgumentException`

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/MessageCenter.java
```diff
@@ -0,0 +1,77 @@
...
+            @Override
+            public void run() {
+                UnityMessage msg = UnityMessage.fromUnity(result);
+                MessageCallbacks callbacks = callbacksInWaiting.remove(msg.identifier);
+                callbacks.onResponse(msg.content);
+            }
+        });
+    }
+
+    static class MessageReceiver {
```
sav007: This part is confusing, does it represent the call that must be performed on Unity side?
If yes, I would suggest to give a better name for this like: `UnityMessageReceiver`, what do you think?

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/MessageCenter.java
```diff
@@ -0,0 +1,77 @@
...
+            }
+        });
+    }
+
+    // Send message with callbacks to invoke when complete.
+    static void sendMessageTo(final UnityMessage msg, final MessageReceiver receiver, final MessageCallbacks callbacks) {
+        messageHandler.post(new Runnable() {
+            @Override
+            public void run() {
+                callbacksInWaiting.put(msg.identifier, callbacks);
```
sav007: What if we never received response from Unity?
It can lead to memory leak, as you keep a strong reference to callback in static class

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
```diff
@@ -124,23 +125,48 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+                UnityMessage msg = UnityMessage.fromAndroid(input.toJsonString());
+                MessageCenter.MessageReceiver receiver = new MessageCenter.MessageReceiver(
+                    unityDelegateObjectName,
+                    MessageCenter.Method.ON_UPDATE_SHIPPING_ADDRESS
+                );
+
+                MessageCenter.sendMessageTo(msg, receiver, new MessageCenter.MessageCallbacks() {
+                    @Override
+                    public void onResponse(String jsonResult) {
+                        UnityAndroidPayFragment.this.maskedWallet = maskedWallet;
```
sav007: Here can be a mem leak that I mentioned above, so if unity won't reply on message, the callback that we created will be always in memory and this callback keep strong reference to activity (UnityAndroidPayFragment.this will create implicit internal reference)

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/utils/ErrorFormatter.java
```diff
@@ -0,0 +1,32 @@
+package com.shopify.unity.buy.utils;
+
+import com.google.android.gms.wallet.WalletConstants;
+
+public final class ErrorFormatter {
```
sav007: Probably better naming is to be `WalletErrorFormatter` as if I'm not wrong this only for android pay wallet errors

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/models/MailingAddressInput.java
sleroux: Android Studio keeps wanting to use this indent - it's kind of annoying :/

sleroux:
@sav007 Updated weak refs for hash map and fragment ref

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/MessageCenter.java
```diff
@@ -0,0 +1,80 @@
...
+
+import java.lang.ref.WeakReference;
+import java.util.HashMap;
+import java.util.Map;
+
+public class MessageCenter {
+    private static final HandlerThread handlerThread = new HandlerThread("UnityMessageThread");
+    private static final Handler messageHandler = new Handler(handlerThread.getLooper());
+
+    private static Map<String, WeakReference<MessageCallbacks>> callbacksInWaiting = new HashMap<>();
```
sav007: Sorry for miscommunication but this is redundant. 
As there won't be a strong ref to anonymous  callback created in the `UnityAndroidPayFragment` like this:
```
MessageCenter.sendMessageTo(msg, receiver, new MessageCenter.MessageCallbacks() {
```

In this case this callback will be GC right away. That's said you should either rollback this changes to `Map<String, MessageCallbacks>` or make sure you keep strong ref to callback in your `UnityAndroidPayFragment`

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/models/MailingAddressInput.java
sav007: it's a setup in AS codestyle

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/MessageCenter.java
sleroux: I thought so but wasn't sure. I'll remove the weak map in favor of the weakified fragment ref.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
```diff
@@ -1,33 +1,78 @@
...
+    private int androidPayEnvironment;
+    private GoogleApiClient googleApiClient;
+
+    private MaskedWallet maskedWallet;
 
     private static final String EXTRA_UNITY_DELEGATE_OBJECT_NAME = "unityDelegateObjectName";
     private static final String EXTRA_PAY_CART = "payCart";
     private static final String EXTRA_COUNTRY_CODE = "countryCode";
+    private static final String EXTRA_ANDROID_PAY_ENVIRONMENT = "androidPayEnvironment";
+    private static final String EXTRA_PUBLIC_KEY = "publicKey";
```
sav007: The order of this section const declarations usually should be before variables.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/models/MailingAddressInput.java
```diff
@@ -0,0 +1,22 @@
+package com.shopify.unity.buy.models;
+
+import com.google.android.gms.identity.intents.model.UserAddress;
+import com.shopify.buy3.pay.PayAddress;
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.Moshi;
+
+// A wrapper around PayAddress to convert it to Json.
+public class MailingAddressInput {
+    private final PayAddress payAddress;
```
sav007: nit: making it public

PR #254, opened 2017-08-03
[Feature] Asset Store Build

mikkoh:
This PR implements having the ability to generate different source for the Unity Buy SDK based on the distribution platform we hit. Mainly we do this for tracking purposes so we can see where signups come from. For instance the `ShopifyOnboardingPanel.cs` looks like this when distributing to Github:
```
if (GUILayout.Button("Start your free trial", Styles.ButtonStyle))
            Application.OpenURL("https://www.shopify.ca/signup?signup_types=unity_buy_sdk,github");
```
And for `asset-store`:
```
if (GUILayout.Button("Start your free trial", Styles.ButtonStyle))
            Application.OpenURL("https://www.shopify.ca/signup?signup_types=unity_buy_sdk,asset-store");
```

### DIFF
--- a/scripts/generate.sh
+++ b/scripts/generate.sh
@@ -1,7 +1,7 @@
 . $(dirname $0)/common.sh
 
 bundle --gemfile=$SCRIPTS_ROOT/generator/Gemfile
-$SCRIPTS_ROOT/generator/update_schema
+$SCRIPTS_ROOT/generator/update_schema $1
 
 if [ $? = 0 ] ; then
     echo "Generate finished"
--- a/scripts/generator/graphql_generator/csharp.rb
+++ b/scripts/generator/graphql_generator/csharp.rb
@@ -20,12 +20,17 @@ module GraphQLGenerator
       end
     end
 
-    attr_reader :scalars, :schema, :namespace, :version
+    attr_reader :scalars, :schema, :namespace, :version, :publish_destination
 
-    def initialize(schema, namespace:, custom_scalars: [], script_name: $0)
+    def initialize(schema, publish_destination, namespace:, custom_scalars: [], script_name: $0)
+      if publish_destination != "asset-store" && publish_destination != "github"
+        raise "publish_destination is invalid valid values are: asset-store or github"
+      end
+
       @schema = schema
       @namespace = namespace
       @script_name = script_name
+      @publish_destination = publish_destination
 
       @version = File.read(File.expand_path("../../../version", __FILE__)).strip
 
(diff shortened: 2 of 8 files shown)


scripts/generator/graphql_generator/csharp.rb
```diff
@@ -20,12 +20,17 @@ def erb_for(template_filename)
       end
     end
 
-    attr_reader :scalars, :schema, :namespace, :version
+    attr_reader :scalars, :schema, :namespace, :version, :publish_destination
+
+    def initialize(schema, publish_destination, namespace:, custom_scalars: [], script_name: $0)
+      if publish_destination != "asset-store" && publish_destination != "github"
+        raise "publish_destination is invalid valid values are: asset-store or github"
```
richgilbank: Probably want this to be a little more readable. 
> 'publish_destination' is invalid. Valid values are 'asset-store' or 'github'

scripts/generator/graphql_generator/csharp/Editor/ShopifyOnboardingPanel.cs.erb
```diff
@@ -73,9 +73,10 @@ namespace <%= namespace %>.SDK {
                 Styles.BodyStyle
             );
 
-            if (GUILayout.Button("Start your free trial", Styles.ButtonStyle))
-            Application.OpenURL("https://www.shopify.ca/signup");
-
+            if (GUILayout.Button("Start your free trial", Styles.ButtonStyle)) {
+                Application.OpenURL("https://www.shopify.ca/signup?signup_types=unity_buy_sdk,<%= publish_destination %>");
```
richgilbank: `.ca` or `.com`? They probably both work, but maybe signup country domains are used somewhere in data land, so `.com` might be safer (since it redirects to country codes when applicable). That said, maybe brochure's nginx has redirects from `.ca` too.

richgilbank:
Can't really speak to the Unity/C# side of things, but the signup URL does indeed seem to be keeping those `signup_types` from the query string: 
image

scripts/generate.sh
```diff
@@ -1,7 +1,7 @@
 . $(dirname $0)/common.sh
 
 bundle --gemfile=$SCRIPTS_ROOT/generator/Gemfile
-$SCRIPTS_ROOT/generator/update_schema
+$SCRIPTS_ROOT/generator/update_schema $1
```
sleroux: Wrap with quotes to prevent globbing

mikkoh:
@sleroux didn't know about ShellCh. Should we add it to run whenever any of the build scripts are run?

scripts/generate.sh
mikkoh: I'd like to do this change in another PR and fix all sh linting issues at the same time.

PR #256, opened 2017-08-04
[Android] Adding additional unit tests and instrument tests

sleroux:
* Adds additional unit tests for parts of the code that don't depend on Android such as UnityMessages and some of our models
* Adds an `androidTest` folder with instrumentation tests for all of our Android-dependent code including the coverage for `AndroidPayCheckoutSession` and `UnityAndroidPayFragment`


android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPaySessionCallbacks.java
```diff
@@ -0,0 +1,12 @@
+package com.shopify.unity.buy;
+
+import com.shopify.unity.buy.models.MailingAddressInput;
+
+interface AndroidPaySessionCallbacks {
```
sav007: usually it singular `AndroidPaySessionCallback`

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/UnityAndroidPayFragment.java
```diff
@@ -197,4 +198,8 @@ public void onWalletRequestCancel(int requestCode) {
     public void setSessionCallbacks(AndroidPaySessionCallbacks callbacks) {
         sessionCallbacks = callbacks;
     }
+
+    public void setGoogleClient(GoogleApiClient client) {
```
sav007: protected scope? `void setGoogleClient(GoogleApiClient client)` as it only used for testing

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/AndroidPaySessionCallbacks.java
sleroux: I actually meant to add this file in the previous PR and opened this one to add this file in https://github.com/Shopify/unity-buy-sdk/pull/264. I'll make the change there and land that before this goes in.

PR #257, opened 2017-08-08
[Fix] iOS tests that use shipping method identifier

dengjeffrey:
+ Before we were using calculated shipping rates, so the identifier was not static. This PR changes the expected shipping methods to match the fixed rates that are now set on the example store.

### DIFF
--- a/Assets/Plugins/iOS/Shopify/BuyTests/ApplePayFlowTests.swift
+++ b/Assets/Plugins/iOS/Shopify/BuyTests/ApplePayFlowTests.swift
@@ -35,9 +35,8 @@ class ApplePayFlowTests: XCTestCase {
     private let timeout = 10.0
     
     // These are the shipping methods expected from unity-buy-sdk.myshopify.com
-    private let expeditedShippingMethod    = PKShippingMethod(label: "Expedited Parcel", amount: 8.44,  identifier: "canada_post-DOM.EP-8.44")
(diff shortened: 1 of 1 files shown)


Assets/Plugins/iOS/Shopify/BuyTests/ApplePayFlowTests.swift
```diff
@@ -74,9 +73,9 @@ class ApplePayFlowTests: XCTestCase {
         
         let expectation      = self.expectation(description: "MockAuthorizationController.invokeDidSelectShippingContact failed to complete")
         let expectedSubtotal = PKPaymentSummaryItem(label: "SUBTOTAL", amount: 25.47)
-        let expectedShipping = PKPaymentSummaryItem(label: "SHIPPING", amount: 8.44)
+        let expectedShipping = PKPaymentSummaryItem(label: "SHIPPING", amount: standardShippingMethod.amount)
         let expectedTaxes    = PKPaymentSummaryItem(label: "TAXES",    amount: 2.93)
-        let expectedTotal    = PKPaymentSummaryItem(label: "TOTAL", amount: 33.91)
+        let expectedTotal    = PKPaymentSummaryItem(label: "TOTAL", amount: 31.30)
```
richardmonette: Should this not be something like `expeditedShippingMethod.amount`? (same as above)

dengjeffrey: it's `standardShippingMethod.amount` here, because we expect it to select the first shipping method. We used `expeditedShippingMethod.amount` above because we selected `expeditedShippingMethod`

PR #258, opened 2017-08-08
[Fix] android web checkout docs

dengjeffrey:
+ Documents Android Web Checkout feature in readme as Chrome Custom Tab
+ Fixes some doc linkings to be local
+ Documents Android build requirements

### DIFF
--- a/BUILDTARGETS.md
+++ b/BUILDTARGETS.md
@@ -5,48 +5,16 @@ The Unity Buy SDK currently supports the following build targets:
 * [Android](#Android)
 * [iOS](#ios)
 
-# Android
+You can find more esoteric details regarding platform specific dependencies and the build phase in [Build Details](BUILDDETAILS.md)
 
-### Requirements
+# Android
 
 The Unity Buy SDK requires Android games to have a minimum API level of Android 4.4 'Kit Kat' (API Level 19)
 
-### Building
-
-The library comes in a form of a `.AAR` file. In doing so we have provided a [mainTemplate.gradle](Assets/Plugins/Android/mainTemplate.gradle) that will be used by Unity when creating a Gradle project.
-
-The Gradle file includes the dependencies used by the libarary. If you would like to use your own Gradle file, add the following dependencies:
-
-```
-dependencies {
-    ...
-
-    compile(name: 'shopify_buy_plugin', ext:'aar')
-
-    compile 'com.android.support:appcompat-v7:26.0.0'
-    compile 'com.android.support:customtabs:26.0.0'
-    compile 'com.shopify.mobilebuysdk:buy3-pay-support:1.0.0'
-    compile 'com.squareup.moshi:moshi:1.5.0'
-}
-```
+In addition Unity 5.6 and above must be used to build on the Android platform. As such the `Build System` option in the `Build Settings` pane must be set to `Gradle`.
 
 
 # iOS
 
-### Requirements
-
 The Unity Buy SDK requires iOS games to have a minimum target SDK of iOS 9 and support Swift 3.
-
-### Building
-
-When running an iOS Player build, the SDK runs a custom post processor that will make changes to the generated Xcode project to support the SDK's requirements. These changes include:
-
-* Updating the Xcode project to support a minimum target of iOS 9.
-* Changing the default Swift supported version from 2 to 3.
-* Linking in a bridging header for the generated Unity target.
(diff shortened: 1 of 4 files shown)


BUILDTARGETS.md
```diff
@@ -2,8 +2,35 @@
...
+
+### Requirements
+
+The Unity Buy SDK requires Android games to have a minimum API level of Android 4.4 'Kit Kat' (API Level 19)
+
+### Building
+
+The library comes in a form of a `.AAR` file. In doing so we have provided a [mainTemplate.gradle](Assets/Plugins/Android/mainTemplate.gradle) that will be used by Unity when creating a Gradle project.
+
+The Gradle file includes the dependencies used by the libarary. If you would like to use your own Gradle file, add the following dependencies:
```
sleroux: Do we need to mention that gradle building is only available in Unity 5.6 and up or do we have a way to mitigate that?

BUILDTARGETS.md
```diff
@@ -2,8 +2,35 @@
...
 
+# Android
+
+### Requirements
+
+The Unity Buy SDK requires Android games to have a minimum API level of Android 4.4 'Kit Kat' (API Level 19)
+
+### Building
+
+The library comes in a form of a `.AAR` file. In doing so we have provided a [mainTemplate.gradle](Assets/Plugins/Android/mainTemplate.gradle) that will be used by Unity when creating a Gradle project.
```
mikkoh: Let's try to remember our target demographic here. These are Unity devs and not mobile devs. Up until this point I've read the requirements and that's all cool. But immediately I get told about `.AAR` files and `.gradle` files things I might not know anything about.

Personally I thin it'd be handy to have a step by step implementation guide where I don't need to learn about this stuff unless I want to learn more. Same sort of goes for iOS 😬

README.md
```diff
@@ -17,6 +17,7 @@ The Unity Buy SDK allows Unity developers to query and sell products from Shopif
 - Create a cart
 - Check out via weblink
 - Checkouts via Apple Pay and Safari Web View
+- Checkouts via Chrome Custom Tab
```
mikkoh: I think we need to say `Android Chrome Custom Tab` or if it makes more sense `Android Custom Chrome Tabs`.

I wonder if somewhere we should images of what the checkouts look like for these. Because`Custom Chrome Tabs` is a bit ambiguous.

EXAMPLES.md
```diff
@@ -22,11 +22,11 @@ Before you can start using the Unity Buy SDK, you need:
 
 - a Shopify store with at least one product
 - [a storefront access token for your app](https://help.shopify.com/api/storefront-api/getting-started#obtaining-a-storefront-access-token)
-- [to install the Unity Buy SDK into your Unity project](https://github.com/shopify/unity-buy-sdk#using-the-unity-buy-sdk-in-unity)
+- [to install the Unity Buy SDK into your Unity project](README.md#using-the-unity-buy-sdk-in-unity)
 
-## Using the SDK on iOS
+## Supported build targets
```
mikkoh: **Supported build targets** and **Build Target Requirements** makes me think I'll see information like:

1. What OS does the Unity SDK support publishing to
2. What devices can the Unity SDK hit
etc.

But that link also has implementation details.

BUILDTARGETS.md
dengjeffrey: The only way to mitigate this would be to include dependencies into the .AAR which might not play nice with other Android plugins who use the same dependencies. I think for now it is easier to support 5.6 and up

README.md
dengjeffrey: `Chrome Custom Tab` is just the name of the library we use as such as `Safari Web View`. We may want to provide images, but I think it could shown better in an example

dengjeffrey:
I made some updates that hopefully simplify things, please 👀  again

BUILDTARGETS.md
```diff
@@ -2,24 +2,19 @@
...
 
+* [Android](#Android)
 * [iOS](#ios)
 
-# iOS
-
-### Requirements
-
-The Unity Buy SDK requires iOS games to have a minimum target SDK of iOS 9 and support Swift 3.
+You can find more esoteric details regarding platform specific dependencies and the build phase in [Build Details](BUILDDETAILS.md)
```
mikkoh: Can we switch `esoteric details` to `more detailed explanation`. I think this would be more inline with Shopify copy.

PR #260, opened 2017-08-09
[Feature] web intent test

dengjeffrey:
+ Adds roboelectric test dependency
+ Tests that the listener is correctly invoked after a pause and resume

### DIFF
--- a/android/shopify_buy_plugin/build.gradle
+++ b/android/shopify_buy_plugin/build.gradle
@@ -40,7 +40,7 @@ dependencies {
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:1.10.19'
     testCompile 'org.json:json:20140107'
-    testCompile 'org.robolectric:robolectric:3.4'
+    testCompile 'org.robolectric:robolectric:3.4.2'
 }
 
 task copyAARToPlugins {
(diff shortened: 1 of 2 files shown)


android/shopify_buy_plugin/build.gradle
```diff
@@ -40,6 +40,7 @@ dependencies {
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:1.10.19'
     testCompile 'org.json:json:20140107'
+    testCompile 'org.robolectric:robolectric:3.4.2'
```
sleroux: IIRC the purpose of using Robolectric is so you can run unit tests on the JVM while still using Android objects as it provides shims for the Android calls. I've actually gone the other way and made anything referring to Android APIs into an instrument test [1]. The downside is that the tests have to be run from the emulator but when we add in the Unity integration tests we'll probably need to do this anyways. Do you think it makes sense to use an instrument test for this as well?

[1] https://github.com/Shopify/unity-buy-sdk/pull/256

sav007: Have you considered to mock android resources?
https://developer.android.com/training/testing/unit-testing/local-unit-tests.html

sav007: Oh, never mind I see what you trying to do with
org.robolectric.android.controller.FragmentController

dengjeffrey:
Will be moving to an integration test

PR #262, opened 2017-08-09
[Fix] Native checkout supported payment network

dengjeffrey:
+ Makes `canShowPaymentSetup` asynchronous because we need to fetch the shop's payment settings


scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -207,11 +207,17 @@ namespace <%= namespace %> {
         /// Determine whether the user can be shown to setup their native payment solution
         /// </summary>
         /// <returns>True if the user can be shown a native payment setup</returns>
```
sleroux: Need to update the docstring to mention the new call back parameter and void return.

dengjeffrey:
I believe we just bump the versioning to suggest that there is a breaking change

PR #267, opened 2017-08-10
[Android Pay] Adds custom override on UnityPlayerActivity in order to forward onActivityResult to fragments

sleroux:
Adds a custom Activity override on UnityPlayerActivity to work around a bug with Android Pay where onActivityResult is not fired to child fragments when the wallet chooser is dismissed.

### DIFF
--- a/Assets/Plugins/Android/AndroidManifest.xml
+++ b/Assets/Plugins/Android/AndroidManifest.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.shopify.unity.buy">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="${applicationId}">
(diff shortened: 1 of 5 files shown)


Assets/Plugins/Android/AndroidManifest.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.shopify.unity.buy">
+    <application android:label="@string/app_name" android:supportsRtl="true">
+        <meta-data
+            android:name="com.google.android.gms.wallet.api.enabled"
+            android:value="true" />
+
+        <activity android:name="com.shopify.unity.buy.ShopifyUnityPlayerActivity"
```
sleroux: The custom launch activity requires you to add it to the AndroidManifest.xml for the plugin so I also moved the wallet API requirement to here as well.

PR #268, opened 2017-08-10
[Fix] Removes usage of handler/looper from MessageCenter

sleroux:
Calling the message sender caushes a crash on the Android side and after giving it some thought realized that we don't actually need it at all. We can simply call the Unity send message commands. So I did just that.

### DIFF
--- a/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/MessageCenter.java
+++ b/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/MessageCenter.java
@@ -1,54 +1,33 @@
 package com.shopify.unity.buy;
 
-import android.os.Handler;
-import android.os.HandlerThread;
-
 import com.unity3d.player.UnityPlayer;
 
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


sleroux:
Probably to be consistent with conventions.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/MessageCenter.java
```diff
@@ -1,54 +1,33 @@
...
-            public void run() {
-                UnityMessage msg = UnityMessage.fromUnity(result);
-                MessageCallbacks callbacks = callbacksInWaiting.remove(msg.identifier);
-                if (callbacks != null) {
-                    callbacks.onResponse(msg.content);
-                }
-            }
-        });
+        UnityMessage msg = UnityMessage.fromUnity(result);
+        MessageCallbacks callbacks = callbacksInWaiting.remove(msg.identifier);
```
sleroux: Oh heh yes I'll make the change 👍

PR #271, opened 2017-08-10
[Refactor] Splits NativeMessage into Android/iOS parts

sleroux:
Splits up NativeMessage into platform-dependent partial classes so we can use the same logic for Android message handling.

### DIFF
--- a/scripts/generator/graphql_generator/csharp/SDK/Android/NativeMessage.Android.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/Android/NativeMessage.Android.cs.erb
@@ -9,7 +9,7 @@ namespace <%= namespace %>.SDK {
     #endif
 
     public partial class NativeMessage {
-        private AndroidJavaObject MessageCenter;
+        private static AndroidJavaObject MessageCenter;
 
         private static void _RespondToNativeMessage(string identifier, string message) {
             #if !SHOPIFY_MONO_UNIT_TEST
--- a/scripts/generator/graphql_generator/csharp/SDK/Android/NativeMessage.Android.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/Android/NativeMessage.Android.cs.erb
@@ -9,7 +9,9 @@ namespace <%= namespace %>.SDK {
     #endif
 
     public partial class NativeMessage {
+        #if !SHOPIFY_MONO_UNIT_TEST
         private static AndroidJavaObject MessageCenter;
+        #endif
 
         private static void _RespondToNativeMessage(string identifier, string message) {
             #if !SHOPIFY_MONO_UNIT_TEST
--- a/scripts/generator/graphql_generator/csharp/SDK/Android/NativeMessage.Android.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/Android/NativeMessage.Android.cs.erb
@@ -2,7 +2,6 @@
 namespace <%= namespace %>.SDK {
     using System;
     using System.Collections.Generic;
-    using System.Runtime.InteropServices;
     using <%= namespace %>.MiniJSON;
 
     #if !SHOPIFY_MONO_UNIT_TEST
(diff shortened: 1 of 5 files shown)


scripts/generator/graphql_generator/csharp/SDK/Android/NativeMessage.Android.cs.erb
```diff
@@ -0,0 +1,22 @@
+#if UNITY_ANDROID
+namespace <%= namespace %>.SDK {
+    using System;
+    using System.Collections.Generic;
+    using System.Runtime.InteropServices;
```
dengjeffrey: I don't think you need this

scripts/generator/graphql_generator/csharp/SDK/Android/NativeMessage.Android.cs.erb
```diff
@@ -0,0 +1,22 @@
...
+    using <%= namespace %>.MiniJSON;
+
+    #if !SHOPIFY_MONO_UNIT_TEST
+    using UnityEngine;
+    #endif
+
+    public partial class NativeMessage {
+        private static void _RespondToNativeMessage(string identifier, string message) {
+            #if !SHOPIFY_MONO_UNIT_TEST
+            using (var messageCenter = new AndroidJavaObject("com.shopify.unity.buy.MessageCenter")) {
```
dengjeffrey: Since we are using `respondToNativeMessage` often, should this be stored? instantiating `AndroidJavaObject` is somewhat expensive

scripts/generator/graphql_generator/csharp/SDK/iOS/NativeMessage.iOS.cs.erb
```diff
@@ -0,0 +1,13 @@
+#if UNITY_IOS
+namespace <%= namespace %>.SDK {
+    using System;
+    using System.Collections.Generic;
+    using System.Runtime.InteropServices;
+    using <%= namespace %>.MiniJSON;
```
dengjeffrey: I don't think this is needed anymore

scripts/generator/graphql_generator/csharp/SDK/Android/NativeMessage.Android.cs.erb
```diff
@@ -0,0 +1,24 @@
...
+    using System;
+    using System.Collections.Generic;
+    using <%= namespace %>.MiniJSON;
+
+    #if !SHOPIFY_MONO_UNIT_TEST
+    using UnityEngine;
+    #endif
+
+    public partial class NativeMessage {
+        private AndroidJavaObject MessageCenter;
```
dengjeffrey: Should this be static? Otherwise we would create a new MessageCenter reference for each message

scripts/generator/graphql_generator/csharp/SDK/Android/NativeMessage.Android.cs.erb
```diff
@@ -0,0 +1,24 @@
...
+    using System;
+    using System.Collections.Generic;
+    using <%= namespace %>.MiniJSON;
+
+    #if !SHOPIFY_MONO_UNIT_TEST
+    using UnityEngine;
+    #endif
+
+    public partial class NativeMessage {
+        private static AndroidJavaObject MessageCenter;
```
dengjeffrey: I think this will 💥 on Mono tests since `UnityEngine` is wrapped in `!SHOPIFY_MONO_UNIT_TEST`, We should probably define `UNITY_ANDROID`as a symbol in `test.sh`

sleroux:
@dengjeffrey I've fixed the preprocessor issue but I'm going to defer on adding the UNITY_ANDROID flag to `mcs` when we run the tests. When we define both UNITY_ANDROID and UNITY_IOS the naming of the native messages collide since they are same on each platform. We probably don't want to be defining both anyways. I've filed away a trello card to look into this for when I come back to the Android stuff.

scripts/generator/graphql_generator/csharp/SDK/iOS/NativeMessage.iOS.cs.erb
```diff
@@ -0,0 +1,12 @@
+#if UNITY_IOS
+namespace <%= namespace %>.SDK {
+    using System;
+    using System.Collections.Generic;
```
dengjeffrey: Nit: I don't think we need these imports anymore

scripts/generator/graphql_generator/csharp/SDK/Android/NativeMessage.Android.cs.erb
```diff
@@ -0,0 +1,26 @@
+#if UNITY_ANDROID
+namespace <%= namespace %>.SDK {
+    using System;
+    using System.Collections.Generic;
```
dengjeffrey: Nit: Do we need these ones?

PR #272, opened 2017-08-10
[Android Pay] Adds interface for bridging calls between Android and Unity 

sleroux:
Mirrors the same technique we used for passing messages from iOS to Unity but for Android. Stubbed out the receiver calls in the partial extension off of AndroidNativeCheckout.

### DIFF
--- a/scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.AndroidPayEventReceiver.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.AndroidPayEventReceiver.cs.erb
@@ -1,0 +1,26 @@
+#if UNITY_ANDROID
+namespace <%= namespace %>.SDK.Android {
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Runtime.InteropServices;
+    using <%= namespace %>.SDK;
+    using <%= namespace %>.MiniJSON;
+
+    using UnityEngine;
+
+    public partial class AndroidNativeCheckout : IAndroidPayEventReceiver {
+        public void OnUpdateShippingAddress(string serializedMessage) {
+        }
+
+        public void OnUpdateShippingLine(string serializedMessage) {
+        }
+
+        public void OnError(string serializedMessage) {
+        }
+
+        public void OnCancel(string serializedMessage) {
+        }
+    }
+}
+#endif
--- a/scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.AndroidPayEventReceiver.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.AndroidPayEventReceiver.cs.erb
@@ -3,7 +3,6 @@ namespace <%= namespace %>.SDK.Android {
     using System;
     using System.Collections;
     using System.Collections.Generic;
-    using System.Runtime.InteropServices;
     using <%= namespace %>.SDK;
     using <%= namespace %>.MiniJSON;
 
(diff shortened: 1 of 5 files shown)


scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.AndroidPayEventReceiver.cs.erb
```diff
@@ -0,0 +1,26 @@
+#if UNITY_ANDROID
+namespace <%= namespace %>.SDK.Android {
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Runtime.InteropServices;
```
dengjeffrey: Don't need this

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
```diff
@@ -23,7 +24,7 @@ namespace <%= namespace %>.SDK.Android {
             CartState = cartState;
         }
 
-        public bool CanShowPaymentSetup() {
+        public bool CanShowPaymentSetup(PaymentSettings paymentSettings) {
             // TODO
             return true;
```
dengjeffrey: Does Android have this capability, if not you can probably return false.

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
```diff
@@ -60,16 +61,15 @@ namespace <%= namespace %>.SDK.Android {
...
-                // TODO: Replace with mono object for receiving callbacks with.
-                object[] args = { "testing", merchantName, key, pricingLineItemsString, currencyCodeString, countryCodeString, requiresShipping };
-                if (androidPayCheckoutSession.Call<bool>("checkoutWithAndroidPay", args)) {
-                    // TODO: Do I need to do anything on the Unity side after we've created the session
-                } else {
-                    // TODO: Create more meaningful error?
-                    var error = new ShopifyError(ShopifyError.ErrorType.NativePaymentProcessingError, "Unable to create Android Pay session.");
-                    OnFailure(error);
-                }
+            AndroidPayCheckoutSession = new AndroidJavaObject("com.shopify.unity.buy.AndroidPayCheckoutSession");
```
dengjeffrey: If `AndroidPayCheckoutSession` exists you don't want to create a new one every time. Any reason to why we may want to keep this resource alive, instead of using `using`?

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
sleroux: Is this selecting your payment method? If so that's the wallet chooser that we show when we request the masked wallet.

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
sleroux: Originally when I wrote this patch I figured I would need it to send messages back but I have the MessageCenter to do that now so this can go back to be `using`.

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
dengjeffrey: nope this is for opening a wallet app to set up Apple Pay/Android Pay. [`openPaymentSetup`](https://developer.apple.com/documentation/passkit/pkpasslibrary/1617100-openpaymentsetup)

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
```diff
@@ -61,15 +59,16 @@ namespace <%= namespace %>.SDK.Android {
...
-                AndroidPayEventBridge.Receiver = this;
-            } else {
-                // TODO: Create more meaningful error?
-                var error = new ShopifyError(ShopifyError.ErrorType.NativePaymentProcessingError, "Unable to create Android Pay session.");
-                OnFailure(error);
+            using (var androidPayCheckoutSession = new AndroidJavaObject("com.shopify.unity.buy.AndroidPayCheckoutSession")) {
+                object[] args = { GlobalGameObject.Name, merchantName, key, pricingLineItemsString, currencyCodeString, countryCodeString, requiresShipping, true };
+                if (androidPayCheckoutSession.Call<bool>("checkoutWithAndroidPay", args)) {
+                    AndroidPayEventBridge = GlobalGameObject.AddComponent<AndroidPayEventReceiverBridge>();
+                    AndroidPayEventBridge.Receiver = this;
```
dengjeffrey: Should we be checking if this exists before creating a new one?

PR #273, opened 2017-08-10
[Feature] Added payment settings check to can checkout

dengjeffrey:


scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.cs.erb
```diff
@@ -43,11 +43,16 @@ namespace <%= namespace %>.SDK.iOS {
...
         /// <param name="paymentSettings">The Shop's payment settings</param>
-        /// <returns>True if the device is capable of paying with Apple Pay</returns>
+        /// <returns>True if the device is capable of paying with Apple Pay and the store accepts payments with Apple Pay</returns>
         public bool CanCheckout(PaymentSettings paymentSettings) {
+
+            if (!paymentSettings.supportedDigitalWallets().Contains(DigitalWallet.APPLE_PAY)) {
+                return false;
+            }
+
             return _CanCheckoutWithApplePay(SerializedPaymentNetworksFromCardBrands(paymentSettings.acceptedCardBrands()));
```
richardmonette: Just a minor style thing, but if we are doing early returns, should be make both cases return false on !, and then the fall through is return true?

PR #277, opened 2017-08-11
[Fix] Added build_android phase to publish script

sleroux:


sleroux:
Closing this PR for now since it will cause CI to always fail since Circle doesn't support building Android targets on their mac instances :(

PR #279, opened 2017-08-14
update checkout query

richardmonette:
~~Looks like the storefront API moved around a few params, fixing tests.~~

cc @Shopify/gaming

### DIFF
--- a/Assets/Editor/TestDefaultCheckoutQueries.cs
+++ b/Assets/Editor/TestDefaultCheckoutQueries.cs
@@ -15,7 +15,7 @@ namespace Shopify.Tests
 
             DefaultQueries.checkout.Create(query, lineItems);
             Assert.AreEqual(
-                "mutation{checkoutCreate (input:{allowPartialAddresses:true,lineItems:[]}){checkout {id webUrl currencyCode requiresShipping subtotalPrice totalTax totalPrice ready lineItems (first:250){edges {node {id variant {id }}cursor }pageInfo {hasNextPage }}}userErrors {field message }}}",
+                "mutation{checkoutCreate (input:{lineItems:[],allowPartialAddresses:true}){checkout {id webUrl currencyCode requiresShipping subtotalPrice totalTax totalPrice ready lineItems (first:250){edges {node {id variant {id }}cursor }pageInfo {hasNextPage }}}userErrors {field message }}}",
                 query.ToString()
             );
         }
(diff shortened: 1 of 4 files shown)


Assets/Editor/TestDefaultCheckoutQueries.cs
```diff
@@ -138,11 +138,11 @@ public class TestDefaultCheckoutQueries {
             string checkoutId = "an-id";
 
             var billingAddress = new MailingAddressInput("123 Test Street", "456", "Toronto", "Shopify", "Canada", "First", "Last", "1234567890", "Ontario", "A1B2C3");
-            var tokenizedPaymentInput = new TokenizedPaymentInput(new decimal(1), billingAddress, "unique_id", "some_utf8_data_string", "apple_pay");
+            var tokenizedPaymentInput = new TokenizedPaymentInput(new decimal(1), "unique_id", billingAddress, "some_utf8_data_string", "apple_pay");
```
dengjeffrey: "apple_pay" needs to go before "some_utf8_data_string"

richardmonette:
Updated description: So it turns out the issue was that I was on the latest version of https://github.com/Shopify/graphql_schema, a gem we are using that is involved in processing the graphql_schema.json. Specifically, between the old version (0.1.3) and the new version (0.1.6) there was a change where the sorting of the fields was removed https://github.com/Shopify/graphql_schema/pull/10 This PR updates our tests to work with the latest gem version.

richardmonette:
@sleroux Yes, mostly its sublime automatically cleaning up trailing spaces on line endings.

dengjeffrey:
I think we should be bumping the `Gemfile` dependency as well to graphql_schema (~> 0.1.4)` at least since `0.1.4` has breaking changes compared to `0.1.3` which had sorting. Currently a dependency of `0.1.0` seems kind of a lie.

dengjeffrey:
Or better yet, if it does work on `0.1.0` we should exclude `0.1.3` since that was the version with sorting

richardmonette:
@dengjeffrey How about, `gem "graphql_schema", "~>0.1", ">=0.1.4"`, which would mean, any version of 0.1.X, but not 0.2.X, as long as 0.1.X has X larger than .4

dengjeffrey:
I don't see a problem with including 0.2.X, I think our only issue is `0.1.3`, perhaps: 
```
gem "graphql_schema", "~>0.1.0", "!=0.1.3"
```
That includes everything we had before but excludes the version that caused the cases to break

richardmonette:
@dengjeffrey Updated!

Issue #269, opened 2017-08-10
Desktop checkout should differentiate between cancelled and success


richardmonette:
Right now, when doing a desktop checkout, we do something like

```
_cart.GetWebCheckoutLink(
  success: (string url) =>
  {
     Application.OpenURL(url);
     OnCheckoutSuccess.Invoke();
  },
  failure: (checkoutError) => {
     OnCheckoutFailure.Invoke(checkoutError.Description);
  }
 );
```

However, the success is just for getting the link, what we'd rather have is knowledge of whether the checkout was successful, cancelled or failed. I believe we can determine this the same way we do for mobile, by checking to see if the checkout url was completed successfully. 

The problem I see is that the SDK won't necessarily know when the Application.OpenURL browser closes, so we may need the function which checks the checkout url to be exposed, so the app can trigger the check itself, when it regains focus.

@Shopify/gaming

richardmonette:
Fixed by https://github.com/Shopify/unity-buy-sdk/pull/280

PR #280, opened 2017-08-15
Add UnityWebCheckout

richardmonette:
When running in the `UNITY_EDITOR` or `UNITY_STANDALONE` setup the `WebCheckout` with the new class `DesktopCheckout`. This new checkout uses `Application.OpenURL(checkoutURL)` to open a browser window via the Unity player, typically this will be Chrome, etc. in a desktop situation. 

In order to trigger a poll on checkout status (in order to call the correct callback) when the player returns to the game from the external browser, `DesktopMessageReceiver` uses `OnApplicationFocus`.

Fixes https://github.com/Shopify/unity-buy-sdk/issues/269

cc @Shopify/gaming


scripts/generator/graphql_generator/csharp/SDK/DesktopMessageReceiver.cs.erb
```diff
@@ -0,0 +1,72 @@
...
+        private bool isPaused = false;
+
+        public void Init(
+            ShopifyClient client,
+            Checkout currentCheckout,
+            CheckoutSuccessCallback onSuccess,
+            CheckoutCancelCallback onCancelled,
+            CheckoutFailureCallback onFailure)
+        {
+            Debug.Log("DesktopMessageReceiver Init");
```
dengjeffrey: Nit: These should be removed? 😄

scripts/generator/graphql_generator/csharp/SDK/DesktopMessageReceiver.cs.erb
```diff
@@ -0,0 +1,72 @@
...
+                            OnSuccess();
+                        } else {
+                            OnCancelled();
+                        }
+                    }
+                });
+            }
+        }
+
+        void OnApplicationPause(bool pauseStatus)
```
dengjeffrey: Do we need this method?

scripts/generator/graphql_generator/csharp/SDK/DesktopMessageReceiver.cs.erb
```diff
@@ -0,0 +1,72 @@
...
+
+            Debug.Log("DesktopMessageReceiver OnApplicationFocus");
+
+            if (!isPaused) {
+                var query = new QueryRootQuery();
+                query.node(
+                    buildQuery: n => n
+                    .onCheckout(c => c.completedAt()),
+                    id: CurrentCheckout.id()
+                );
```
dengjeffrey: Would it be worthwhile to make this one of the `DefaultCheckoutQueries` since we also have it in `NativeWebViewMessageReceiver`

scripts/generator/graphql_generator/csharp/SDK/DesktopMessageReceiver.cs.erb
```diff
@@ -0,0 +1,72 @@
+namespace <%= namespace %>.SDK {
+    using System;
+    using System.Runtime.InteropServices;
```
dengjeffrey: Not needed

scripts/generator/graphql_generator/csharp/SDK/DesktopCheckout.cs.erb
```diff
@@ -0,0 +1,44 @@
+#if UNITY_IOS
+namespace <%= namespace %>.SDK.iOS {
+    using System;
+    using System.Runtime.InteropServices;
```
dengjeffrey: Not needed

scripts/generator/graphql_generator/csharp/SDK/DesktopCheckout.cs.erb
```diff
@@ -0,0 +1,44 @@
...
+            _cart = cart;
+            _client = client;
+        }
+
+        public override void Checkout(string checkoutURL, CheckoutSuccessCallback success, CheckoutCancelCallback cancelled, CheckoutFailureCallback failure) {
+            #if !SHOPIFY_MONO_UNIT_TEST
+            if (_desktopMessageReceiver == null) {
+                _desktopMessageReceiver = GlobalGameObject.AddComponent<DesktopMessageReceiver>();
+            }
+            _desktopMessageReceiver.Init(_client, _cart.CurrentCheckout, success, cancelled, failure);
```
dengjeffrey: I feel a bit uneasy about having `DesktopCheckout` not use `SetupNativeMessageReceiver`. Creating a `_desktopMessageReceiver` should ideally be set up there to make things consistent and clear.

scripts/generator/graphql_generator/csharp/SDK/DesktopMessageReceiver.cs.erb
richardmonette: I believe it's it's relevant when on Android, the use case here is for desktop, but since the Unity documentation seems to recommend its use I included it https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnApplicationFocus.html

richardmonette:
Made a few changes:

- Introduce `SetupDesktopMessageReceiver` which sets up the `DesktopMessageReceiver`. Wondering now if there is maybe a base class between the `NativeWebViewMessageReceiver` and this new class?
- In order to be able to request an external browser on all platforms, introduce `CheckoutWithExternalBrowser` as a function on the cart.
- Added the `completedAt` field to `Poll`, so it can be used in the polling queries https://github.com/Shopify/unity-buy-sdk/pull/280/files#diff-b84040fe4fe9a206d138e8636754e02bR40

scripts/generator/graphql_generator/csharp/SDK/DesktopMessageReceiver.cs.erb
```diff
@@ -0,0 +1,56 @@
...
+    using System;
+    using Shopify.Unity.SDK;
+    using Shopify.Unity.GraphQL;
+
+    #if !SHOPIFY_MONO_UNIT_TEST
+    using UnityEngine;
+    public partial class DesktopMessageReceiver: MonoBehaviour {}
+    #endif
+
+    public partial class DesktopMessageReceiver {
```
sleroux: > Introduce SetupDesktopMessageReceiver which sets up the DesktopMessageReceiver. Wondering now if there is maybe a base class between the NativeWebViewMessageReceiver and this new class?

Yea I can see that too. They pretty much do the same thing except the name of the method that invoked to do the validation check. I can see moving this logic to an abstract base, wrapping the validation in a method and calling it from the subclasses

dengjeffrey: I agree on this too. It might be worthwhile to explore a delegate pattern where the `Receiver` is just basically a listener for when the Web browser is closed. Ideally the handling of verification and callbacks should be handled elsewhere not in the receiver class.

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -122,6 +122,8 @@ namespace <%= namespace %> {
             State = new CartState(client);
             Client = client;
 
+            ExternalBrowserCheckout = new DesktopCheckout(this, client);
```
richardmonette: I think it makes sense to have this always available, as Application.OpenUrl can be attempted by the Unity player on every platform (I believe), however the name should definitely be changed from Desktop, that was just my first guess at the right name, it should probably reference that it will be an ExternalBrowserCheckout

scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -159,6 +161,21 @@ namespace <%= namespace %> {
         }
 
         /// <summary>
+        /// Launches an external browser with the Cart's web checkout link loaded. This can be used to perform
+        /// a cart checkout when no native web view is available, such as on desktop, or from the Unity editor.
+        /// </summary>
+        /// <param name="success">called when the game has regained focus and the checkout was successful.</param>
+        /// <param name="cancelled">called when the game has regained focus but the checkout has not completed.</param>
+        /// <param name="failure">called when the game has regained focused and an error has occured during checkout.</param>
+        public void CheckoutWithExternalBrowser(CheckoutSuccessCallback success, CheckoutCancelCallback cancelled, CheckoutFailureCallback failure) {
```
richardmonette: @dengjeffrey @sleroux Question for you guys, thinking in terms of https://github.com/Shopify/unity-buy-sdk/issues/291, it seems there are a few ways we would go with adding in the functionality around `Application.OpenURL`

Option 1. We make this method apply only to standalone and editor builds, which entails making the name something like, CheckoutWithDesktopBrowser, and we use ifdefs to only define that type of checkout for standalone.

Option 2. There isn't any technical reason (AFAIK) that we couldn't call `Application.OpenURL` in a mobile context, so we could keep this as a separate method, but allow it to be called on mobile. If that was the case, I think we drop the Desktop part of the name, and call it ExternalCheckout or similar, and do not use ifdefs to prevent it from being defined for mobile.

Option 3. We don't add a new method, but instead we add logic to `CheckoutWithNativeWebView` in order to use `Application.OpenUrl` when not on a mobile device. This would mean the name of the method no longer makes sense, so we'd want to change it to perhaps just `CheckoutWithWebView`. One detail here, on iOS we do launch a "true" web view (in that you can't just browse away from it) but on android we are basically just launching an external browser anyways (although it does feel a but more like a web view).

So, there are some further questions about if and how we handle the callbacks on the function we choose to implement, but I think that's our first decision to make.

dengjeffrey: As per the discussion with @mikkoh, I believe the best is `Option 2`. This keeps the intent of the methods clear. The amount of work to do `Option 1` seems the same as `Option 2` so might as well do `Option 2`

sleroux: I'm kind of leaning towards Option 3 since there isn't much a difference between an external browser and the browsers we show on mobile. Technically we don't show a 'web view' on either platform - we launch into a browser. I wonder if a developer would ever want to differentiate between the external browser and the browsers we launch on mobile. I can see it being an implementation detail on our side. To them I can see if being the "non-customizable web checkout". Maybe we rename this method to be `performWebCheckout` to wrap both approaches.

The only caveat is we would need to be able to reliably call the success/failure delegates in the same way across both approaches (which I think we can do?)

sleroux: Just to add one last thought: It seems that we either have the developer wrap the calls with their own platform #ifdefs or we do it internally behind one method. We should probably only be requiring them to the #ifdef check if we have a good reason why we can't.

richardmonette: So if we go options 3, then yeah, we'd definitely need to change `CheckoutWithNativeWebView` to something else, I was thinking`CheckoutWithWebView`.

We'd do something like to pick which WebCheckout implementation to use,

```
#if UNITY_IOS
WebCheckout = new iOSWebCheckout(this, client);
#elif UNITY_ANDROID
WebCheckout = new AndroidWebCheckout(this, client);
NativeCheckout = new AndroidNativeCheckout(State);
#elif STANDALONE OR DESKTOP
WebCheckout = new UnityWebCheckout(this, client);;
NativeCheckout = null;
#else
WebCheckout = null;
NativeCheckout = null;
#endif
```

That's actually pretty elegant. @dengjeffrey you cool with that too? cc @sleroux

richardmonette: > We should probably only be requiring them to the #ifdef check if we have a good reason why we can't.

That sort of leads us towards Option 3.

dengjeffrey: Good points, I forgot the reason why @mikkoh preferred opt 2, but now that we are aligning external web browsers and web views, it seems like opt 3 provides ease of use for the implementor

scripts/generator/graphql_generator/csharp/SDK/WebCheckoutMessageReceiver.cs.erb
```diff
@@ -32,25 +32,43 @@ namespace <%= namespace %>.SDK {
...
+                    if (checkout.completedAt() != null) {
+                        OnSuccess();
                     } else {
-                        var checkout = (Checkout) response.node();
-                        if (checkout.completedAt() != null) {
-                            OnSuccess();
-                        } else {
-                            OnCancelled();
-                        }
+                        OnCancelled();
```
dengjeffrey: Is the decision to keep `OnCancelled()` in?

richardmonette: At least for this PR, imho, yes. The goal of this PR is to add UnityWebCheckout, so we have a way to do web checkouts in the editor / stand alone in the SDK. I think our path forward is to solve that problem with this PR, then we can do another which, as a separate change, potentially modifies which callbacks we offer, and alters the lifecycle on how they are called.

dengjeffrey:
You need to remove `NativeWebViewMessageReceiver.cs.erb` from `csharp.rb` too.

PR #288, opened 2017-08-18
[iOS11] Added extension on PKError that adds constructors for JSON

dengjeffrey:
+ Adds an extension on `NSError` that creates `PKErrors` from JSONs for convenience
    + PKErrors are also an extension on NSError
+ Enums have CamelCaps strings as values to match the Enums in C#
+ CI tests do not pass because there are licensing issues with the latest Unity Betas

##### Motivation
+ If we receive `UserErrors` from making a Graphql request in Unity, we serialize the `UserErrors` into a JSON string that contains the necessary information to construct a `PKError`. The JSON string is passed to unmanaged code, our iOS plugin. This PR adds the functionality to deserialize the JSON string and construct a `PKError` using the suggested conveniences on `NSError`.
+ These `PKErrors` are used when responding to Apple Pay update requests that we have sent to our managed code, the Unity application


Assets/Plugins/iOS/Shopify/PKError+JSON/PKError+JSON.swift
```diff
@@ -0,0 +1,88 @@
...
+
+// ----------------------------------
+//  MARK: - Class Payment Error Creator -
+//
+@available(iOS 11.0, *)
+extension NSError {
+
+    /// Convenience method to create a payment error with a JSON
+    /// Returns a Payment error, payment errors can be found in PKError.h
+    class func paymentError(with json: JSON) -> Error? {
```
dbart01: Do you really want this to be a `Error?` protocol return type and not a `PKError`?

dbart01:
I'm not sure I understand the design here. There's a bunch of types and extensions that map one kind of error to another but they aren't used anywhere? Now, I imagine they *will* be used at some point, so what's the use case? Should you include the usage of these in this PR for clarity?

Assets/Plugins/iOS/Shopify/PKError+JSON/PKError+JSON.swift
dengjeffrey: I believe `PKError` isn't actually a type, however `PKPaymentError` is but in the [docs](https://developer.apple.com/documentation/passkit/pkpaymenterror) it suggests to use the extensions on `NSError` and those also return an `Error?` protocol

dengjeffrey:
@dbart01 Yes, good idea! I added the motivation behind this PR in description, let me know if anything else is unclear

dbart01:
So the final goal here is a `PKError` from json? That seems like a single method map between JSON  values and various `PKError` cases. Are the `...Meta` classes used outside of this mapping at all?

dengjeffrey:
No `Meta` classes are only used as a helper. Should I have wrapped them in a `NSError` extension? Other option was to put them all in the same file as `PKError+JSON` but this seems messy.

Assets/Plugins/iOS/Shopify/PKError+JSON/ContactInvalidErrorMeta.swift
```diff
@@ -0,0 +1,80 @@
...
+        case postalAddress = "PostalAddress"
+        case emailAddressa = "EmailAddress"
+        case phoneNumber   = "PhoneNumber"
+        case name          = "Name"
+        case phoneticName  = "PhoneticName"
+        
+        func contactField() -> PKContactField {
+            switch self {
+            case .postalAddress: return PKContactField.postalAddress
+            case .emailAddressa: return PKContactField.emailAddress
```
sleroux: `.emailAddressa` <- Typo?

Assets/Plugins/iOS/Shopify/PKError+JSON.swift
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    fileprivate enum ErrorType: String {
+        case paymentBillingAddress       = "PaymentBillingAddress"
+        case paymentShippingAddress      = "PaymentShippingAddress"
+        case paymentContactInvalid       = "PaymentContactInvalid"
+        case shippingAddressUnservicable = "ShippingAddressUnservicable"
+    }
+
+    fileprivate class func errorType(in json: JSON) -> ErrorType? {
```
dbart01: I would make this an initializer on `ErrorType` instead.

dengjeffrey:
I will be updating to reflect API changes with the new Xcode Beta 😞

Assets/Plugins/iOS/Shopify/PKError+JSON.swift
```diff
@@ -0,0 +1,141 @@
...
+    }
+}
+
+// ----------------------------------
+//  MARK: - Parsing Errors -
+//
+@available(iOS 11.0, *)
+extension PKPaymentRequest {
+
+    fileprivate enum PaymentErrorKey: String {
```
dbart01: This doesn't look like the values are ever used for their `enum` types but rather for their underlying value. I personally prefer to represent keys as static constants:
```swift
private enum Keys {
    static let type        = "type"
    static let description = "description"
    static let field       = "field"
}
```
You can also "un-nest" them from `PKPaymentRequest` and convert to `private` to simplify naming since they are just simple JSON keys.

Assets/Plugins/iOS/Shopify/String/CNPostalAddress+String.swift
```diff
@@ -0,0 +1,46 @@
...
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Contacts
+
+extension CNPostalAddress {
+
+    @available(iOS 10.3, *)
+    class func postalAddressKey(for string: String) -> String? {
```
dbart01: Can this be an initializer instead?

dengjeffrey: An initializer for `CNPostalAddress`? `CNPostalAddress` is a class with some postal fields.

Assets/Plugins/iOS/Shopify/PKError+JSON.swift
dengjeffrey: I'm going to keep it as an enum here to match the rest of the plugin's implementation

Assets/Plugins/iOS/Shopify/String/CNPostalAddress+String.swift
dbart01: Ohh my bad, didn't realize the return type is `String?`. This is fine then 👍

PR #292, opened 2017-08-21
[Feature] Input field constants

dengjeffrey:
+ Adds string constants that reflect the naming of each field
    + Does not follow C# naming conventions, capitalizing fields would add unneeded complexity
+ We do not use Enums, incase a field is named the same as a reserved word

##### Motivation
+ Since `UserErrors` directly map to an Input type's field, it would be safer to compare that the `UserError` is equivalent to variable rather than a string. This allows us to have compile time safety when/if the API for the input type changes

### DIFF
--- a/scripts/generator/graphql_generator/csharp/type.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/type.cs.erb
@@ -120,14 +120,19 @@ namespace <%= namespace %>.GraphQL {
     <% when 'INPUT_OBJECT' %>
         <%= docs_input_object(type) %>
         public class <%= type.classify_name %> : InputBase {
+
+            <% type.input_fields.each do |field| %>
+            public const string <%= field.name %>FieldKey = "<%= field.name %>";
+            <% end %>
+
             <% type.input_fields.each do |field| %>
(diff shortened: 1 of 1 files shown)


scripts/generator/graphql_generator/csharp/type.cs.erb
```diff
@@ -120,14 +120,19 @@ namespace <%= namespace %>.GraphQL {
     <% when 'INPUT_OBJECT' %>
         <%= docs_input_object(type) %>
         public class <%= type.classify_name %> : InputBase {
+
+            <% type.input_fields.each do |field| %>
+            public const string <%= field.name %>FieldKey = "<%= field.name %>";
```
richardmonette: This is public because some later PR is going to need to reference these (or could these be made private, seems like, at least in this PR alone, they are used only internally to the given class)?

dengjeffrey: Yep this is made public because we need to reference these when handling `UserErrors` down the road

richardmonette: 🆗

PR #294, opened 2017-08-22
[Feature] Apple pay event response error

dengjeffrey:
+ Adds functionality to create ApplePayError based on the different UserErrors received
+ In cases where we receive an error that we don't expect or an error that can not be created into an ApplePayError we return a Failure.

### DIFF
--- a/scripts/generator/graphql_generator/csharp/SDK/iOS/ApplePayEventResponse.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/iOS/ApplePayEventResponse.cs.erb
@@ -5,14 +5,21 @@ namespace <%= namespace %>.SDK.iOS {
     using System.Collections.Generic;
 
     public class ApplePayEventResponse: Serializable {
-        public string AuthorizationStatus;
-        public List<SummaryItem> SummaryItems;
-        public List<ShippingMethod> ShippingMethods;
+        public readonly string AuthorizationStatus;
+        public readonly List<SummaryItem> SummaryItems;
+        public readonly List<ShippingMethod> ShippingMethods;
+        public readonly List<ApplePayError> Errors;
 
-        public ApplePayEventResponse(ApplePayAuthorizationStatus authorizationStatus, List<SummaryItem> summaryItems = null, List<ShippingMethod> shippingMethods = null) {
+        public ApplePayEventResponse(
+            ApplePayAuthorizationStatus authorizationStatus,
+            List<SummaryItem> summaryItems = null,
+            List<ShippingMethod> shippingMethods = null,
+            List<ApplePayError> errors = null)
+        {
             AuthorizationStatus = authorizationStatus.ToString("G");
             SummaryItems = summaryItems ?? new List<SummaryItem>();
             ShippingMethods = shippingMethods ?? new List<ShippingMethod>();
+            Errors = errors ?? new List<ApplePayError>();
         }
 
         public override object ToJson() {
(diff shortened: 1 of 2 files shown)


dengjeffrey:
Yes it is failing, taking a 👀  now

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -41,6 +83,111 @@ namespace <%= namespace %>.SDK.iOS {
...
+
+            var payErrors = new List<ApplePayError>();
+
+            // Check to see if any of the user errors are not billing address errors
+            // If it is not a billing address error return a failure
+            foreach (var error in errors) {
+                var fields = error.field();
+                if (fields.Contains("billingAddress")) {
+                    string description = error.message();
+                    string lastField = fields[fields.Count - 1];
```
richardmonette: Maybe a helper method for getting the last field, which handles avoiding out of bounds exceptions.

dengjeffrey: I don't think it would be appropriate here to throw an exception, there is no way we can handle it either. The only option is to let it 💥 or handle it gracefully by saying it's a failure. The case where the number of fields is 0 is when the SDK is broken or the Storefront API is broken.

scripts/generator/graphql_generator/csharp/SDK/iOS/ApplePayEventResponse.cs.erb
```diff
@@ -5,14 +5,21 @@ namespace <%= namespace %>.SDK.iOS {
...
+        public ApplePayEventResponse(
+            ApplePayAuthorizationStatus authorizationStatus,
+            List<SummaryItem> summaryItems = null,
+            List<ShippingMethod> shippingMethods = null,
+            List<ApplePayError> errors = null)
+        {
             AuthorizationStatus = authorizationStatus.ToString("G");
             SummaryItems = summaryItems ?? new List<SummaryItem>();
             ShippingMethods = shippingMethods ?? new List<ShippingMethod>();
+            Errors = errors ?? new List<ApplePayError>();
```
dengjeffrey: Seems like this doesn't work, default param needs to be constant

dengjeffrey:
@sleroux @richardmonette I fixed the issues, please take another look 😄

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
richardmonette: So here https://github.com/Shopify/unity-buy-sdk/pull/294/files#diff-dd4b7dc40c95a7f0a9eb368e629d5e53R137 we have the access in the try, so if it explodes, it reports Failure, which I think makes sense, but then in this case its not under the try, so I think it could let an out of bounds reach the user. Should this not be under the try too?

dengjeffrey: We could put it into the try as well, but it'll fail on `fields.Contains("billingAddress")`

richardmonette: Sorry, yeah, I was thinking we'd put anything that needs fields to be 'ok' under the try

dengjeffrey: Sure I'll put that under the `try` to keep it consistent

PR #296, opened 2017-08-23
omit includes for iOSPostProcessor unless on ios

richardmonette:
Fixes https://github.com/Shopify/unity-buy-sdk-examples/issues/17

> UNITY_IOS  #define directive for compiling/executing code for the iOS platform.

https://docs.unity3d.com/Manual/PlatformDependentCompilation.html

It appears we were actually already using conditional compilation here, we just need to make a bit of an adjustment so that the using statements are omitted as well, unless doing an ios build.

@Shopify/gaming @pushmatrix


pushmatrix:
🎩

Issue #148, opened 2017-07-05
CartLineItems Subtotal


mikkoh:
At some point users may want the subtotal of the current cart. We may want to expose the subtotal to the user.

PR #297, opened 2017-08-24
add CartLineItems Subtotal

richardmonette:
Fixes https://github.com/Shopify/unity-buy-sdk/issues/148

Adds functionality for getting the cart subtotal. 

@Shopify/gaming

### DIFF
--- a/scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
@@ -18,6 +18,12 @@ namespace <%= namespace %>.SDK {
             }
         }
 
+        public decimal Price {
+            get {
+                return _Price;
+            }
+        }
+
         public long Quantity {
             get {
                 return _Quantity;
@@ -48,12 +54,14 @@ namespace <%= namespace %>.SDK {
         public string ID = null;
         private string _VariantId;
         private long _Quantity;
+        private decimal _Price;
         private ObservableDictionary<string, string> _CustomAttributes;
         private OnCartLineItemChange OnChange;
 
-        public CartLineItem(string variantId, OnCartLineItemChange onChange, long quantity = 1, IDictionary<string, string> customAttributes = null) {
-            _VariantId = variantId;
+        public CartLineItem(ProductVariant variant, OnCartLineItemChange onChange, long quantity = 1, IDictionary<string, string> customAttributes = null) {
+            _VariantId = variant.id();
             _Quantity = quantity;
+            _Price = variant.price();
             OnChange = onChange;
 
             if (customAttributes != null) {
@@ -126,6 +134,18 @@ namespace <%= namespace %>.SDK {
             }
         }
 
+        public decimal Subtotal {
+            get {
+                decimal subtotalOut = 0;
+
+                foreach (var lineitem in LineItems) {
+                    subtotalOut += lineitem.Price * lineitem.Quantity;
+                }
+
+                return subtotalOut;
+            }
+        }
+
         private bool _IsSaved = false;
         private List<CartLineItem> LineItems = new List<CartLineItem>();
         private Action<string> OnDeleteLineItem;
@@ -165,7 +185,7 @@ namespace <%= namespace %>.SDK {
 
                 LineItems.Add(
                     new CartLineItem(
-                        variantId: variant.id(),
+                        variant: variant,
                         onChange: OnLineItemChange,
                         quantity: (long) quantity,
                         customAttributes: customAttributes
--- a/Assets/Editor/MockLoaderCheckouts.cs
+++ b/Assets/Editor/MockLoaderCheckouts.cs
@@ -51,7 +51,8 @@ namespace Shopify.Tests {
                                             ""node"": {
                                                 ""id"": ""line-item-id1"",
                                                 ""variant"": {
-                                                    ""id"": ""Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0VmFyaWFudC8yMDc1NjEyOTE1NQ==""
+                                                    ""id"": ""Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0VmFyaWFudC8yMDc1NjEyOTE1NQ=="",
+                                                    ""price"": ""1.00""
                                                 }
                                             },
                                             ""cursor"": ""line-item-1-cursor""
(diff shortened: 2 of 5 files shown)


scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
```diff
@@ -147,7 +155,7 @@ namespace <%= namespace %>.SDK {
         ///
         /// cart.LineItems.AddOrUpdate(variantId, 3);
         /// \endcode
-        public void AddOrUpdate(string variantId, long? quantity = null, Dictionary<string, string> customAttributes = null) {
+        public void AddOrUpdate(string variantId, decimal price, long? quantity = null, Dictionary<string, string> customAttributes = null) {
             CartLineItem input = Get(variantId);
 
             if (input != null) {
```
dengjeffrey: Under this case when the cart line item already exists, should price be updated as well?

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
```diff
@@ -213,13 +222,24 @@ namespace <%= namespace %>.SDK {
         /// cart.LineItems.AddOrUpdate(product, selectedOptions, 3);
         /// \endcode
         public void AddOrUpdate(Product product, Dictionary<string, string> selectedOptions, long? quantity = null, Dictionary<string, string> customAttributes = null) {
-            string variantId = VariantIdFromSelectedOptions(product, selectedOptions);
+            var variantId = VariantIdFromSelectedOptions(product, selectedOptions);
+            var variant = GetVariantById(product, variantId);
```
dengjeffrey: Would it be possible to refactor this so we get `ProductVariant` from one go? Versus making `VariantIdFromSelectedOptions` and `GetVariantById` which is two for loops over `Products.variants`

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
```diff
@@ -48,12 +54,14 @@ namespace <%= namespace %>.SDK {
         public string ID = null;
         private string _VariantId;
         private long _Quantity;
+        private decimal _Price;
         private ObservableDictionary<string, string> _CustomAttributes;
         private OnCartLineItemChange OnChange;
-        
-        public CartLineItem(string variantId, OnCartLineItemChange onChange, long quantity = 1, IDictionary<string, string> customAttributes = null) {
+
+        public CartLineItem(string variantId, OnCartLineItemChange onChange, decimal price, long quantity = 1, IDictionary<string, string> customAttributes = null) {
```
dengjeffrey: This feels kind of shaky since we give power to the user to enter their own price, which is always a bad idea. What are the implications if instead `CartLineItem` takes in a `ProductVariant` instead of `variantID` and `price`. That way `CartLineItem` can get the variant ID, price, and any other variant info down the road

dengjeffrey: Same thing goes with removing `AddOrUpdate(string variantId, decimal price ...)` since we have `AddOrUpdate(ProductVariant variant, ...`

richardmonette: > This feels kind of shaky since we give power to the user to enter their own price, which is always a bad idea.

Definitely agree with you on this point.

So we've got three overloads of AddOrUpdate:

```public void AddOrUpdate(string variantId, long? quantity = null, Dictionary<string, string> customAttributes = null)

public void AddOrUpdate(ProductVariant variant, long? quantity = null, Dictionary<string, string> customAttributes = null)

public void AddOrUpdate(Product product, Dictionary<string, string> selectedOptions, long? quantity = null, Dictionary<string, string> customAttributes = null)
```

The later two are defined as basically calling through to the first version, where we actually use the `variantId`. I checked through our examples and, in all the cases I saw, we are recommending that developers use the second version. This makes the most sense too, imho, since when developers query products they get back a list of variants, so its reasonable they then takes those ProductVariant instances, and pass them back to Cart. 

With that in mind, I think we have two refactors here:

1. We make `public void AddOrUpdate(string variantId, long? quantity = null, Dictionary<string, string> customAttributes = null)` private (or remove it entirely). 
2. We change `CartLineItem` to accept a `ProductVariant`, instead of just an id.

@sleroux @dengjeffrey If you guys are good with that, I can do that as a separate, precursor PR to this.

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
richardmonette: I think once we fix the main issue, by working on the ProductVariant level, instead of separate id and price, this sort of becomes less of an issue, but yes, I think it would update it.

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
richardmonette: For sure, I think I'd break that out into another PR as well, as a precursor, so that can be a straight refactor which shouldn't be affecting any tests.

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
dengjeffrey: Yes this sounds good, I personally prefer we remove it entirely. And use 
`public void AddOrUpdate(ProductVariant variant, long? quantity = null, Dictionary<string, string> customAttributes = null) ` as the base.

To me a `CartLineItem` is a `ProductVariant` with a quantity. Extracting properties of `ProductVariant` to create a `CartLineItem` causes confusion and adds room for error.

scripts/generator/graphql_generator/csharp/SDK/CartLineItems.cs.erb
```diff
@@ -391,14 +412,26 @@ namespace <%= namespace %>.SDK {
...
                 }
             }
 
             _IsSaved = true;
         }
 
+        private ProductVariant GetVariantById(Product product, string variantId) {
+            List<ProductVariant> variants = (List<ProductVariant>) product.variants();
+
+            foreach(var variant in variants) {
```
sleroux: Nit: I think you can use `Find` here instead of iterating over the list yourself:

```csharp
var variants = (List<ProductVariant>)product.variants();
return variants.Find(variant => variant == variantID);
```

PR #300, opened 2017-08-25
[Fix] Apple pay error handling

dengjeffrey:
+ Fixes typo causing shipping address errors to e handled like billing address error
+ Adds shipping error handling within checkout user errors.
   + Setting a checkout to allow partial addresses voids the validation when making update address queries. So the addresses are validated on the checkout mutation

### DIFF
--- a/scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
@@ -100,34 +100,47 @@ namespace <%= namespace %>.SDK.iOS {
         }
 
         private static UpdateRequestStatus GetUpdateRequestStatusFromCheckoutUserErrors(List<UserError> errors) {
-
             var payErrors = new List<ApplePayError>();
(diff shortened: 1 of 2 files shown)


scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.ApplePayEventReceiver.cs.erb
```diff
@@ -100,65 +100,72 @@ namespace <%= namespace %>.SDK.iOS {
...
+                var fieldsSet = new HashSet<string>(fields);
 
-                    try {
-                        string lastField = fields[fields.Count - 1];
-                        var payError = new ApplePayBillingAddressInvalidError(description, lastField);
-                        payErrors.Add(payError);
-                    } catch {
-                        return new UpdateRequestStatus(ApplePayAuthorizationStatus.Failure);
+                try {
+                    string lastField = fields[fields.Count - 1];
```
sleroux: Since we removed the `Contains` statement, how do we know that `fields.Count` is not 0?

dengjeffrey: We don't, but the Try statement will catch it, in which case it is handled as an error

PR #306, opened 2017-09-12
Updated unitypackage/docs for 0.4.1 release

sleroux:


mikkoh:
I just landed #307 which should be in this PR. #307 adds the ability for us to track distribution method.

sleroux:
Closing this since there are more changes that will be landing prior to making a new release.

PR #309, opened 2017-09-13
[Docs] Added additional caveats to the Android building

sleroux:
Added some issues when trying to build for release from Unity 5.6.1 using the new Gradle build system.


mikkoh:
I don't have enough context to know how things work in Android land but I don't see anything wrong with documenting caveats. I wonder if it would be good to get a second opinion.

sleroux:
Closing this for now because we might have found a real solution.

Issue #310, opened 2017-09-15
 java.util.zip.ZipException: duplicate entry: android/support/annotation/RequiresPermission.class when using SDK


sleroux:
When bringing the SDK into a large project with other native plugins, the project will produce the following Gradle error:

```
Execution failed for task ':transformClassesWithJarMergingForRelease'.
> com.android.build.api.transform.TransformException: java.util.zip.ZipException: duplicate entry: android/support/annotation/RequiresPermission.class
```

Currently we link in the support-annotations library when we build our instrument tests for the shopify_buy_plugin .AAR we generate:

https://github.com/Shopify/unity-buy-sdk/blob/beta/android/shopify_buy_plugin/build.gradle#L29

The issue is potentially caused by other .AARs/.JARs that link in differing versions of the same library causing duplicate entries to be added to project's classes.jar. Google has a plugin that we might be able to use to break this dependency management issue across plugins:

https://github.com/googlesamples/unity-jar-resolver#plugin-redistribution

sleroux:
Note: This seems to only be an issue with multidex builds when building release variants.

PR #311, opened 2017-10-02
[Android] Excluded support-annotations and removed AAR dependencies

sleroux:
support-annotations is a popular library amongst Android plugins and causes conflicts when present within multiple plugins within a user's game. This usually only occurs when trying to make a signed release build for multidex-enabled builds. This PR excludes the dependency from our own dependencies to reduce the risk of collision.

### DIFF
--- a/Assets/Plugins/Android/mainTemplate.gradle
+++ b/Assets/Plugins/Android/mainTemplate.gradle
@@ -43,9 +43,4 @@ android {
 dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
     compile(name: 'shopify_buy_plugin', ext:'aar')
-
-    compile 'com.android.support:appcompat-v7:26.0.0'
-    compile 'com.android.support:customtabs:26.0.0'
-    compile 'com.shopify.mobilebuysdk:buy3-pay-support:1.0.0'
-    compile 'com.squareup.moshi:moshi:1.5.0'
 }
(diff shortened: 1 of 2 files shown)


sleroux:
Potential fix for https://github.com/Shopify/unity-buy-sdk/issues/310

Assets/Plugins/Android/mainTemplate.gradle
```diff
@@ -43,9 +43,4 @@ android {
 dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
     compile(name: 'shopify_buy_plugin', ext:'aar')
-
-    compile 'com.android.support:appcompat-v7:26.0.0'
-    compile 'com.android.support:customtabs:26.0.0'
-    compile 'com.shopify.mobilebuysdk:buy3-pay-support:1.0.0'
-    compile 'com.squareup.moshi:moshi:1.5.0'
```
dengjeffrey: we don't need these?
`'com.shopify.mobilebuysdk:buy3-pay-support:1.0.0'`
`compile 'com.squareup.moshi:moshi:1.5.0'`

sleroux: Turns out we do. I ran the example app without these and got `ClassNotFound` errors when trying to checkout out.

Now that I think of it, even if we exclude `support-annotations` from everything, if the plugin is added to a project that doesn't have another plugin that has it then that will also be missing which might cause issues within the dependencies.

I think the only way around all of this is trying to figure out how the ill-named https://github.com/googlesamples/unity-jar-resolver works. Looks like Facebook is using this for the same reason since when you pull in that .unitypackage it pulls out the dependency .jars into the plugins folder so the plugins can all use the same one instead of having duplicates.

I'm going to close this issue for now.

PR #313, opened 2017-10-02
[Fix] Set native payment to null for Android for now until feature is complete

sleroux:
Sets AndroidNativePayment to null for now until native Android Pay is implemented. This will allow all of the API calls to fail.

### DIFF
--- a/scripts/generator/graphql_generator/csharp/Cart.cs.erb
+++ b/scripts/generator/graphql_generator/csharp/Cart.cs.erb
@@ -130,7 +130,7 @@ namespace <%= namespace %> {
             NativeCheckout = new iOSNativeCheckout(State);
             #elif UNITY_ANDROID
             WebCheckout = new AndroidWebCheckout(this, client);
-            NativeCheckout = new AndroidNativeCheckout(State);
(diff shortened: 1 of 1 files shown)


scripts/generator/graphql_generator/csharp/Cart.cs.erb
```diff
@@ -130,7 +130,7 @@ namespace <%= namespace %> {
             NativeCheckout = new iOSNativeCheckout(State);
             #elif UNITY_ANDROID
             WebCheckout = new AndroidWebCheckout(this, client);
-            NativeCheckout = new AndroidNativeCheckout(State);
+            NativeCheckout = null;
```
mikkoh: In the rest of the codebase do we need to add `if (NativeCheckout  == null)`  checks and log with a message saying coming soon?

sleroux: For `CanCheckoutWithNativePay`, it returns false as expected and when trying to call `CheckoutWithNativePay` the `PlatformNotSupported` exception is thrown.

PR #318, opened 2017-10-05
[Cleanup] Removes dependency on moshi and uses standard Android JSON library for serializing/deserializing

sleroux:
I found a solution to our dependency problem https://github.com/Shopify/unity-buy-sdk/issues/310 which involves using the PlayServicesResolver to pull in all Google dependencies (like the support library stuff) but requires us to bake in the other kinds of dependencies like Moshi/Buy3. Buy3 only depends on the support library but Moshi relies on quite a few. Each dependency and each dependency's dependencies would need to be manually added to the Unity project. Since we're barely using Moshi for handling JSON, I've gone ahead and removed it from the project in favor of using the provided Android JSON library. This will allow us to keep the number of dependencies we need to include to a minimum.


android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/models/MailingAddressInput.java
```diff
@@ -1,22 +1,47 @@
...
+    private static final String LAST_NAME = "lastName";
+    private static final String PHONE = "phone";
+    private static final String PROVINCE = "province";
+    private static final String ZIP = "zip";
 
     public MailingAddressInput(UserAddress userAddress) {
         payAddress = PayAddress.fromUserAddress(userAddress);
     }
 
     public String toJsonString() {
```
flavioarfaria: Shouldn't this implement `JsonSerializable`?

PR #319, opened 2017-10-05
Import example project into main repo

JakeCataford:
Might be best to just 🎩  this one yourself instead of reading code.

This imports the examples that were worked on in another repo to this project. I dont think they will be packaged in the build just yet this way? but I can follow up with a few other PRs that do the small fixes.

Also, `cart.Subtotal()` was missing from the generated files last time I checked. So the cart subtotal in the ShopPopup doesn't work yet

This might also take some massaging to get it to pass CI


sleroux:
> I dont think they will be packaged in the build just yet this way? 

Yup it already gets added to the exported .unitypackage since we export everything inside the Shopify/ folder. I 🎩 by publishing and pulling it into a new project and everything gets imported fine. When building for iOS I do get some issues - I just don't remember if these were there before so I'll need to check. At first glance it looks like some of the editor scripts might not be running properly from SDK. Ill investigate and see what it could be.

JakeCataford:
I did add commits where some of the editor scripts were missing, found that when telling people to install and follow the tutorial from a package that included the example

sleroux:
Whenever I run `./scripts/test.sh` I get a bunch of syntax errors related to missing UnityEngine stuff. I think we might be including the Unity-specific files from the pop up in the paths we look to for the tests.

JakeCataford:
yea we definitely are, @sleroux do you know where this is configured?

sleroux:
https://github.com/Shopify/unity-buy-sdk/blob/master/scripts/test.sh#L12

JakeCataford:
@sleroux made the test script only compile what was in the "Unity" directory instead of the whole "Shopify" directory

sleroux:
@JakeCataford FYI - When I ran the `test_iOS` script it was failing because of some weird syntax error that doesn't seem related to any of these changes. I've gone ahead and pushed a patch for it so it works now

JakeCataford:
Not sure why these tests keep failing... but I have a few upstream PRs that have green tests so its definitely flaky... any objections to me merging this now @richardmonette @mikkoh?

sleroux:
The Android test is flaky because the string comparison for the produced JSON sometimes alters the ordering of it (since JSON isn't ordered) so I wouldn't worry about that. The Circle CI tests fail when testing on iOS because of a Unity thing I'm looking into. I haven't been able to find a solution in the past week so I wouldn't worry about it for now. I'll continue to investigate.

richardmonette:
iirc @sleroux is looking into the tests, should be OK to merge for now

PR #320, opened 2017-10-05
[Fix/Feature] Fixes duplicate dependency issues and removes Gradle requirement

sleroux:
Finally got to the root of https://github.com/Shopify/unity-buy-sdk/issues/310.

The main reason developers were experiencing duplicated classes when building in Gradle w/ multidex enabled is our mainTemplate.gradle file pulls in the Android dependencies. If another plugin also pulls in theirs (such as Facebook), a collision occurs. The way Facebook pulls in it's dependencies is using the PlayServicesResolver project [1]. 

Related Aside:

The Unity Jar Resolver is a pretty neat project. The problem with Android native plugins is that they usually always depend on the Google libraries (such as Play services or the support library). When you have multiple plugins within the same Unity app, you're bound to get collisions. This library runs a script whenever the developer switches platforms in the Unity editor and fetches the Google dependency JAR/AAR files and places them into the Plugins folder. Each plugin that wishes to use the PlayServicesResolver to fetch it's dependency needs to specify which ones in a script (in our case this is located in ShopifyDependencies.cs). 

At the end of the day, the solution was to do the following:

1. Instead of relying on mainTemplate.gradle to fetch our dependencies, use PlayServicesResolver.
2. Move any non-Google dependencies into the project. To make this cleaner, I've removed the Moshi library (https://github.com/Shopify/unity-buy-sdk/pull/318). The only dependency we need to add is our Buy3 mobile support AAR library.

On top of this, there were some additional tweaks and fixes the AndroidManifest that lets us build the app properly. Doing all this gives us some huge upsides:

1. Multidex builds are now supported w/ Gradle when exporting to Android Studio.
2. You don't need to use Gradle anymore to build the project. You can just use the default tooling!

[1] https://github.com/googlesamples/unity-jar-resolver


Assets/Plugins/Android/AndroidManifest.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="${applicationId}">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" xmlns:tools="http://schemas.android.com/tools" package="${applicationId}">
```
sleroux: Unity bug! Just needed to add the tools so when the developer specifies the non-default SDK requirement of 19 it works.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/models/PricingLineItems.java
```diff
@@ -4,35 +4,70 @@
...
+
+    private static BigDecimal nullableDecimalPropertyFromJson(JSONObject obj, String property) throws JSONException {
+        if (obj.has(property)) {
+            return decimalPropertyFromJson(obj, property);
+        }
+        return null;
+    }
+
+    private static BigDecimal decimalPropertyFromJson(JSONObject obj, String property) throws JSONException {
+        return new BigDecimal(obj.getString(property)).setScale(2, BigDecimal.ROUND_FLOOR);
```
sleroux: Note: All of these Moshi changes are in this PR https://github.com/Shopify/unity-buy-sdk/pull/318

scripts/generator/graphql_generator/csharp/Editor/ShopifyDependencies.cs.erb
```diff
@@ -0,0 +1,128 @@
+// <copyright file="ShopifyDependencies.cs" company="Google Inc.">
```
sleroux: This file is a template from the PlayServicesResolver library with changes to which dependencies we use.

scripts/publish.sh
```diff
@@ -79,12 +82,14 @@ printf "Exporting the unitypackage at: %s\n" "$UNITY_PACKAGE"
...
 "$UNITY_PATH" \
+    -gvh_disable \
     -batchmode \
     -nographics \
     -silent-crashes \
     -logFile "$UNITY_LOG_PATH" \
+    -importPackage "ThirdParty/play-services-resolver-1.2.12.0.unitypackage" \
     -projectPath "$PROJECT_ROOT" \
-    -exportPackage Assets/Shopify Assets/Plugins "$UNITY_PACKAGE" \
+    -exportPackage Assets/Shopify Assets/Plugins Assets/PlayServicesResolver "$UNITY_PACKAGE" \
```
sleroux: In order to defer the DLL loading functionality for PlayServicesResolver to work: https://github.com/googlesamples/unity-jar-resolver#plugin-redistribution

sleroux:
@JakeCataford Are you familiar at all with the PlayServicesResolver stuff? I've seen Facebook use it in their SDK and from what I've read it's kind of a thing plugin developers incorporate into their addon when they build out native plugins. Jeff is on vacation so I tagged you as a reviewer on this PR instead.  Note this also include code changes from https://github.com/Shopify/unity-buy-sdk/pull/318 that you can probably ignore.

JakeCataford:
I've never used this before but it makes sense to me... one question though: will we be shipping with the play services resolver embedded in the unitypackage? Or is this a pre-processing step to make sure we spit out the correct plugins? What happens if a shop wants to use the facebook plugin + the buy SDK? will there be a collision?

sleroux:
We ship the resolver within the package so it shows up alongside the Shopify root folder. From what I was reading, when we export the plugin it uses code delayed import thing to allow it to defer loading it's DLLs when imported into a project which handles versioning between two copies of PlayServicesResolver
https://github.com/googlesamples/unity-jar-resolver#plugin-redistribution

JakeCataford:
hmm kk, ill 🎩 this and if it works I don't see any reason to be afraid?

JakeCataford:
one thing is that this introduces a dev dependency on `java`... guess thats not that annoying tho haha

JakeCataford:
@sleroux tophatted this and im happy with what it spits out so 🍏 from me

PR #327, opened 2017-10-10
Optimize scrollrect performance

JakeCataford:
Removes layouting components and does the layout calculations once in
the code instead of every scroll frame. Changes UIMask For UIRectMask2D
which is significantly faster at masking. Also disables the "pixel
perfect" option on the UI canvas, which introduces a very significant
performance increase when scrolling without a perceptible difference in
quality.

Before:
screenshot_2017-10-10__12_01_pm

After:

screenshot_2017-10-10__11_56_am


sleroux:
Note: Probably needs a rebase to pass on Circle CI.

PR #342, opened 2017-10-12
Flavioarfaria/confirmation ui

flavioarfaria:
### Task description

After receiving a masked wallet, we can present the user with a confirmation screen which we'll have to overlay over the Unity Activity using our own Activity.

#### Task bullets

- [x] Create custom ConfirmationActivity with layout containing SupportWalletFragment in SELECTION_MODE;
- [x] Launch custom activity from our custom fragment modally.

### DIFF
--- a/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/MaskedWalletHolder.java
+++ b/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/MaskedWalletHolder.java
@@ -1,0 +1,25 @@
+package com.shopify.unity.buy;
+
+import android.content.Intent;
+
+import com.google.android.gms.wallet.MaskedWallet;
+
+/**
+ * Class that holds a {@link MaskedWallet} instance to be shared across
+ * {@link UnityAndroidPayFragment} and {@link ConfirmationActivity}.
+ *
+ * The masked wallet instance cannot be shared by {@link Intent}, since both components
+ * need to use the same object, not just a copy.
+ */
+
+public final class MaskedWalletHolder {
+
+    /**
+     * Unique shared {@link MaskedWallet} object.
+     */
+    public static MaskedWallet maskedWallet;
+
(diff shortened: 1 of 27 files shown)


android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/MaskedWalletHolder.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Class that holds a {@link MaskedWallet} instance to be shared across
+ * {@link UnityAndroidPayFragment} and {@link ConfirmationActivity}.
+ *
+ * The masked wallet instance cannot be shared by {@link Intent}, since both components
+ * need to use the same object, not just a copy.
+ */
+
+public final class MaskedWalletHolder {
```
sleroux: Is the holder pattern common in Android? I remember running into this problem where you can't pass things across intents but having the need to. Seems pretty nice to just stick it in a static variable when you're confident about how and when it gets accessed 👍

flavioarfaria: No, it's not, and it's extremely dangerous. If the user sends the app to background and the system decides to reclaim memory for other processes, this field will be vanished and the app will crash if things are not handled properly. The first thing I tried was to pass the masked wallet across activities as an extra, HOWEVER, when we do that, we're actually passing a copy of the object, not the object itself, and Android Pay doesn't like that because there can be only a single `MaskedWallet` instance throughout the checkout. I'll handle all the concerns related to the checkout state recovery in these extreme cases in a separate task.

flavioarfaria:
@sleroux just [fixed](https://github.com/Shopify/unity-buy-sdk/pull/342/commits/2d9aff08e54e02904d77aa8d2f4aa4a9c73c74e5) lint errors. Merging.

PR #343, opened 2017-10-12
[Fix] Fix CI issues for Android tests

sleroux:
Comparing string output of the Android JSON library is not a good idea since the ordering of properties is not guaranteed (which is expected but annoying). This patch modifies the class to allow being built from a JSON string so in the test we can reconstruct the object to do assertions on the properties. Also fixes an issue on Travis not accepting the latest build tools because we're using version 26 now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,9 @@ android:
   components:
     - tools
     - platform-tools
-    - build-tools-25.0.2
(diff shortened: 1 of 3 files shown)


sleroux:
Note: CircleCI will still fail. Just Travis is fixed since Circle is out iOS/Unity tests.

PR #348, opened 2017-10-17
[Cleanup] Refactors ILoader into BaseLoader with header extraction

sleroux:
Refactors the ILoader interface into an abstract base class named BaseLoader that is modelled after our ILoader in the C++ SDK. This allows the setting of headers to be done at the common base class level instead of inside UnityLoader and adds methods to enable this for future subclasses for different platforms.


scripts/generator/graphql_generator/csharp/SDK/BaseLoader.cs.erb
```diff
@@ -0,0 +1,58 @@
...
+        private string _Domain;
+
+        public BaseLoader(string domain, string accessToken) {
+            _AccessToken = accessToken;
+            _Domain = domain;
+
+            SetHeader("Content-Type", "application/graphql");
+            SetHeader("X-SDK-Version", "ShopifyBuy.VERSION");
+            SetHeader("X-Shopify-Storefront-Access-Token", accessToken);
+            SetHeader("X-SDK-Variant", SDKVariantName());
```
sleroux: By moving all this here and exposing an abstract method named `SDKVariantName`, we can factor out the required headers for communicating to our GraphQL services away from any SDK-specific loader.

scripts/generator/graphql_generator/csharp/SDK/UnityLoader.cs.erb
```diff
@@ -57,46 +46,32 @@ namespace <%= namespace %>.SDK {
...
+            .SetHeaders(_Headers)
             .Load(query, callback);
         }
+
+        override public void SetHeader(string key, string value) {
+            _Headers[key] = value;
+        }
+
+        override public string SDKVariantName() {
+            return "unity";
```
mikkoh: I think this Variant Name now needs to be generated so we can track whether the code was distributed via Github/Asset Store

scripts/generator/graphql_generator/csharp/SDK/BaseLoader.cs.erb
```diff
@@ -0,0 +1,58 @@
...
+
+        private string _AccessToken;
+        private string _Domain;
+
+        public BaseLoader(string domain, string accessToken) {
+            _AccessToken = accessToken;
+            _Domain = domain;
+
+            SetHeader("Content-Type", "application/graphql");
+            SetHeader("X-SDK-Version", "ShopifyBuy.VERSION");
```
mikkoh: I think this needs to be `ShopifyBuy.VERSION` cast to a String.

scripts/generator/graphql_generator/csharp/SDK/UnityLoader.cs.erb
sleroux: This was used to identify the engine code before and we use the `ShopifyBuy.VERSION` to indicate which version/distribution of the codebase we use (`github`/`asset-store`). Do we want to keep this as the engine identifier or change it to also identify the distro?

sleroux:
@mikkoh Should be good for re-review/merge

scripts/generator/graphql_generator/csharp/SDK/UnityLoader.cs.erb
mikkoh: Ah yes. At one time we were using the SDK variant name to differentiate and I decided later on that version would be a better place to do this.

PR #355, opened 2017-10-20
Send shipping line to Unity

flavioarfaria:
Whenever the user changes the shipping method, this change is propagated down to Unity to update the cart and then Unity propagates the updated pricing lines back up to refresh the UI.

### DIFF
--- a/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/ShopifyUnityPlayerActivity.java
+++ b/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/ShopifyUnityPlayerActivity.java
@@ -20,8 +20,6 @@ import com.shopify.unity.buy.models.CheckoutInfo;
 import com.shopify.unity.buy.models.ShippingMethod;
 import com.unity3d.player.UnityPlayerActivity;
 
-import static com.shopify.unity.buy.androidpay.view.widget.ShippingMethodSelectDialog.OnShippingMethodSelectListener;
-
 public class ShopifyUnityPlayerActivity extends UnityPlayerActivity
         implements AndroidPayCheckout.Listener {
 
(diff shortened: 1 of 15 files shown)


android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/ShopifyUnityPlayerActivity.java
```diff
@@ -14,16 +14,18 @@
...
 import com.shopify.unity.buy.androidpay.view.viewmodel.ConfirmationViewModel;
 import com.shopify.unity.buy.androidpay.view.widget.ConfirmationView;
+import com.shopify.unity.buy.androidpay.view.widget.ShippingMethodSelectDialog;
 import com.shopify.unity.buy.androidpay.view.widget.WalletFragmentInstaller;
+import com.shopify.unity.buy.models.CheckoutInfo;
 import com.shopify.unity.buy.models.ShippingMethod;
 import com.unity3d.player.UnityPlayerActivity;
 
-import java.util.List;
+import static com.shopify.unity.buy.androidpay.view.widget.ShippingMethodSelectDialog.OnShippingMethodSelectListener;
```
sleroux: Nit: Could you just import the package instead of statically importing the member?

scripts/generator/graphql_generator/csharp/SDK/ShippingMethod.cs.erb
```diff
@@ -16,10 +19,30 @@ namespace <%= namespace %>.SDK {
...
+            if (dataJSON.ContainsKey(AMOUNT)) {
+                amount = dataJSON[AMOUNT].ToString();
+            }
+            string identifier = null;
+            if (dataJSON.ContainsKey(IDENTIFIER)) {
+                identifier = dataJSON[IDENTIFIER].ToString();
+            }
+            string detail = null;
+            if (dataJSON.ContainsKey(DETAIL)) {
+                amount = dataJSON[DETAIL].ToString();
```
sleroux: Copy/paste error. Should be assigning to detail not amount!

flavioarfaria: Good catch! https://github.com/Shopify/unity-buy-sdk/pull/355/commits/3883d129eaf86fe8b1199fc3984352e95bb9055c

PR #356, opened 2017-10-24
Flavioarfaria/checkout confirmation

flavioarfaria:
Requests full wallet, extracts token, updates email address and completes checkout.

### DIFF
--- a/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/models/NativePayment.java
+++ b/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/models/NativePayment.java
@@ -20,11 +20,20 @@ public final class NativePayment implements JsonSerializable {
     private static final String TOKEN_DATA = "TokenData";
 
     @NonNull public final MailingAddressInput billingContact;
-    @NonNull public final ShippingContact shippingContact;
-    @Nullable public final String identifier;
+    @Nullable public final ShippingContact shippingContact;
+    @NonNull public final String identifier;
     @NonNull public final TokenData tokenData;
 
     private NativePayment(Builder builder) {
+        if (builder.billingContact == null) {
+            throw new IllegalArgumentException("Missing billingContact field");
+        }
+        if (builder.identifier == null) {
+            throw new IllegalArgumentException("Missing identifier field");
+        }
+        if (builder.tokenData == null) {
+            throw new IllegalArgumentException("Missing tokenData field");
+        }
         billingContact = builder.billingContact;
         shippingContact = builder.shippingContact;
         identifier = builder.identifier;
@@ -39,9 +48,11 @@ public final class NativePayment implements JsonSerializable {
     public JSONObject toJson() {
         final JSONObject json = new JSONObject();
         try {
-            json.put(BILLING_CONTACT, billingContact.toJson())
-                .put(SHIPPING_CONTACT, shippingContact.toJson())
-                .put(IDENTIFIER, identifier)
+            json.put(BILLING_CONTACT, billingContact.toJson());
+            if (shippingContact != null) {
+                json.put(SHIPPING_CONTACT, shippingContact.toJson());
+            }
+            json.put(IDENTIFIER, identifier)
                 .put(TOKEN_DATA, tokenData.toJson());
         } catch (JSONException e) {
             final String className = getClass().getSimpleName();
--- a/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/models/NativePayment.java
+++ b/android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/models/NativePayment.java
@@ -1,0 +1,88 @@
+package com.shopify.unity.buy.models;
+
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+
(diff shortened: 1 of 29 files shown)


flavioarfaria:
Since `ApplePayPayment` has been refactored into `NativePayment` that *might* have some impact on iOS. Since the only changes made were basically renaming the class and adding a new member, the impact, if exists, is minimal.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/models/NativePayment.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final String IDENTIFIER = "Identifier";
+    private static final String TOKEN_DATA = "TokenData";
+
+    @NonNull public final MailingAddressInput billingContact;
+    @NonNull public final ShippingContact shippingContact;
+    @Nullable public final String identifier;
+    @NonNull public final TokenData tokenData;
+
+    private NativePayment(Builder builder) {
+        billingContact = builder.billingContact;
```
sleroux: Probably minor but what happens if we never set `billingContact` in the builder and we assign null to this non-null property? Does it make sense to add a layer of validation in the `build` method to make sure we don't try to do this?

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.AndroidPayEventReceiver.cs.erb
```diff
@@ -32,6 +35,49 @@ namespace <%= namespace %>.SDK.Android {
...
+                NativeMessage message) {
+            CartState.CheckoutWithTokenizedPayment(tokenizedPaymentInput, (ShopifyError error) => {
+                AndroidPayCheckoutResponse.Status status = error == null
+                        ? AndroidPayCheckoutResponse.Status.Success
+                        : AndroidPayCheckoutResponse.Status.Failure;
+                message.Respond(new AndroidPayCheckoutResponse(status).ToJsonString());
+            });
+        }
+
+        private void SetFinalCheckoutFieldsForPayment(NativePayment payment, Checkout checkout,
```
sleroux: I don't see this being called anywhere. Shouldn't this be called just before performing the checkout?

flavioarfaria:
@sleroux regarding `SetFinalCheckoutFieldsForPayment()`, here's what it does on iOS:
1. Delegates to `CartState.SetFinalCheckoutFields()` (more on that later);
1. Tries to recover from user errors in its callback, such as invalid address information.

    There's not much to try to recover on Android Pay, since Android Pay itself validates address information in its own address form.

Regarding `CartState.SetFinalCheckoutFields()` (mentioned above in 1), it performs three mutations:

1. Email update;

    We're doing this in `OnConfirmCheckout()`.

1. Shipping address update;

    We're doing this whenever the user explicitly changes the address in the confirmation screen. That's why I needed to be able to send messages to Unity from an activity other than Unity's and that's why that single-activity approach came into play.

1. Shipping rate update.

    We're also doing this whenever the user explicitly does so.

To wrap up, it turns out I found `SetFinalCheckoutFieldsForPayment()` does a bunch of stuff we've done previously and we don't need to do it over (as far as I understand). The only thing it does we actually need to do is to update the email address, and we're calling it directly instead.

For 2. and 3., we're also setting defaults right after we get the masked wallet. The default shipping rate being set is the first one in the array. You can see this being done in `OnUpdateShippingAddress()` and `RespondToUpdateAddressSuccessForMessage()` respectively.

Let me know if I'm missing (or misunderstanding) something.

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/models/NativePayment.java
flavioarfaria: Of course it does. Will do it. :)

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.AndroidPayEventReceiver.cs.erb
flavioarfaria: This is garbage and must be removed.

sleroux: Ah okay. Regarding your comment below, having everything be set at each point in the flow is the best approach. I just saw this method here and no references being made to it. All good 👍

flavioarfaria: https://github.com/Shopify/unity-buy-sdk/pull/356/commits/4cc425f48b0758cf55b7724c2e4057b3b63f7db9

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/models/NativePayment.java
flavioarfaria: https://github.com/Shopify/unity-buy-sdk/pull/356/commits/3b721d459bbdfcf562a5a2f0900809b1796330fd

PR #357, opened 2017-10-25
Propagate Unity errors to Android plugin and lib client

flavioarfaria:


android/shopify_buy_plugin/src/main/res/values/strings.xml
```diff
@@ -10,4 +10,6 @@
     <string name="confirmation_shipping_method_price_not_available">N/A</string>
     <string name="confirmation_shipping_method_select_title">Select Shipping Method</string>
     <string name="confirmation_confirm">Confirm Order</string>
+    <string name="confirmation_address_error">Error updating shipping address</string>
+    <string name="confirmation_method_error">Error updating shipping method</string>
```
sleroux: Since this is a public facing error message we might want to make it less developer-like and more friendly:

"We've encountered an error while updating your shipping address. Please try again."

I'm not sure if the try again bit makes sense in this context though :P

flavioarfaria: https://github.com/Shopify/unity-buy-sdk/pull/357/commits/06face232ada17ecc10fa3fbb7fea2ac53b879b3

PR #360, opened 2017-10-30
Check for native pay support

flavioarfaria:
Added support for checking whether Android Pay is supported or not.


sleroux:
Looks like Travis is throwing linting errors but I can't seem to reproduce them on my local machine. Could you investigate?

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
```diff
@@ -16,10 +16,28 @@ namespace <%= namespace %>.SDK.Android {
         private String MerchantName;
         private String CountryCodeString;
         private AndroidPayEventReceiverBridge AndroidPayEventBridge;
+        #if !SHOPIFY_MONO_UNIT_TEST
+        private AndroidJavaObject _androidPayCheckoutSession;
+        private AndroidJavaObject AndroidPayCheckoutSession {
```
sleroux: This pattern seems a bit weird since we're using 2 private variables here. Could we not just create the session object in the constructor? Not sure what benefit we get by lazy loading it like this unless this was meant to be a public var.

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
```diff
@@ -16,10 +16,28 @@ namespace <%= namespace %>.SDK.Android {
...
+            get {
+                if (_androidPayCheckoutSession == null) {
+                    const string target = "com.shopify.unity.buy.androidpay.AndroidPayCheckoutSession";
```
sleroux: Nit: I would probably move this up to the class as a `const` string just to make it more obvious when looking at this source which class we're creating.

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
```diff
@@ -33,9 +51,16 @@ namespace <%= namespace %>.SDK.Android {
             // TODO
         }
 
-        public bool CanCheckout(PaymentSettings paymentSettings) {
-            // TODO: Invoke methods in the plugin to determine if we can perform Android Pay
-            return true;
+        public void CanCheckout(PaymentSettings paymentSettings, CanCheckoutWithNativePayCallback callback) {
+            if (paymentSettings.supportedDigitalWallets().Contains(DigitalWallet.ANDROID_PAY)) {
```
sleroux: I wonder if we should check to see if we already had a callback set and pass `false` to it before setting another one.

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
```diff
@@ -60,27 +85,23 @@ namespace <%= namespace %>.SDK.Android {
...
+                    CountryCodeString,
+                    requiresShipping
+            };
+            if (AndroidPayCheckoutSession.Call<bool>("checkoutWithAndroidPay", args)) {
+                if (AndroidPayEventBridge == null) {
+                    AndroidPayEventBridge = GlobalGameObject.AddComponent<AndroidPayEventReceiverBridge>();
+                    AndroidPayEventBridge.Receiver = this;
                 }
+            } else {
+                // TODO: Create more meaningful error?
```
sleroux: What kind of things could we tell the consumer about? Would this only fail because it's not enabled on their store? That would be a great thing to relay along if that's the case. We can also introduce additional error types if it makes sense here.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.cs.erb
```diff
@@ -48,15 +48,14 @@ namespace <%= namespace %>.SDK.iOS {
...
         /// <param name="paymentSettings">The Shop's payment settings</param>
+        /// <param name="callback">The callback that will deliver the result to tha caller</param>
         /// <returns>True if the device is capable of paying with Apple Pay and the store accepts payments with Apple Pay</returns>
-        public bool CanCheckout(PaymentSettings paymentSettings) {
-
+        public void CanCheckout(PaymentSettings paymentSettings, CanCheckoutWithNativePayCallback callback) {
             if (paymentSettings.supportedDigitalWallets().Contains(DigitalWallet.APPLE_PAY) &&
                 _CanCheckoutWithApplePay(SerializedPaymentNetworksFromCardBrands(paymentSettings.acceptedCardBrands()))) {
-                return true;
+                callback(true);
```
sleroux: Missing a return here or else you'll call the callback twice potentially.

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
flavioarfaria: C# doesn't seem to have backing fields for properties other than its own extra property dedicated for this purpose. I just wanted to avoid the Java-like 
`GetAndroidPayCheckoutSession()` approach, but you're right, there's really no benefit in lazy loading it, I did not think much about that. Thanks for the heads up.

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
flavioarfaria: Fair enough!

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
flavioarfaria: The best approach would be to have each callback mapped to a request identifier in order to make sure every callback is called with the right information at the right time just to enforce consistency in case the client decides to have multiple `CanCheckout()` calls running along each other. On the other hand, I thought it'd be overkill for a simple verification that is unlikely to run concurrently. What you suggested ensures no client call will be "callbackless", but the default `false` answer is not necessarily correct. Any thoughts on the overengineered approach I just mentioned?

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
flavioarfaria: Actually, this `else` will only run if the Java `checkoutWithAndroidPay()` method returns `false` due to invalid or missing arguments, or incorrect JSON representation. It's been like this since the beginning, but if it happens, it's more likely our SDK's fault rather than the game dev's. I honestly think we should just validate user input on the Unity side and omit this error propagation. If the execution falls in this `else` once, it will do it every time, since our SDK would have a bug in this case.

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.cs.erb
flavioarfaria: Good catch! Will fix it.

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
sleroux: I'd probably keep it simple then. I don't see this being an issue for more developers anyways.

sleroux:
Also I just noticed in the failing Circle CI tests for iOS that there is some missing references to PaymentNetwork. Something might have gotten mixed up when refactoring the PaymentNetwork code in C#?

flavioarfaria:
I strongly suspect Travis is broken. @sleroux see this: https://github.com/travis-ci/travis-ci/issues/8676.

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
flavioarfaria: https://github.com/Shopify/unity-buy-sdk/pull/360/commits/dd0e537d23c5f6c87b4d1cfb0ac469cecee70a79

scripts/generator/graphql_generator/csharp/SDK/iOS/iOSNativeCheckout.cs.erb
flavioarfaria: https://github.com/Shopify/unity-buy-sdk/pull/360/commits/c24aa7bb3dfa3770e1d73741004f421ee6831fd6

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
flavioarfaria: https://github.com/Shopify/unity-buy-sdk/pull/360/commits/b922c0be01824c8ce9ff43eb19506937080002fb

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
flavioarfaria: https://github.com/Shopify/unity-buy-sdk/pull/360/commits/61e063a40038be9856987e8bbb47e827c3ab61e4

scripts/generator/graphql_generator/csharp/SDK/Android/AndroidNativeCheckout.cs.erb
flavioarfaria: https://github.com/Shopify/unity-buy-sdk/pull/360/commits/82e61e32dd4888f6e521453d458bb65e248727ae

android/shopify_buy_plugin/src/main/java/com/shopify/unity/buy/androidpay/AndroidPayCheckoutSession.java
```diff
@@ -59,19 +58,14 @@ public void canCheckoutWithAndroidPay(String cardBrandsString) {
...
-    public boolean checkoutWithAndroidPay(
+    public void checkoutWithAndroidPay(
             String merchantName,
             String publicKey,
             String pricingLineItemsString,
             String currencyCode,
             String countryCode,
             boolean requiresShipping
     ) {
-        if (!PayHelper.isAndroidPayEnabledInManifest(rootActivity)) {
```
flavioarfaria: I suspect this check is not needed, since we explicitly add AP support to our manifest, right? Or is this something the client has to worry about?

sleroux: You're right - if we add it ourselves this check is redundant 🔥

flavioarfaria:
After rebasing on top of changes to `travis.yml`, CI is back to normal.

Issue #364, opened 2017-11-02
Not cleaning listeners after checkout success


coreypollock:
We received this message from Philip at Planet 316:

"My name is Philip, I am one of the developers that was in the meeting you had with Tony Fouts on Monday morning. I was testing out the checkout with WebView and completed a purchase successfully. I was able to get the success callback which simply printed out a message. While the iPad was still sitting at the in game test store, the device fell asleep. When I woke it up, I noticed I got another success callback, seeing the message print out again. This can also be replicated by locking and unlocking the iPad, triggering the same success message.

Our question was if this is the intended response to that situation? We are planning to use the callback to trigger a reward of coins in our games and want to make sure that there isn't a way to cheat the system. We can program our own checks to work around this but thought that it could be a bug in the system."

![img_0393 1](https://user-images.githubusercontent.com/12721181/32335241-c6f62d9a-bfc2-11e7-98f5-030c2aa77892.jpg)

sleroux:
Looks like we don't remove the `WebCheckoutMessageReceiver` from the Shopify global game object we setup after we call the cancel or success callbacks. We could fix this simply by destroying that component after we trigger the callback. 

This leads to the another problem though: If I go ahead to checkout using the web view then decide I want to do it later and come back into the game we'll send call the cancelled event. We'll destroy the listeners at this point. If I then go back into Safari to complete the checkout the game will never check to see if it was complete again leading to the player not getting their coins. 

Option 1 seems the most likely and will remove the false positives. @JakeCataford @richardmonette thoughts?

richardmonette:
So the issue you present, where if we remove the listener on 'cancel' then we wouldn't call it later on  'success' is I think the conundrum that caused this to be left unfixed until now. 

Given the limitation of not knowing whether the external browser has been closed, I would agree our best option is leave the listener on cancel, but remove it on success, as there should only ever be one success callback fired.

PR #365, opened 2017-11-08
[Fix] Destroy web checkout message receiver after we've successfully completed a checkout

sleroux:
Fixes https://github.com/Shopify/unity-buy-sdk/issues/364


scripts/generator/graphql_generator/csharp/SDK/WebCheckoutMessageReceiver.cs.erb
```diff
@@ -64,6 +64,7 @@ namespace <%= namespace %>.SDK {
                     var checkout = (Checkout) response.node();
                     if (checkout.completedAt() != null) {
                         OnSuccess();
+                        Destroy(this);
```
JakeCataford: This will only destroy the script and not the associated game object, is this intended?

sleroux:
Yup that was the intention 😸

PR #372, opened 2017-11-16
Create Unity Editor Loader

JakeCataford:
This is a loader class that handles loading when the normal game loop is
not active. It binds to the EditorApplication.update event and manually
ticks the www coroutine forward to complete the request.

This is important because if we want more robust, smart editors we need a way to query the API from a custom editor within unity.

### DIFF
--- a/Assets/Editor/TestUnityEditorLoader.cs
+++ b/Assets/Editor/TestUnityEditorLoader.cs
@@ -1,0 +1,33 @@
+﻿namespace Shopify.Tests {
+    using NUnit.Framework;
+    using UnityEngine.TestTools;
+    using UnityEngine;
+    using Shopify.Unity.SDK.Editor;
+    using Shopify.Unity.SDK;
+    using System.Collections;
+
+    [TestFixture]
+    public class TestUnityEditorLoader {
+        [UnityTest]
+        public IEnumerator TestWithValidQuery() {
+            var requestIsFinished = false;
+
+            string domain = "graphql.myshopify.com";
+            string authorization = "351c122017d0f2a957d32ae728ad749c";
+
+            var queryLoader = new QueryLoader(new UnityEditorLoader(domain, authorization));
+
+            queryLoader.Query(TestQueries.Query, (QueryResponse response) => {
+                Assert.True(response.HTTPError == null, "http errors were not null: " + response.HTTPError);
+                Assert.True(response.errors == null, "GraphQL errors were null");
+                Assert.True(response.data != null, "Received a response");
+                Assert.True(response.data.shop().name() != null, "graphql");
+                requestIsFinished = true;
+            });
+
+            while (!requestIsFinished) {
+                yield return null;
+            }
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


Assets/Editor/TestUnityEditorLoader.cs
```diff
@@ -0,0 +1,33 @@
...
+    using NUnit.Framework;
+    using UnityEngine.TestTools;
+    using UnityEngine;
+    using Shopify.Unity.SDK.Editor;
+    using Shopify.Unity.SDK;
+    using System.Collections;
+
+    [TestFixture]
+    public class TestUnityEditorLoader {
+        [UnityTest]
```
JakeCataford: This is a special testing feature that allows coroutines to be used so we can test the whole request cycle. This could be an integration test but the integration tests dont run in edit mode, and therefore dont give a good representation of the behaviour of the editor. These tests do run in edit mode so we can check the whole flow of the routine by making sure the request completes in the editor.

Assets/Editor/TestUnityEditorLoader.cs
```diff
@@ -0,0 +1,33 @@
...
+            queryLoader.Query(TestQueries.Query, (QueryResponse response) => {
+                Assert.True(response.HTTPError == null, "http errors were not null: " + response.HTTPError);
+                Assert.True(response.errors == null, "GraphQL errors were null");
+                Assert.True(response.data != null, "Received a response");
+                Assert.True(response.data.shop().name() != null, "graphql");
+                requestIsFinished = true;
+            });
+
+            while (!requestIsFinished) {
+                yield return null;
```
JakeCataford: In the editor we can't use `WaitUntil` so instead we just yield null

scripts/generator/graphql_generator/csharp/SDK/Editor/UnityEditorLoader.cs.erb
```diff
@@ -0,0 +1,80 @@
...
+                callback(null, w.error);
+            } else {
+                callback(w.text, null);
+            }
+
+            w.Dispose();
+        }
+
+        private void Update() {
+            foreach(var request in new List<IEnumerator>(_activeRequests)) {
```
sleroux: What's the purpose of using `new` here? Does this mean we're creating a new list on each update?

JakeCataford: We remove elements from `_activeRequests` within the loop, which is an illegal operation unless you have a copy of the list... there might be a linq way to do this but i think it would work the same way internally?

scripts/generator/graphql_generator/csharp/SDK/Editor/UnityEditorLoader.cs.erb
```diff
@@ -0,0 +1,80 @@
...
+                if (!request.MoveNext()) {
```
richardmonette: This is basically just removing the last element?

JakeCataford: no, the elements in this list are essentially coroutines. In unity, coroutines are just an IEnumerator where MoveNext returns true if it should keep running or false if it is finished. the WWW will start a background download, so we just need to wait until www.isDone is true, the coroutine that wraps the download and is stored in activeRequests will yield true for MoveNext until the function exits, in which case we dont need to tick the reference anymore so we can just remove it from the list of references we need to tick.

scripts/generator/graphql_generator/csharp/SDK/Editor/UnityEditorLoader.cs.erb
sleroux: Ah okay I thought that might be it. The good ol' delete-and-iterate.

PR #375, opened 2017-11-20
[UI Toolkit] Create base theme controller class

JakeCataford:
Creates a base theme controller that contains the access token and shop
domain, as well as the ability to initialize a client.

Deleted the sanity test as it was only used to make sure the test setup stayed sane, now with a real test in that folder we dont need it anymore


Assets/Shopify/UIToolkit/ThemeControllerBase.cs
```diff
@@ -0,0 +1,28 @@
...
+    public abstract class ThemeControllerBase : MonoBehaviour {
+        /// <summary>
+        /// The Shop Domain to connect to, in the form of "myshop.myshopify.com"
+        /// </summary>
+        public string ShopDomain;
+
+        /// <summary>
+        /// The Storefront Access Token used to authenticate with the connected shop
+        /// </summary>
+        public string AccessToken;
```
sleroux: Should these properties be `readonly`?

sleroux: With the lazy loading code below it might be safer to restrict the developer from changing these because you can into the scenario where you lazy load the client with one domain/token pair the update it but you'll still be using the old cached client.

JakeCataford: Hmm this is problematic because readonly properties will not be serialized properly by unity and will be empty on save/restore/play, as well as writing to these fields using the editor....

We could also sanity check the cached client to make sure the shop domain and access token match? then we would be able to produce a new client if these are changed at runtime.

JakeCataford:
@sleroux now it invalidates if the shop domain or the access token dont match the one in the client

PR #378, opened 2017-11-21
Freeze Generated File Namespaces

JakeCataford:
Freezing namespaces means that we can have non-generated files
referencing these namespaces without fear that it might change at some
point.


JakeCataford:
@sleroux yea i would like to do that... i think there are some prereqs to make sure that they are checked into git etc

mikkoh:
@sleroux I think @JakeCataford's idea was to freeze the namespace. See if it causes issues anywhere and then move of files if it still makes sense.

PR #379, opened 2017-11-21
[UI Toolkit] Scaffold Single Product Theme

JakeCataford:
Provides the scaffolding and docs for a single product theme

avoiding the binding methods right now because they depend on the theme controllers being around.

### DIFF
--- a/Assets/Shopify/UIToolkit/Themes/SingleProductTheme.cs
+++ b/Assets/Shopify/UIToolkit/Themes/SingleProductTheme.cs
@@ -1,0 +1,15 @@
+﻿namespace Shopify.UIToolkit {
+    using Shopify.Unity;
+
+    /// <summary>
+    /// Override this class to create a theme that sells a single product
+    /// </summary>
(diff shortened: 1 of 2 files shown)


Assets/Shopify/UIToolkit/Themes/SingleProductTheme.cs
```diff
@@ -0,0 +1,14 @@
...
+
+    /// <summary>
+    /// Override this class to create a theme that sells a single product
+    /// </summary>
+    public abstract class SingleProductTheme : ThemeBase {
+        /// <summary>
+        /// Called when the theme should show information about the product it's selling
+        /// </summary>
+        /// <param name="product">The product that should be displayed</param>
+        public abstract void OnShouldShowProduct(Product product);
```
JakeCataford: I thought about maybe also providing ProductVariant[] in this callback, to avoid implementors having to know how to dig through the gql structure of edges and nodes to get at the variants

PR #380, opened 2017-11-21
[UI Toolkit] Collection theme scaffolding

JakeCataford:
A scaffold of methods that should be in the collection theme

### DIFF
--- a/Assets/Shopify/UIToolkit/Themes/CollectionTheme.cs
+++ b/Assets/Shopify/UIToolkit/Themes/CollectionTheme.cs
@@ -1,0 +1,41 @@
+﻿using Shopify.Unity;
+
+namespace Shopify.UIToolkit {
+    /// <summary>
+    /// A base class for a theme that sells products from a single collection
+    /// </summary>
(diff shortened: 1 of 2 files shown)


Assets/Shopify/UIToolkit/Themes/CollectionTheme.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// </summary>
+        /// <param name="product">The product to show details for</param>
+        /// <param name="variants">The variants of the product</param>
+        public abstract void OnShouldShowProductDetails(Product product, ProductVariant[] variants);
+
+        /// <summary>
+        /// Called when the implementor should show the cart UI with the provided line items.
+        /// </summary>
+        /// <param name="lineItems">The line items to display in the cart</param>
+        public abstract void OnShouldShowCart(CheckoutLineItem[] lineItems);
```
JakeCataford: Worth noting that im using CheckoutLineItems here but this is subject to change... depends on how we implement the cart functionality... but ill leave this in for now as is

PR #382, opened 2017-11-21
[WIP] Remote Image Loader + Web Resource Cache

sleroux:
* Implements a RemoteImageLoader behaviour that fetches a web image using a given URL.
* RemoteImageLoader internally caches images using Last-Modified/If-Modified-Since HTTP Headers.
* WebImageCache implements a LRU memory cache for storing web textures (PNG/JPEG)


sleroux:
@JakeCataford Pushing up the work I did today for the image loader/caching stuff. It should almost be done - just need to fix up some of the tests. Let me know what you think.

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,140 @@
...
+            _loader = loader;
+        }
+
+        /// <summary>
+        /// Downloads the given web image located at the URL.
+        /// </summary>
+        /// <param name="url">URL of the image resource to download and cache.</param>
+        /// <param name="cache">Determine if the downloaded resource should be cached or not.</param>
+        /// <returns>On completion, check texture/error properties for result.</returns>
+        public IEnumerator DownloadImageURL(string url, bool cache = true) {
```
sleroux: Nit: I'm not using the `cache` flag yet... I probably should.

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        public IEnumerator Load(string url, Dictionary<string, string> headers) {
+            _www = new WWW(url, null, headers);
+            yield return _www;
+        }
+    }
+
+    /// <summary>
+    /// A behaviour that fetches and caches a remote image (JPEG/PNG) from the web.!-- from the web.
```
JakeCataford: something weird with this doc here?

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,140 @@
...
+            public const string LAST_MODIFIED = "Last-Modified";
+            public const string STATUS_CODE = "STATUS";
+            public const string IF_MODIFIED_SINCE = "If-Modified-Since";
+        }
+
+        /// <summary>
+        /// The resulting error after downloading a web image.
+        /// </summary>
+        /// <returns></returns>
+        public string error {
```
JakeCataford: nit: public properties should be `Pascal Case`

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,140 @@
...
+            yield return _www;
+        }
+    }
+
+    /// <summary>
+    /// A behaviour that fetches and caches a remote image (JPEG/PNG) from the web.!-- from the web.
+    /// </summary>
+    public class RemoteImageLoader : MonoBehaviour {
+
+        private class HTTPHeaderFields {
```
JakeCataford: maybe static too? idk why it would be better just feels right?

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
JakeCataford: The implementor is responsible for enumerating this coroutine? Maybe we could make this a private method and create a public method with optional callbacks that fire after this coroutine runs? That way the primary use case is more "fire and forget" with an image URL...

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,140 @@
...
+        public IEnumerator Load(string url, Dictionary<string, string> headers) {
+            _www = new WWW(url, null, headers);
+            yield return _www;
+        }
+    }
+
+    /// <summary>
+    /// A behaviour that fetches and caches a remote image (JPEG/PNG) from the web.!-- from the web.
+    /// </summary>
+    public class RemoteImageLoader : MonoBehaviour {
```
JakeCataford: If this is a behaviour, why doesn't it collaborate directly with an Image component? If its designed more as a service to just support the image loading, I would rather see it as a plain c# object instead of a behaviour. You can still create a gameobject in the constructor and use that game object to run a coroutine if thats the reason why its a MonoBehaviour in the first place too?

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,140 @@
...
+            // Bail out early if we hit an error.
+            if (_loader.GetError() != null) {
+                _error = _loader.GetError();
+                yield return null;
+            }
+
+            var responseHeaders = _loader.GetResponseHeaders();
+
+            // Check to see if we got a 304 Not Modified response. In that case, just return the cached image.
+            if (responseHeaders.ContainsKey(HTTPHeaderFields.STATUS_CODE)) {
```
JakeCataford: If we have to make the request anyways, why not just use the downloaded image? The damage is already done? Unless I'm missing something?

Assets/Shopify/UIToolkit/Components/WebImageCache.cs
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    /// <summary>
+    /// This class implements a LRU cache for images (PNG/JPEG) that were downloaded from the web.
+    /// </summary>
+    public class WebImageCache {
+        /// <summary>
+        /// The default estimated memory size limit in bytes.
+        /// </summary>
+        public const int DEFAULT_MEMORY_SIZE_LIMIT = 20971520; // 20 MB in bytes
```
JakeCataford: Where'd you get this limit from?

Assets/Shopify/UIToolkit/Components/WebImageCache.cs
```diff
@@ -0,0 +1,163 @@
...
+
+        private static WebImageCache _sharedCache;
+
+        /// <summary>
+        /// Shared WebImageCache instanced.
+        /// </summary>
+        /// <returns>A singleton instance of the WebImageCache.</returns>
+        public static WebImageCache SharedCache {
+            get {
+                _sharedCache = _sharedCache != null ? _sharedCache : new WebImageCache(DEFAULT_MEMORY_SIZE_LIMIT);
```
JakeCataford: you can use null propogation here `_sharedCache = _sharedCache ?? new Whatever()`, its a lil more concise?

Assets/Shopify/UIToolkit/Components/WebImageCache.cs
```diff
@@ -0,0 +1,163 @@
...
+                return _sharedCache;
+            }
+        }
+
+        /// <summary>
+        /// The estimated size limit in bytes of how much memory we want to use for the cache (readonly).
+        /// </summary>
+        /// <returns></returns>
+        public int MemorySizeLimit {
```
JakeCataford: You can probably clean up some of these fields by just doing `public int MemorySizeLimit { get; private set; }`

JakeCataford:
Would have also been nice to split this into 2 prs, one for the image loading (naiive) and one for the cached loading

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
sleroux: I wasn't sure what the most common way of interfacing this would be but callbacks work for me!

JakeCataford: yea my intuition is that a dev will probably want all of the network/async concern hidden and will pretty much 90% of the time just call `RemoteImageLoader.LoadImageURL(url)` and forget about it...

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
sleroux: Reading the docs it looks like in order to start a coroutine from a game object I'll need to add a component to it and call `StartCoroutine` from that. Seems like I can either create a dummy behaviour or use this loader as a behaviour that references an image component. What do you think of doing the latter? The cache is shared between all instances that would be created so we could have many loaders using the same cache.

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
sleroux: In most cases we can probably get by with just using the saved texture directly. Using the Last-Modified/If-Not-Modified dance gives us the ability to send a super cheap partial GET and only download the image if it has changed. For our in-memory cache we probably don't need this complexity but we could cache the images on disk using this method without worrying about them getting stale. Might be nice to implement a disk cache for these to make the store images load instantly on subsequent launches of the game.

Assets/Shopify/UIToolkit/Components/WebImageCache.cs
sleroux: I pulled it out of thing air 👀 I'll do some analysis over what a good default size would be.

If we keep in the Last-Modified stuff we can also add a disk caching layer to keep more textures available.

sleroux:
Split up the naiive web loader into this PR https://github.com/Shopify/unity-buy-sdk/pull/383

sleroux:
Web image cache https://github.com/Shopify/unity-buy-sdk/pull/384.
 
I'm going to close this PR so we can move discussions to the other two relevant PRs.

PR #383, opened 2017-11-22
[Feature] Implements naiive image loader behaviour

sleroux:
* Adds a RemoteImageLoader behaviour for downloading images without caching logic.


richardmonette:
Did you have a look at https://github.com/Shopify/unity-buy-sdk/blob/master/Assets/Shopify/Examples/Scripts/Helpers/ImageHelper.cs#L12 ? A version of this already exists in the vending machine example, which includes caching for image loads.

sleroux:
Oh interesting - nope I didn't see that.

How would you feel about removing ImageHelper.cs in favour of this for the example app? I'm hoping to extend this later with more complex caching.

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,76 @@
...
+
+    public interface ImageLoader {
+        string GetError();
+
+        Texture2D GetTexture();
+
+        IEnumerator Load(string url);
+    }
+
+    public class UnityImageLoader : ImageLoader {
```
JakeCataford: this feels kinda Yagni... when are we going to need a non-unityloader in the unity UI Kit?

richardmonette:
That seems fine, my thought was just that it'd be great if we could use the same functionality in both places (incl. caching).

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// Completion callback that returns the downloaded texture and error description from a load operation.
+    /// </summary>
+    /// <param name="texture">Downloaded image as a Texture2D</param>
+    /// <param name="error">Error description</param>
+    public delegate void RemoteImageCompletionDelegate(Texture2D texture, string error);
+
+    /// <summary>
+    /// A behaviour that fetches a remote image (JPEG/PNG) from the web.
+    /// </summary>
+    public class RemoteImageLoader : MonoBehaviour {
```
JakeCataford: Same comment as before you broke this pr up into different pieces... if this is a service object for loading images it should not be a component... if its meant to be used alongside a UI.Image then we can abstract away getting the downloaded image an assigning it to a UI.Image class by including a UI.Image as a dependency.

My reasoning is that components describe some component of the total behaviour of an object in the game, if this is just a tool to load images remotely then creating a gameobject is extra work because the component doesnt need any features from the game object itself (for example, configuration in the editor or collaborating with neighboring components). Its best to keep it as a "code only" object as a result, because that is the intended way to interact with it.

If you need monobehaviour for the coroutine, another approach might be to have a static reference to a lazy instantiated GameObject, then you can call StartCoroutine on that game object to run a coroutine stored in your plain c# object

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            yield return _loader.Load(url);
+
+            // Bail out early if we hit an error.
+            if (_loader.GetError() != null) {
+                completion(null, _loader.GetError());
+                yield return null;
+            }
+
+            Texture2D downloadedTexure = _loader.GetTexture();
```
JakeCataford: I'm impartial to using the implicit `var` here?

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,76 @@
...
+            completion(downloadedTexure, null);
+            yield return null;
```
JakeCataford: this will skip a frame at the end of the enumerator before it cleans itself up... you can let the enumerator just fall through here because you are guaranteed to yield when starting the download

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        private IEnumerator LoadImageURLRoutine(string url, RemoteImageCompletionDelegate completion) {
+            _loader = _loader != null ? _loader : new UnityImageLoader();
+
+            yield return _loader.Load(url);
+
+            // Bail out early if we hit an error.
+            if (_loader.GetError() != null) {
+                completion(null, _loader.GetError());
+                yield return null;
```
JakeCataford: this won't break out of the enumerator, you might be looking for `yield break`?

Assets/Shopify/UIToolkit/Test/Integration/TestRemoteImageLoaderIntegration.cs
```diff
@@ -0,0 +1,33 @@
...
+            string url = "https://cdn.shopify.com/s/files/1/2094/7261/products/product-image-305453751.jpg?v=1497377691";
+
+            Texture2D downloadedTexture = null;
+            string returnedError = null;
+            loader.LoadImageURL(url, (texture, error) => {
+                downloadedTexture = texture;
+                returnedError = error;
+            });
+
+            while (downloadedTexture == null) {
```
JakeCataford: if the network is borked for whatever reason this test will run forever... maybe just make a temp bool called `requestFinished` and wait on that instead?

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
JakeCataford: oh i see in the tests... clever!

sleroux: Yup - mostly just to stub out the network with a mocked call.

sleroux:
@JakeCataford Simplified it quite a bit with your suggestions

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        private Sprite SpriteFromTexture(Texture2D texture) {
+            return Sprite.Create(
+                texture,
+                new Rect(0, 0, texture.width, texture.height),
+                new Vector2(0.5f, 0.5f),
+                100f,
+                0,
+                SpriteMeshType.FullRect
```
JakeCataford: nice catch... this one is a horrible performance gotcha if you let it trace the outlines 🌮

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,77 @@
...
+        private delegate void RemoteImageCompletionDelegate(Texture2D texture, string error);
+
+        void Start() {
+            _image = gameObject.GetComponent<Image>();
+        } 
+
+        public void LoadImage() {
+            LoadImageURL(ImageURL, (texture, error) => {
+                if (error != null) {
+                    Debug.Log("Failed to download image at " + ImageURL + " Reason: " + error);
```
JakeCataford: Not in this pr, but we could add a spot to stick a sprite to show when the image is broken or loading in this script too...

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,77 @@
...
+                    return;
+                }
+
+                if (texture == null) {
+                    Debug.Log("Failed to generate texture from image located at " + ImageURL);
```
JakeCataford: theres also Debug.Warning... this feels warningey

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,77 @@
...
+                SpriteMeshType.FullRect
+            );
+        }
+
+        /// <summary>
+        /// Downloads the given web image located at the URL.
+        /// </summary>
+        /// <param name="url">URL of the image resource to download and cache.</param>
+        /// <param name="completion">Callback called when the download is complete.</param>
+        void LoadImageURL(string url, RemoteImageCompletionDelegate completion) {
```
JakeCataford: can probably drop the docs if this method is not public?

JakeCataford: also we've done explicit private everywhere else (maybe its a javaism?) should we just keep with it?

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        private Image _image;
+
+        private delegate void RemoteImageCompletionDelegate(Texture2D texture, string error);
+
+        void Start() {
+            _image = gameObject.GetComponent<Image>();
+        } 
+
+        public void LoadImage() {
```
JakeCataford: Should have docs here

JakeCataford: Also why not pass the URL in here? Then you wont have a useless field in the editor for the url (if you know what the image is gonna be ahead of time then you might as well download it and get mipmaps and other nice things)

Assets/Shopify/UIToolkit/Test/Integration/TestRemoteImageLoaderIntegration.cs
```diff
@@ -0,0 +1,29 @@
...
+        [UnityTest] 
+        public IEnumerator TestDownloadImage() {
+            var gameObject = new GameObject();
+            var loader = gameObject.AddComponent<RemoteImageLoader>();
+
+            loader.ImageURL = "https://cdn.shopify.com/s/files/1/2094/7261/products/product-image-305453751.jpg?v=1497377691";
+            loader.LoadImage();
+
+            var image = gameObject.GetComponent<Image>();
+            while (image.sprite == null) {
```
JakeCataford: same deal, if the network is broken on ci or on someones machine this tests will run forever... you might need to check the error as well here or include an overload to loader.loadimage that notifies when the request is complete (I like it with no parameters too, but in testing and for some cases a more robust method might be better?

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -0,0 +1,77 @@
...
+            yield return www;
+
+            // Bail out early if we hit an error.
+            if (www.error != null) {
+                completion(null, www.error);
+                yield break;
+            }
+
+            completion(www.texture, null);
+            yield break;
```
JakeCataford: dont need to break here, you can let the method fall through

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
sleroux: We can explicit private it. This was mostly a carry over from the previous iteration of this PR.

Assets/Shopify/UIToolkit/Test/Integration/TestRemoteImageLoaderIntegration.cs
```diff
@@ -0,0 +1,36 @@
+#if !UNITY_IOS
+namespace Shopify.UIToolkit.Test.Integration {
+	using UnityEngine;
+    using UnityEngine.UI;
+    using UnityEngine.TestTools;
+    using System.Collections;
+    using Shopify.Unity.SDK;
+
+	[IntegrationTest.DynamicTest("UIToolkitIntegrationTests")]
```
richardmonette: nit: identation

Assets/Shopify/UIToolkit/Test/Integration/TestRemoteImageLoaderIntegration.cs
```diff
@@ -0,0 +1,36 @@
+#if !UNITY_IOS
+namespace Shopify.UIToolkit.Test.Integration {
+	using UnityEngine;
```
richardmonette: nit: indentation

sleroux:
@JakeCataford good to go 👍?

Assets/Shopify/UIToolkit/Test/Integration/TestRemoteImageLoaderIntegration.cs
```diff
@@ -0,0 +1,36 @@
+#if !UNITY_IOS
+namespace Shopify.UIToolkit.Test.Integration {
+    using UnityEngine;
+    using UnityEngine.UI;
+    using UnityEngine.TestTools;
+    using System.Collections;
+    using Shopify.Unity.SDK;
+
+    [IntegrationTest.DynamicTest("UIToolkitIntegrationTests")]
+    [IntegrationTest.Timeout(10)]
```
JakeCataford: I think @mikkoh removed this... can you rebase on master to make sure this still compiles without the test tools?

PR #384, opened 2017-11-22
[Feature] Web Image Cache

sleroux:
* Adds an in-memory LRU cache for storing Texture2Ds against URLs
* Uses HTTP Last-Modified timestamps to keep track of staleness of data.

### DIFF
--- a/Assets/Shopify/UIToolkit/Components/WebImageCache.cs
+++ b/Assets/Shopify/UIToolkit/Components/WebImageCache.cs
@@ -26,7 +26,9 @@
         /// </summary>
         public const int DEFAULT_MEMORY_SIZE_LIMIT = 20971520; // 20 MB in bytes
 
-        private Dictionary<string, CachedWebResource<Texture2D>> _urlTextureCache = new Dictionary<string, CachedWebResource<Texture2D>>();
+        private Dictionary<string, CachedWebResource<Texture2D>> _urlTextureCache = 
+            new Dictionary<string, CachedWebResource<Texture2D>>();
+
         private LinkedList<string> _recentlyUsed = new LinkedList<string>();
 
         private static WebImageCache _sharedCache;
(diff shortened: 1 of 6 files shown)


Assets/Shopify/UIToolkit/Components/WebImageCache.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// <returns></returns>
+        public int MemorySizeLimit { get; private set; }
+
+        /// <summary>
+        /// The current estimated size the cache takes up in memory (readonly)
+        /// </summary>
+        /// <returns></returns>
+        public int EstimatedMemorySize { get; private set; }
+
+        public WebImageCache(int memorySizeLimit) {
```
JakeCataford: if there is a singleton, should the constructor be public?

Assets/Shopify/UIToolkit/Components/WebImageCache.cs
```diff
@@ -0,0 +1,156 @@
...
+        public void Clear() {
+            _urlTextureCache.Clear();
+            _recentlyUsed.Clear();
+            EstimatedMemorySize = 0;
+        }
+
+        private int EstimateMemoryFootprintForTexture(Texture2D texture) {
+            // Since we're only dealing with JPEG (RGB24) or PNG (ARGB32) formats, we can assume 4 bytes per.
+            // We can also assume these images will only have a single mipmap level.
+            return texture.width * texture.height * 4;
```
richardmonette: Could be made more accurate by checking https://docs.unity3d.com/ScriptReference/Texture2D-format.html for a few common formats, and then fallback on assuming 4.

Assets/Shopify/UIToolkit/Components/WebImageCache.cs
richardmonette: 🤔

JakeCataford: as in shouldnt the singleton method be the only access point, therefore the constructor be private lol... bad rhetoric

sleroux: Heh yea it should be private 😶

sleroux:
I was thinking this morning that even though this is called a `WebImageCache`, it would be trivial to make this into a generic LRU cache that we can also use for requests.

JakeCataford:
> I was thinking this morning that even though this is called a WebImageCache, it would be trivial to make this into a generic LRU cache that we can also use for requests.

I agree, but we should refactor forward

richardmonette:
Looking at that example image, it's 800*800*4 channels(RGBA)*byte per channel = ~2.5mb so that seems about right for the space it would require.

sleroux:
> Looking at that example image, it's 8008004 channels(RGBA)*byte per channel = ~2.5mb so that seems about right for the space it would require.

That's quite a bit considering game developers probably want to optimize how much memory they are using at any given time. Let's say our default page size for the SDK is now 10 products. If we're rendering a list view, we would want to render 10 images x 2.5 mbs = 25mbs of memory space.

Also I'm not sure if it's an actual issue because I'd imagine a layer on top of the cache requesting images that are small enough to fit better in memory. I think once we add in that layer on top that starts pulling in images we can see if we want to implement disk cache layer as well.

sleroux:
I'm just going to close this PR since https://github.com/Shopify/unity-buy-sdk/pull/403 is pretty much v2 of this.

PR #385, opened 2017-11-22
[wip] Run Mono Tests in CI so we don't break them

JakeCataford:
wip


JakeCataford:
closing cause I dont have time to follow up

PR #387, opened 2017-11-22
Product pagination

mikkoh:
This PR does the following:
+ Forces pagination. You can no longer load all products
+ Drops the maximum amount of products loaded per page to 20 (this arbitrary and I wonder if we should think about it more)
+ Forces the user to make callbacks that accept `after` eg. `(products, errors, after)`
+ `after` will be `null` if there are no more pages

### DIFF
--- a/Assets/Editor/MockLoaderProducts.cs
+++ b/Assets/Editor/MockLoaderProducts.cs
@@ -21,9 +21,9 @@ namespace Shopify.Tests {
 
                 DefaultQueries.products.ShopProducts(
                     query: query,
-                    first: DefaultQueries.MaxProductsSize,
+                    first: DefaultQueries.MaxProductPageSize,
                     imageResolutions: ShopifyClient.DefaultImageResolutions,
-                    after: i > 0 ? (i * DefaultQueries.MaxProductsSize - 1).ToString() : null
+                    after: i > 0 ? (i * DefaultQueries.MaxProductPageSize - 1).ToString() : null
                 );
 
                 string response = String.Format(@"{{
@@ -48,8 +48,8 @@ namespace Shopify.Tests {
         private string GetProductEdges(int page) {
             StringBuilder edges = new StringBuilder();
 
-            for(int i = 0; i < DefaultQueries.MaxProductsSize; i++) {
-                int product = page * DefaultQueries.MaxProductsSize + i;
+            for(int i = 0; i < DefaultQueries.MaxProductPageSize; i++) {
+                int product = page * DefaultQueries.MaxProductPageSize + i;
                 bool imagesHasNextPage = product == 1 || product == 2;
                 bool variantsHasNextPage = product == 2;
 
@@ -127,7 +127,7 @@ namespace Shopify.Tests {
                 GetCollectionsOnProduct(0, product, 1),
                 UtilsMockLoader.GetJSONBool(false),
                 resolutionImageResponses.ToString(),
-                i < DefaultQueries.MaxProductsSize - 1 ? "," : ""));
+                i < DefaultQueries.MaxProductPageSize - 1 ? "," : ""));
             }
 
             return edges.ToString();
--- a/Assets/Editor/MockLoaderProducts.cs
+++ b/Assets/Editor/MockLoaderProducts.cs
@@ -21,9 +21,9 @@ namespace Shopify.Tests {
 
                 DefaultQueries.products.ShopProducts(
                     query: query,
-                    first: DefaultQueries.MaxPageSize,
+                    first: DefaultQueries.MaxProductsSize,
                     imageResolutions: ShopifyClient.DefaultImageResolutions,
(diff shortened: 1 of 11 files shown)


Assets/Editor/MockLoaderProducts.cs
```diff
@@ -21,9 +21,9 @@ public class MockLoaderProducts : BaseMockLoader {
 
                 DefaultQueries.products.ShopProducts(
                     query: query,
-                    first: DefaultQueries.MaxPageSize,
+                    first: DefaultQueries.MaxProductsSize,
```
richardmonette: `MaxProductPageSize`?

scripts/generator/graphql_generator/csharp/SDK/Delegates.cs.erb
```diff
@@ -18,6 +18,7 @@ namespace Shopify.Unity.SDK {
     public delegate void MergeFieldDelegate(string field, Dictionary<string, object> into, Dictionary<string, object> responseA, Dictionary<string, object> responseB);
 
     public delegate void ProductsHandler(List<Product> products, ShopifyError error);
```
richardmonette: Is there a scenario where we are still using this? I'm thinking we should remove all the non-paginated convenience methods, in which case this could also be removed.

scripts/generator/graphql_generator/csharp/ShopifyBuy.cs.erb
```diff
@@ -849,9 +856,17 @@ namespace Shopify.Unity {
...
                     var error = (ShopifyError)response;
                     if (error != null) {
-                        callback(null, error);
+                        callback(null, error, null);
                     } else {
-                        callback((List<Product>) response.data.shop().products(), null);
+                        string lastCursor = null;
+
+                        if (response.data.shop().products().pageInfo().hasNextPage()) {
+                            List<ProductEdge> productEdges = response.data.shop().products().edges();
```
sleroux: Is there a chance that productEdges has no elements? Not sure since we do the `hasNextPage` check here. If so we'll crash below.

scripts/generator/graphql_generator/csharp/SDK/Delegates.cs.erb
mikkoh: So we're using this when we're passing a list of ids to load products.

scripts/generator/graphql_generator/csharp/SDK/DefaultProductQueries.cs.erb
```diff
@@ -7,7 +7,7 @@ namespace Shopify.Unity.SDK {
     /// Generates default queries for <see ref="ShopifyClient.products">ShopifyClient.products </see>.
     /// </summary>
     public class DefaultProductQueries {
-        public void ShopProducts(QueryRootQuery query, Dictionary<string,int> imageResolutions, int first = 250, string after = null) {
+        public void ShopProducts(QueryRootQuery query, Dictionary<string,int> imageResolutions, int first = 20, string after = null) {
```
richardmonette: We define `20` as a constant here https://github.com/Shopify/unity-buy-sdk/pull/387/files#diff-4056dfc8033dc677ee07c55e6dfe0bc6R11 which is 👍 but then here we use it as a magic number 👎 Can we reference the constant in all places? (Then if we change it, we only need to make the update in one place, therefore preventing these getting out of sync.)

scripts/generator/graphql_generator/csharp/ShopifyBuy.cs.erb
```diff
@@ -287,18 +288,24 @@ namespace Shopify.Unity {
...
+        /// ShopifyBuy.Client().products((products, error, after) => {
         /// 	Debug.Log(products[0].title());
         /// 	Debug.Log(products[1].title());
         /// 	Debug.Log(products.Count);
         /// });
         /// \endcode
-        public void products(ProductsHandler callback, int? first = null, string after = null) {
-            GetProductsList((products, error) => {
+        public void products(ProductsPaginatedHandler callback, int? first = 20, string after = null) {
+            GetProductsList((products, error, afterProductsList) => {
```
richardmonette: I'm guessing the variable name here is a result of not wanting to shadow the outer `after`. Minor thing, but maybe it should be `afterGetProductsList` which makes it totally clear what function its basically being re-scoped for.

scripts/generator/graphql_generator/csharp/ShopifyBuy.cs.erb
mikkoh: Yeh I guess the only way you could break this is by doing something like `first: 0`

scripts/generator/graphql_generator/csharp/SDK/DefaultProductQueries.cs.erb
mikkoh: It's unfortunate but in C# there's now way to do `#define PAGE_SIZE 20` and you can't have a variable as the default for optional types.

richardmonette:
how pagination works

PR #388, opened 2017-11-22
Update integration tests

mikkoh:
This PR updates integration tests to run using the new unity test tools.

Sorry this effects many files because I deleted the old testing tools.


Assets/IntegrationTests/TestGenericQuery.cs
```diff
@@ -1,25 +1,31 @@
...
-                    IntegrationTest.Assert("graphql" == data.shop().name(), "Shop name was \"graphql\"");
-
-                    IntegrationTest.Pass();
+					hadResult = true;
+					Assert.IsNull(error, "No errors");
+					Assert.AreEqual("graphql", data.shop().name(), "Shop name was \"graphql\"");
                 }
             );
+
+			yield return Utils.GetWaitMaxQueryDuration ();
```
JakeCataford: does this work if you are running in play mode? I know in edit mode you have to return `null` here

JakeCataford: Will this always wait some large amount of time? What is the request ends sooner? Can this GetWaitMaxQueryDuration take the hadResult bool and stop waiting if that is true?

Assets/IntegrationTests/TestShopifyCollections.cs
```diff
@@ -1,45 +1,40 @@
...
     using Shopify.Unity;
+	using System.Collections;
     using System.Collections.Generic;
 
-    [IntegrationTest.DynamicTest ("TestScene")]
-    [IntegrationTest.SucceedWithAssertions]
     public class TestLoad3Collections : MonoBehaviour {
-        void Start() {
+		[UnityTest]
+		public IEnumerator Load3Collections() {
```
JakeCataford: formatting seems to be weird

Assets/IntegrationTests/TestShopifyCollections.cs
```diff
@@ -1,45 +1,40 @@
...
             ShopifyBuy.Init("351c122017d0f2a957d32ae728ad749c", "graphql.myshopify.com");
+			bool hadResult = false;
 
             ShopifyBuy.Client().collections(
-                first: 3,
+                first: 1,
```
JakeCataford: Should this test now be called "loadCollection" if it only loads one collection? Because the new test stack is unit-testey do you plan on following up by combining these tests into one file?

mikkoh:
@JakeCataford all of your concerns should be addressed. I created a class `StoppableWaitForTime` which can be used to either wait a max time or stop when a query is complete.

Assets/IntegrationTests/StoppableWaitForTime.cs
```diff
@@ -0,0 +1,41 @@
...
+            }
+        }
+
+        public float DurationToStop {
+            get {
+                return StopTime - StartTime;
+            }
+        }
+
+        public bool IsStopped {
```
sleroux: Nit: We can avoid having the duplicate private var by declaring this:

```csharp
public bool IsStopped { get; private set; }
```

Assets/IntegrationTests/StoppableWaitForTime.cs
```diff
@@ -0,0 +1,41 @@
...
+            get {
+                return _IsStopped;
+            }
+        }
+
+        float _MaxTime;
+        bool _IsStopped;
+
+        float StopTime;
+        float StartTime;
```
sleroux: Nit: Should these be private?

Assets/IntegrationTests/TestGenericQuery.cs
```diff
@@ -1,25 +1,31 @@
...
+    using System.Collections;
 
-    [IntegrationTest.DynamicTest ("TestScene")]
-    [IntegrationTest.SucceedWithAssertions]
     public class TestGenericQuery : MonoBehaviour {
-        void Start() {
+        [UnityTest]
+        public IEnumerator LoadShopName() {
             ShopifyBuy.Init("351c122017d0f2a957d32ae728ad749c", "graphql.myshopify.com");
+            StoppableWaitForTime waiter = Utils.GetWaitQuery ();
```
sleroux: Nit: Whitespace between () and y

Assets/IntegrationTests/StoppableWaitForTime.cs
mikkoh: Not sure you should be able to modify them because its just doing bad things if you do.

Assets/IntegrationTests/TestGenericQuery.cs
mikkoh: MonoDevelop 😢

Assets/IntegrationTests/StoppableWaitForTime.cs
mikkoh: Gah sorry... Just realize I read your comment wrong. They are private by default.

JakeCataford: they should follow the naming convention for private member variables though (underscore prefix)

Assets/IntegrationTests/TestGenericQuery.cs
```diff
@@ -1,25 +1,31 @@
...
-                    IntegrationTest.Pass();
+                    waiter.Stop();
+                    Assert.IsNull(error, "No errors");
+                    Assert.AreEqual("graphql", data.shop().name(), "Shop name was \"graphql\"");
                 }
             );
+
+            yield return waiter;
+
+            Assert.IsTrue (waiter.IsStopped, Utils.MaxQueryMessage);
```
JakeCataford: if the waiter bottoms out then we could just fail an assertion inside the waiter instead of asserting here?

PR #389, opened 2017-11-23
update webplayer check

richardmonette:
Quick update because `isWebPlayer` is deprecated, per https://docs.unity3d.com/ScriptReference/Application-isWebPlayer.html


mikkoh:
#388 deletes this code entirely.

richardmonette:
Ah, even better!

PR #391, opened 2017-11-24
[UI Toolkit] Add Single Product Theme Controller

JakeCataford:
Adds a controller that loads a single product and notifies a theme on
Show()


Assets/Shopify/UIToolkit/Theme Controllers/Editor/SingleProductThemeControllerEditor.cs
```diff
@@ -0,0 +1,40 @@
...
+                return target as SingleProductThemeController;
+            }
+        }
+
+        public void OnEnable() {
+            BindThemeIfPresent();
+        }
+
+        public override void OnInspectorGUI() {
+            if (Target.Theme == null) {
```
JakeCataford: These things are going to be hard to test... I might create an EditorView layer that is testable and just have the editor call out to it... Right now what this does is pretty trivial though

Assets/Shopify/UIToolkit/Theme Controllers/Editor/SingleProductThemeControllerEditor.cs
```diff
@@ -0,0 +1,40 @@
...
+
+Override SingleProductTheme with your own custom script and add it to this class to continue.
+            ";
+
+            EditorGUILayout.HelpBox(message, MessageType.Warning);
+        }
+
+        private void BindThemeIfPresent() {
+            if (Target.Theme != null) return;
+            Target.Theme = Target.GetComponent<ISingleProductTheme>();
```
JakeCataford: This is something I did not know worked tbh... We could potentially drop the abstract classes because it exists and only have the interfaces... I will do that in a future pr

Assets/Shopify/UIToolkit/Themes/Interfaces/IThemeBase.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Shopify.UIToolkit {
+    using Shopify.Unity.SDK;
+
+    public interface IThemeBase {
+        void OnLoadingStarted();
+        void OnLoadingFinished();
+        void OnError(ShopifyError error);
+        void OnPurchaseStarted();
+        void OnPurchaseCancelled();
+        void OnPurchaseCompleted();
```
JakeCataford: These interfaces make it a lot easier to test the behaviours because we can mock them

Assets/Shopify/UIToolkit/Test/Unit/Editor/TestSingleProductThemeController.cs
```diff
@@ -0,0 +1,109 @@
...
+
+                callbackCalled = true;
+            });
+
+            while (!callbackCalled) {
+               yield return null;
+            }
+        }
+
+        /* TODO: Fix bug where exception is thrown when productGID is valid but not found on shop
```
JakeCataford: Oh yea this throws an error in the json deserializer???? I don't wanna forget about this test though so I'll keep it around as a TODO and prioritize this fix

Assets/Shopify/UIToolkit/Test/Unit/Editor/TestSingleProductThemeController.cs
```diff
@@ -0,0 +1,109 @@
...
+                var product = x.Args()[0] as Product;
+                var productVariants = x.Args()[1] as ProductVariant[];
+
+                Assert.AreEqual("Arena Zip Boot", product.title());
+                Assert.AreEqual(10, productVariants.Length);
+
+                callbackCalled = true;
+            });
+
+            while (!callbackCalled) {
```
JakeCataford: I can use @mikkoh's Wait stuff here instead

Assets/Shopify/UIToolkit/Test/Unit/Editor/TestSingleProductThemeController.cs
sleroux: I wonder if this is a bug in how we deserialize the JSON when we encounter errors in the GraphQL response. I feel like this should return GraphQL errors saying the product wasn't found.

Assets/Shopify/UIToolkit/Theme Controllers/ThemeControllerBase.cs
```diff
@@ -30,5 +30,23 @@ public abstract class ThemeControllerBase : MonoBehaviour {
             }
         }
         private ShopifyClient _cachedClient;
+
+    /// <summary>
```
sleroux: Nit: Whitespace

Assets/Editor/EditorTimeoutSwitch.cs
```diff
@@ -0,0 +1,36 @@
+
+namespace Shopify.Tests {
+    using NUnit.Framework;
+    using UnityEditor;
+    using UnityEngine.TestTools;
+
+    public class EditorTimeoutWaiter {
```
JakeCataford: @mikkoh I built this as a complement to `StoppableWaitForTime`... in the editor you can only yield null from a `UnityTest` so the normal stoppable waiter wont work. WDYT?

Assets/Editor/Utils.cs
```diff
@@ -1,13 +1,17 @@
...
+        public const string TestAccessToken = "351c122017d0f2a957d32ae728ad749c";
+
         public static bool DoesExist(string className) {
             return (from assembly in AppDomain.CurrentDomain.GetAssemblies()
-            from type in assembly.GetTypes()
-            where type.Name == className
-            select type).Any();
+                    from type in assembly.GetTypes()
+                    where type.Name == className
+                    select type).Any();
```
JakeCataford: `gg=g`

Assets/Editor/EditorTimeoutSwitch.cs
mikkoh: Is there any way to know whether we're in play mode or editor mode? This way we could use the same class

JakeCataford: yea with compiler directives, the downside is we cant use `CustomYieldInstruction` so the interface is completely different for the two.

JakeCataford:
@mikkoh or @richardmonette please review this its suddenly green i don't know how

PR #392, opened 2017-11-24
Add pagination to example app

richardmonette:
Updates the example app to use the updated pagination functionality.

Specifically, `FetchProducts` now takes `after`. 

Since the viewing area shows about 4-5 products, the method has been updated to request just 5. This also mean that since the example app has only 13 or so products, we are able to actually view the paging behaviour.

In order to ensure that products aren't added to the list multiple times, a cache of the ids is used. 

To prevent additional queries after all the products have been loaded, the products page tracks when `null` is returned, which indicates there are no more values to follow.

There is also a bit of logic which debounces the tracking of scrolling to the bottom of the view.


Assets/Shopify/Examples/Scripts/Panels/ProductsPanel.cs
```diff
@@ -14,19 +16,30 @@ public class ProductsPanel : MonoBehaviour {
...
         private RectTransform RectTransform;
 
+        private bool HitEndCursor;
+
+        private string After;
+
+        private bool WasScrolledToBottom;
+        private bool IsScrolledToBottom;
+
+        private int ScrollCount;
```
JakeCataford: we've been prefixing private member vars with a `_`

Assets/Shopify/Examples/Scripts/Panels/ProductsPanel.cs
```diff
@@ -47,9 +81,29 @@ public class ProductsPanel : MonoBehaviour {
             foreach (var productView in visibleProductViews) {
                 productView.Load();
             }
+				
+            IsScrolledToBottom = scrollOffset.y < 0;
+        }
+
+        private void Update() {
```
JakeCataford: Why do you need an Update method here? could you just invoke Fetch products from within `OnScrollRectPositionChanged`?

Assets/Shopify/Examples/Scripts/Panels/ProductsPanel.cs
richardmonette: updated!

richardmonette:
@JakeCataford updated, I also added in events on the animation, so the resource fetching is paused when the animation is happening (as during this time scroll events are being triggered)

Assets/Shopify/Examples/Scripts/Panels/ProductsPanel.cs
```diff
@@ -14,42 +16,99 @@ public class ProductsPanel : MonoBehaviour {
...
+            if (!_wasScrolledToBottom && isScrolledToBottom) {
+                if (!_hitEndCursor) {
+                    FetchProducts ();
+                }
+            }
+
+            _wasScrolledToBottom = isScrolledToBottom;
         }
 
         public void AddProduct(Shopify.Unity.Product product) {
```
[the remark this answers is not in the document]
richardmonette: Updated!

PR #393, opened 2017-11-24
Adds collections by id API to ShopifyBuy.Client

sleroux:
* Adds the ability to pass in a list of collection IDs to query for from the server

### DIFF
--- a/Assets/IntegrationTests/TestShopifyCollections.cs
+++ b/Assets/IntegrationTests/TestShopifyCollections.cs
@@ -11,6 +11,13 @@ namespace Shopify.Unity.Tests
     using System.Collections.Generic;
 
     public class TestShopifyCollections : MonoBehaviour {
+        [SetUp]
+        public void Setup() {
+            #if (SHOPIFY_TEST)
+            ShopifyBuy.Reset();
+            #endif
+        }
+
         [UnityTest]
         public IEnumerator LoadACollection() {
             StoppableWaitForTime waiter = Utils.GetWaitQuery ();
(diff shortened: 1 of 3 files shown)


Assets/IntegrationTests/TestShopifyCollections.cs
```diff
@@ -37,6 +38,40 @@ public class TestShopifyCollections : MonoBehaviour {
...
+        public IEnumerator LoadCollectionById() {
+            ShopifyBuy.Init("351c122017d0f2a957d32ae728ad749c", "graphql.myshopify.com");
+            StoppableWaitForTime getFirstCollectionWaiter = Utils.GetWaitQuery ();
+
+            Collection foundCollection = null;
+            ShopifyBuy.Client().collections(
+                first: 1,
+                callback: (collections, error) => {
+                    getFirstCollectionWaiter.Stop();
+                    foundCollection = collections[0];
```
mikkoh: Can we just hardcode the id?

Assets/IntegrationTests/TestShopifyCollections.cs
```diff
@@ -37,6 +38,40 @@ public class TestShopifyCollections : MonoBehaviour {
...
+                }
+            );
+
+            yield return getFirstCollectionWaiter;
+            Assert.IsTrue (getFirstCollectionWaiter.IsStopped, Utils.MaxQueryMessage);
+
+            StoppableWaitForTime getSecondCollectionWaiter = Utils.GetWaitQuery ();
+            var collectionIds = new List<string>() { foundCollection.id() };
```
mikkoh: Can we just hardcode ids?

richardmonette: Since this is also mean to work on a list of IDs, I think it would be good if the test cover 2 or more ids.

scripts/generator/graphql_generator/csharp/ShopifyBuy.cs.erb
```diff
@@ -485,6 +485,60 @@ namespace Shopify.Unity {
             }, first: first, after: after);
         }
 
+        /// <summary>
+        /// Generates a query to fetch the specified <c>collections</c> by id from a Shopify store. The generated query will query the following on collections:
```
richardmonette: nit: can the second sentence be on a new line? (makes it easier to read in narrow displays)

scripts/generator/graphql_generator/csharp/ShopifyBuy.cs.erb
```diff
@@ -485,6 +485,60 @@ namespace Shopify.Unity {
...
+        /// <param name="callback">callback that will receive responses from server</param>
+        /// <param name="first">the list of collection ids you want to receive from the server</param>
+        /// \code
+        /// // Example usage querying two collection ids using a List<string>
+        /// List<string> collectionIds = new List<string>() {
+        ///     "Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0Lzk4OTUyNzYwOTk=",
+        ///     "Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0Lzk4OTUyNzkwNDM="
+        /// };
+        ///
+        /// ShopifyBuy.Client().collections((collections, error) => {
```
mikkoh: This example needs to be updated to match the API

scripts/generator/graphql_generator/csharp/ShopifyBuy.cs.erb
```diff
@@ -485,6 +485,60 @@ namespace Shopify.Unity {
...
+        ///     Debug.Log(collections[0].title());
+        ///     Debug.Log(collections[1].title());
+        /// }, collectionIds);
+        /// \endcode
+        public void collections(CollectionsHandler callback, List<string> collectionIds) {
```
mikkoh: We should also add in a version of this function that looks similar to:
```
public void products (ProductsHandler callback, string firstProductId, params string[] otherProductIds) {
```

PR #394, opened 2017-11-27
Pagination integration tests

mikkoh:
Added some integration tests for pagination. Noticed there was an issue where the same client would be used for two tests which was causing issues where the wrong default client was being used.

### DIFF
--- a/Assets/IntegrationTests/TestShopifyCollections.cs
+++ b/Assets/IntegrationTests/TestShopifyCollections.cs
@@ -69,7 +69,6 @@ namespace Shopify.Unity.Tests
 
             yield return waiter;
 
-            Debug.Log(AllCollections[0].title());
             Assert.IsTrue(waiter.IsStopped, "Query did not complete in " + maxDuration + " seconds");
             Assert.IsNull(errorsFromQueries);
             Assert.GreaterOrEqual(AllCollections.Count, maxCollections);
--- a/Assets/IntegrationTests/TestShopifyCollections.cs
+++ b/Assets/IntegrationTests/TestShopifyCollections.cs
@@ -15,7 +15,7 @@ namespace Shopify.Unity.Tests
 
             ShopifyBuy.Client().collections(
                 first: 1,
-                callback: (collections, error) => {
+                callback: (collections, error, after) => {
                 waiter.Stop();
 
                     Assert.IsNull(error, "No errors");
--- a/Assets/IntegrationTests/TestShopifyCollections.cs
+++ b/Assets/IntegrationTests/TestShopifyCollections.cs
@@ -1,16 +1,18 @@
 namespace Shopify.Unity.Tests
 {
+    using System;
     using UnityEngine;
     using UnityEngine.TestTools;
     using NUnit.Framework;
     using Shopify.Unity;
+    using Shopify.Unity.SDK;
     using System.Collections;
     using System.Collections.Generic;
 
     public class TestShopifyCollections : MonoBehaviour {
         [UnityTest]
         public IEnumerator LoadACollection() {
-            ShopifyBuy.Init("351c122017d0f2a957d32ae728ad749c", "graphql.myshopify.com");
+            ShopifyBuy.Init("43b7fef8bd2f27f1d645586b72c9b825", "graphql-many-products.myshopify.com");
             StoppableWaitForTime waiter = Utils.GetWaitQuery ();
 
             ShopifyBuy.Client().collections(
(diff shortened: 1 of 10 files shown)


scripts/generator/graphql_generator/csharp/SDK/DefaultQueries.cs.erb
```diff
@@ -8,7 +8,7 @@ namespace Shopify.Unity.SDK {
     /// </summary>
     public class DefaultQueries {
         public static readonly int MaxPageSize = 250;
-        public static readonly int MaxProductPageSize = 20;
+        public static readonly int MaxProductPageSize = 10;
```
sleroux: Is this change intended? Just making sure since it wasn't mentioned in the PR description.

mikkoh: Yes it's intended. I think 10 gives us room to grow while still not being a major hit for users. These integration tests really informed those numbers.

Assets/IntegrationTests/TestShopifyCollections.cs
```diff
@@ -36,5 +37,41 @@ public class TestShopifyCollections : MonoBehaviour {
...
             Assert.IsTrue (waiter.IsStopped, Utils.MaxQueryMessage);
         }
+
+        [UnityTest]
+        public IEnumerator LoadAllCollections() {
+            float maxDuration = 6f;
+            int maxCollections = 13;
+            StoppableWaitForTime waiter = new StoppableWaitForTime (maxDuration);
+            ShopifyError errorsFromQueries = null;
+            List<Collection> AllCollections = new List<Collection>();
```
JakeCataford: We should be using type inference here where possible

Assets/IntegrationTests/TestShopifyCollections.cs
```diff
@@ -36,5 +37,41 @@ public class TestShopifyCollections : MonoBehaviour {
...
+
+            Action<string> loadPage = null;
+
+            loadPage = (string after) => {
```
JakeCataford: This can be condensed to `var loadPage = (...) => ...`

Assets/IntegrationTests/TestShopifyCollections.cs
```diff
@@ -36,5 +37,41 @@ public class TestShopifyCollections : MonoBehaviour {
...
+                    if (errors != null) {
+                        errorsFromQueries = errors;
+                        waiter.Stop();
+                    } else if (afterCollectionsLoaded == null || AllCollections.Count >= maxCollections) {
+                        waiter.Stop();
+                    } else {
+                        loadPage(afterCollectionsLoaded);
+                    }
+                }, after: after);
+            };
```
JakeCataford: We could build this into the sdk at some point, the recursive paginator

Assets/IntegrationTests/TestShopifyProducts.cs
```diff
@@ -46,17 +47,16 @@ public class TestShopifyProducts : MonoBehaviour {
 
             yield return waiter;
 
-            Assert.IsTrue (waiter.IsStopped, Utils.MaxQueryMessage);
+            Assert.IsTrue(waiter.IsStopped, Utils.MaxQueryMessage);
```
JakeCataford: This is a comment on shipped code, but I noticed `MaxQueryMessage` is wrong. As per the docs this message is shown when the test fails:

> Verifies that the specified condition is true. The assertion fails if the condition is false. Displays a message if the assertion fails.

So that message should be something like "Query failed to run in x seconds" or whatever instead of "Query finished in 1 seconds"

Assets/IntegrationTests/TestShopifyProducts.cs
```diff
@@ -84,9 +84,45 @@ public class TestShopifyProducts : MonoBehaviour {
                 "Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0Lzk4OTUyODE0NzU=", "Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0Lzk4OTUyODQyOTE="
             );
 
-            yield return Utils.GetWaitQuery ();
+            yield return waiter;
+
+            Assert.IsTrue(waiter.IsStopped, Utils.MaxQueryMessage);
```
JakeCataford: I still think it should be the job of the waiter to fail the test if it times out, which would remove this boilerplate

Assets/IntegrationTests/TestShopifyCollections.cs
mikkoh: Unfortunately it can't because it's a recursive lambda if you don't drop it to two lines it means that `loadPage` doesn't exist at the time of declaration.

Assets/IntegrationTests/TestShopifyCollections.cs
mikkoh: Hmmm... Why's that?

Assets/IntegrationTests/TestShopifyCollections.cs
mikkoh: Yeah I was thinking about that to just return a lambda.

Assets/IntegrationTests/TestShopifyCollections.cs
JakeCataford: Type inference makes the code more concise, especially where it is clear from the declaration what type the resulting variable will be. I was looking at https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions in their implicit local var section

Assets/IntegrationTests/TestShopifyCollections.cs
JakeCataford: ahh, right

Assets/IntegrationTests/TestShopifyCollections.cs
mikkoh: Yeah. In this case it might be cool to do. I just hate using `var` on a function return because I think it makes the code less readable especially considering that this is a public repo.

Assets/IntegrationTests/TestShopifyProducts.cs
mikkoh: Yeah maybe. Let's leave for now and refactor later when there's time.

PR #395, opened 2017-11-28
Wrapped integration tests with UNITY_EDITOR check

sleroux:
Fixes compile errors when building project for iOS platform and running test_iOS.sh

### DIFF
--- a/Assets/IntegrationTests/TestGenericQuery.cs
+++ b/Assets/IntegrationTests/TestGenericQuery.cs
@@ -1,3 +1,4 @@
+#if UNITY_EDITOR
 namespace Shopify.Unity.Tests
 {
     using UnityEngine;
@@ -29,3 +30,4 @@ namespace Shopify.Unity.Tests
         }
     }   
 }
+#endif
--- a/Assets/IntegrationTests/TestShopifyCollections.cs
+++ b/Assets/IntegrationTests/TestShopifyCollections.cs
@@ -1,3 +1,4 @@
+#if UNITY_EDITOR
 namespace Shopify.Unity.Tests
 {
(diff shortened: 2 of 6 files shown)


sleroux:
> A better solution is just to put these tests under an Editor folder. It will accomplish the same thing

Whenever I move the IntegrationTests into an Editor folder or move the tests themselves into an Editor folder they disappear from the UnityTestRunner's Play Mode area :(

JakeCataford:
thats to be expected though because anything in the Editor folder is not available in play mode, its a namespace of sorts thats totally excluded from builds

sleroux:
Whenever I move the Integration Tests into an Editor/ subfolder, they no longer show up as 'Play Mode' tests and show up under 'Edit Mode'. However, the tests fail when run in Edit Mode because you're only allowed to `yield return null` when using `UnityTest`. From what I can tell we have two options:

1. Move the tests into an `Editor` folder, remove the `StoppableWaiter` code and replace it with a `yield return null` + wait for boolean to be set to `true` hack
2. Keep the integration tests outside an editor folder, wrap the code with the `UNITY_EDITOR` directive, and have they only runnable in 'Play Mode'

Thoughts?

mikkoh:
Is there any way to define a folder to ignore from compiles/builds?

JakeCataford:
> Is there any way to define a folder to ignore from compiles/builds? Might be another option that might be more scalable

No way to do this in unity (thats what the editor folder is supposed to do). I think this is because the play mode is supposed to emulate your game build as closely as possible. It is possible to make these tests run in the editor, but it requires making StoppableWaitForTime work via the parent function yielding null. Lets go with your og solution.

PR #398, opened 2017-11-29
[WIP] Refactored WebImageCache into DataCache interface and LRUCache abstract class

sleroux:
This PR abstracts out a DataCache interface and LRUCache abstract class from WebImageCache. This will open us to using the LRU cache heuristic for request caching if we want and also gives us an interface for defining other types of caches with.
WebImageCache is implemented using the refactored abstraction and uses the same test cases.

### DIFF
--- a/Assets/Shopify/UIToolkit/Components/DataCache.cs
+++ b/Assets/Shopify/UIToolkit/Components/DataCache.cs
@@ -1,0 +1,60 @@
+ namespace Shopify.UIToolkit {
+    using UnityEngine;
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// An interface for defining a data cache.
+    /// </summary>
+    public interface DataCache<T> {
+        /// <summary>
+        /// The estimated size limit in bytes of how much memory we want to use for the cache (readonly).
+        /// </summary>
+        /// <returns></returns>
+        int MemorySizeLimit { get; }
+
+        /// <summary>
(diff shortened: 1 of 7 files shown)


sleroux:
Based off of https://github.com/Shopify/unity-buy-sdk/pull/384.

Assets/Shopify/UIToolkit/Components/DataCache.cs
```diff
@@ -0,0 +1,60 @@
+ namespace Shopify.UIToolkit {
+    using UnityEngine;
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// An interface for defining a data cache.
+    /// </summary>
+    public interface DataCache<T> {
```
JakeCataford: Kinda nitpicky.... but what kind of cache stores something other than data? lol

Assets/Shopify/UIToolkit/Components/LRUCache.cs
```diff
@@ -0,0 +1,136 @@
+ namespace Shopify.UIToolkit {
```
JakeCataford: These are nice tools, we might wanna consider moving them to a more general namespace at some point

Assets/Shopify/UIToolkit/Components/DataCache.cs
sleroux: Heh true enough. I guess I was concerned about naming collisions but since we're within our namespace I don't think we'll have any issues.

PR #400, opened 2017-11-29
[UI Toolkit] Editor Utility for showing a product picker in the editor

JakeCataford:
Implementation of a tool that is to be called within an editor script
that allows the loading and rendering of a product picker within the
editor.

Follows a view/controller pattern that allows us to mock EditorGUI calls
out for tests.

### DIFF
--- a/Assets/Shopify/UIToolkit/Editor/ProductPicker.cs
+++ b/Assets/Shopify/UIToolkit/Editor/ProductPicker.cs
@@ -1,0 +1,155 @@
+﻿namespace Shopify.UIToolkit.Editor {
+    using System.Collections;
+    using System.Collections.Generic;
+    using UnityEngine;
+    using UnityEditor;
+    using Shopify.Unity;
+    using Shopify.Unity.SDK;
+    using System;
+    using System.Linq;
+
+    public interface IProductPickerView {
+        void DrawProductLoadingState(SerializedObject serializedObject);
+        void DrawProductPicker(SerializedObject serializedObject);
+        void DrawProductLoadingError(SerializedObject serializedObject);
+        void DrawShopHasNoProductsState(SerializedObject serializedObject);
+    }
+
+    public class ProductPicker : IProductPickerView {
+        private class ProductNameToGIDMapEntry {
+            public string ProductName;
+            public string ProductGID;
+        }
+
+        public IProductPickerView View;
+        private static List<ProductNameToGIDMapEntry> _cachedProductNameToGIDMap;
+        private List<ProductNameToGIDMapEntry> _productNameToGIDMap;
+        private ShopifyClient _client;
+        private ShopifyError _error;
+
+        public bool ProductsFinishedLoading() {
+            return _productNameToGIDMap != null || _error != null;
+        }
+
+        public ProductPicker(ShopifyClient client) {
+            _client = client;
+            View = this;
+            LoadProducts();
+        }
+
+        public void DrawInspectorGUI(SerializedObject serializedObject) {
+            if (_error != null) {
+                View.DrawProductLoadingError(serializedObject);
+                return;
+            }
+
+            if (!ProductsFinishedLoading()) {
+                View.DrawProductLoadingState(serializedObject);
+                return;
+            }
+
+            if (_productNameToGIDMap.Count == 0) {
+                View.DrawShopHasNoProductsState(serializedObject);
+                return;
+            }
+
+            View.DrawProductPicker(serializedObject);
+        }
+
+        public void LoadProducts() {
+            if (_cachedProductNameToGIDMap != null) {
+                _productNameToGIDMap = _cachedProductNameToGIDMap;
+            }
+
+            _client.products((products, error) => {
+                if (error != null) {
+                    OnFailedToLoadProducts(error);
+                    return;
+                }
+
+                OnLoadedProducts(products);
+            });
+        }
+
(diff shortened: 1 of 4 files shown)


Assets/Shopify/UIToolkit/Editor/ProductPicker.cs
```diff
@@ -0,0 +1,155 @@
...
+                View.DrawProductLoadingError(serializedObject);
+                return;
+            }
+
+            if (!ProductsFinishedLoading()) {
+                View.DrawProductLoadingState(serializedObject);
+                return;
+            }
+
+            if (_productNameToGIDMap.Count == 0) {
```
sleroux: I wonder if it this boolean checking logic would be more solid if it was implementing using a simple state machine with EMPTY, LOADING, HAS_PRODUCTS states or something. It's probably overkill for this though.

Assets/Shopify/UIToolkit/Editor/ProductPicker.cs
```diff
@@ -0,0 +1,155 @@
...
+        public void RefreshProducts() {
+            _error = null;
+            _productNameToGIDMap = null;
+            _cachedProductNameToGIDMap = null;
+            LoadProducts();
+        }
+
+        private void OnFailedToLoadProducts(ShopifyError error) {
+            _error = error;
+            _cachedProductNameToGIDMap = null;
```
sleroux: Just because we failed doesn't mean we should invalidate previously cached products no?

Assets/Shopify/UIToolkit/Editor/ProductPicker.cs
```diff
@@ -0,0 +1,155 @@
...
+
+    public class ProductPicker : IProductPickerView {
+        private class ProductNameToGIDMapEntry {
+            public string ProductName;
+            public string ProductGID;
+        }
+
+        public IProductPickerView View;
+        private static List<ProductNameToGIDMapEntry> _cachedProductNameToGIDMap;
+        private List<ProductNameToGIDMapEntry> _productNameToGIDMap;
```
sleroux: What's the benefit of having two lists here instead of just one?

Assets/Shopify/UIToolkit/Editor/ProductPicker.cs
```diff
@@ -0,0 +1,155 @@
...
+        private ShopifyClient _client;
+        private ShopifyError _error;
+
+        public bool ProductsFinishedLoading() {
+            return _productNameToGIDMap != null || _error != null;
+        }
+
+        public ProductPicker(ShopifyClient client) {
```
sleroux: Forgot to mention - This feels a bit weird, assigning this to the View property here. Why not extract it into a ProductPickerView object, passing in a IProductPickerView instance that has a default value of a new ProductPickerView instance:

```csharp
public ProductPicker(ShopifyClient client, IProductPickerView view = new ProductPickerView()) {
    // where the magic happens
}
```

Assets/Shopify/UIToolkit/Editor/ProductPicker.cs
```diff
@@ -0,0 +1,155 @@
...
+
+    public interface IProductPickerView {
+        void DrawProductLoadingState(SerializedObject serializedObject);
+        void DrawProductPicker(SerializedObject serializedObject);
+        void DrawProductLoadingError(SerializedObject serializedObject);
+        void DrawShopHasNoProductsState(SerializedObject serializedObject);
+    }
+
+    public class ProductPicker : IProductPickerView {
+        private class ProductNameToGIDMapEntry {
```
sleroux: Nit: Does it ever make sense to have a null `ProductNameToGIDMapEntry`? Might make more sense to use value types here by declaring it as a `struct` instead of `class`.

Assets/Shopify/UIToolkit/Editor/ProductPicker.cs
```diff
@@ -0,0 +1,168 @@
...
+                _view.DrawShopHasNoProductsState(serializedObject);
+                return;
+            }
+
+            _view.DrawProductPicker(serializedObject, _productNameToGIDMap);
+        }
+
+        public void LoadProducts() {
+            if (_cachedProductNameToGIDMap != null) {
+                _productNameToGIDMap = _cachedProductNameToGIDMap;
```
richardmonette: What is this cache for? We seem to set it in a few places, but I don't see it being read anywhere..?

JakeCataford: its being read right here :) its written to after the products have loaded once. The reason for this is we get a new editor instance everytime that we select the object, so this cache will only be cleared after a recompilation instead of the product list reloading every time we look at the object. Its used mostly for an optimistic load, where we are showing the last results while still fetching in the background to see if anything changed

richardmonette: Doesn't `_productNameToGIDMap =` get called at https://github.com/Shopify/unity-buy-sdk/pull/400/files#diff-181c13ea5c2cfc8a4f52c91322bb3d54R102 anyways, once the recursive page load finishes?

JakeCataford: yep, the idea is that while the page is loading you can still browse the previous results. Especially cause 99% of the time those results will be the same as the results you have.

PR #403, opened 2017-11-29
Adds UseCache flag to RemoteImageLoader

sleroux:
* Adds boolean flag `UseCache` to `RemoteImageLoader` to cache
downloaded images into the shared WebImageCache instance.

* Images downloaded are cached alongside their "Last Modified" timestamp
to prevent re-downloads of cached data and cache invalidation when they
are updated on the API side.

### DIFF
--- a/Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
+++ b/Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
@@ -118,7 +118,7 @@ namespace Shopify.UIToolkit {
             var responseHeaders = www.responseHeaders;
             if (responseHeaders.ContainsKey("STATUS")) {
                 string statusCodeLine = responseHeaders["STATUS"];
-                if (ParseResponseCode(statusCodeLine) == 304) {
+                if (HTTPUtils.ParseResponseCode(statusCodeLine) == 304) {
                     if (cachedResource != null) {
                         completion(cachedResource.texture, null);
                     } else {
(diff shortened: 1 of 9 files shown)


sleroux:
Based off of https://github.com/Shopify/unity-buy-sdk/pull/398.

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -10,10 +10,18 @@ namespace Shopify.UIToolkit {
     /// </summary>
     [RequireComponent(typeof(Image))]
     public class RemoteImageLoader : MonoBehaviour {
+
+        /// <summary>
+        /// When true, downloaded images will be cached.
+        /// </summary>
+        public bool UseCache;
```
JakeCataford: Is there a situation where caching is not an overall win? at the very least this should default to true imo.

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -83,6 +95,68 @@ public class RemoteImageLoader : MonoBehaviour {
...
             completion(www.texture, null);
         }
+
+        /// <summary>
+        /// Loads an image from the provided URL if we don't have one in the cache or if it's new. Results are cached
+        /// so subsequent images are not downloaded from the web.
+        /// </summary>
+        /// <param name="url">URL to fetch image from.</param>
+        /// <param name="completion">Callback invoked with the cached or downloaded texture.</param>
+        /// <returns></returns>
```
JakeCataford: imo you dont need docs on an implementation detail, its not a big deal tho... Up to you!

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
sleroux: Setting the default to true makes sense 👍

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -83,6 +95,61 @@ public class RemoteImageLoader : MonoBehaviour {
...
+            _imageCache.SetTextureForURL(url, texture, lastModifiedString);
+            completion(texture, null);
+        }
+
+        /// <summary>
+        /// Parses the 'Status XXX' HTTP header field for the status code number.
+        /// </summary>
+        /// <param name="statusLine">HTTP STATUS field</param>
+        /// <returns>HTTP Status code</returns>
+        private static int ParseResponseCode(string statusLine) {
```
richardmonette: Is this a bit brittle? Maybe we should try and regex this out?

sleroux:
I've replaced the status code parsing logic with a regex and moved it into a utils class so I can unit test it 👌

PR #406, opened 2017-11-30
Allow injectable LoaderProvider for changing loader

JakeCataford:
Allows configuration of the loader used in the theme controllers via a
loader provider that instantiates a loader given the access token and
shop domain. This allows tests to use mock loaders while maintaining the
functionality of the client lazy load + cache. It also reworks some of
the cache invalidation in the controller to simplify the workflow, as
changing the loader now also invalidates the cached client.


scripts/generator/graphql_generator/csharp/SDK/UnityLoader.cs.erb
```diff
@@ -78,5 +78,11 @@ namespace Shopify.Unity.SDK {
             return "unity";
         }
     }
+
+    public class UnityLoaderProvider : ILoaderProvider {
+        BaseLoader ILoaderProvider.GetLoader(string accessToken, string domain) {
```
sleroux: I haven't seen this before - prefixing the method name with the interface type. What's the reason for doing this vs just implementing `GetLoader`?

JakeCataford: Its called explicit interface implementation, the neat thing about it is that the explicit interface methods are only available when the class is cast to the interface type, which ensures that we are using the least specific reference to the type. It also has the added niceness of knowing what methods are satisfying a particular interface at a glance.

Assets/Shopify/UIToolkit/Theme Controllers/ThemeControllerBase.cs
```diff
@@ -1,44 +1,72 @@
...
         /// The client that the controller is using to make requests against the SDK.
         /// </summary>
         /// <returns></returns>
         public ShopifyClient Client {
             get {
-                if (_cachedClient != null &&
-                    _cachedClient.AccessToken == AccessToken && 
-                    _cachedClient.Domain == ShopDomain) {
-                        return _cachedClient;
+                if (_cachedClient == null) {
```
richardmonette: This is kind of cool because it defers the creation until someone tries to use the Client but I was wondering, when invalidating the client, could we not just regenerate it then?

JakeCataford: yea we could for sure... can't think of a reason why one approach is better than the other though

PR #408, opened 2017-11-30
Freeze non schema files

JakeCataford:
Freezes all the files that were erb and didn't really have much generated in them.

Specifically, the erb that was replaced was the QueryRoot name, plus the MutationRoot name (which I assume should never change).

Now there is only one generated non-schema file called VersionInformation which contains the publish destination and the version string.

This also bumps generated things under the "Generated" folder for better visibility... but it might be advisable to flatten this structure when publishing a unitypackage.


JakeCataford:
I formatted the Test expectations lol thats why the tests are failing

PR #410, opened 2017-11-30
Add Shop Domain/Access Token verification UI to ThemeControllers

sleroux:
* Adds a new `BaseThemeControllerEditor` editor script that other ThemeControllerEditors can inherit from to render/perform credential validation.
* Changes IShopCredentials to use getter methods to allow MonoBehaviour field types to work with Editor.
* Added credential rendering/validation to `SingleProductThemeControllerEditor`.
* Fixes exception issue when passing blank domain to verifier.


Assets/Shopify/UIToolkit/Editor/ShopCredentialsVerifier.cs
```diff
@@ -124,8 +131,12 @@ public class ShopCredentialsVerifier {
             }
         }
 
-        private ShopifyClient Client() {
-            return new ShopifyClient(new UnityEditorLoader(_Credentials.ShopDomain, _Credentials.AccessToken));
+        private ShopifyClient MakeClient() {
```
sleroux: If we pass in a blank domain, when we go to contruct a URL inside UnityLoader it blows up. Before continuing we check to see if it was empty/blank string and return null otherwise so I've renamed this method to `MakeClient`.

Assets/Shopify/UIToolkit/Interfaces/IShopCredentials.cs
```diff
@@ -1,7 +1,8 @@
 ﻿namespace Shopify.UIToolkit {
     public interface IShopCredentials {
         ShopCredentialsVerificationState CredentialsVerificationState { get; set; }
-        string ShopDomain { get; }
-        string AccessToken { get; }
+
+        string GetShopDomain();
```
sleroux: I didn't really want to do this but when I defined the properties to match the interface like below:

```csharp
public string AccessToken; // Unity likes this
public string AccessToken { get; set; } // No bueno
```

The `PropertyField` on the serializeObject didn't like that because I guess Unity treats the second case not as a field.

sleroux:
Some snaps for what this looks like:

screen shot 2017-11-30 at 4 19 08 pm
screen shot 2017-11-30 at 4 19 49 pm

Assets/Shopify/UIToolkit/Editor/ShopCredentialsVerifier.cs
richardmonette: This only fixes the issue here, should this check be inside the ShopifyClient constructor, and raise an exception if the constructor arguments are not valid?

sleroux: Definitely. I've updated it so the constructor throws and we catch in the `ShopCredentialsVerifier` if it's invalid and update our state accordingly.

Assets/Shopify/UIToolkit/Interfaces/IShopCredentials.cs
JakeCataford: yea this is true... Good solution!

scripts/generator/graphql_generator/csharp/SDK/Editor/UnityEditorLoader.cs.erb
```diff
@@ -6,12 +6,17 @@ namespace Shopify.Unity.SDK.Editor {
...
 
     /// <summary>
     /// A Query Loader built to load queries in editor scripts, when the normal game loop is not available.
     /// </summary>
     public class UnityEditorLoader : BaseLoader {
         public UnityEditorLoader(string domain, string accessToken) : base(domain, accessToken) {
+            if (domain.Trim().Length == 0) {
+                throw new ArgumentException("Domain is invalid. Make sure that it is not empty/blank.");
+            }
+
```
JakeCataford: nothing should be erb anymore except for the schema files

PR #413, opened 2017-12-05
Variant/Selection Option Lookups

sleroux:
ProductVariants on a product give us a set of selectable options but doesn't do a good job of seperating those options. For some of the UI we're looking to build for the UI Toolkit, we'll want to be able to select options independently of options. For example, select the Color of the shirt, then select the Size. This PR adds a class named `VariantOptionLookup` that takes in the variants on Product and builds an initial list of SelectionOptions to choose from. Once a user selects an option, you can pass this option back into the lookup class to get a set of the next selectable options.

### DIFF
--- a/Assets/Shopify/UIToolkit/Components/VariantSelector.cs
+++ b/Assets/Shopify/UIToolkit/Components/VariantSelector.cs
@@ -2,7 +2,6 @@ namespace Shopify.UIToolkit {
     using System;
     using System.Collections;
     using System.Collections.Generic;
-    using System.Collections.Specialized;
     using System.Linq;
     using Shopify.Unity;
 
--- a/Assets/Shopify/UIToolkit/Components/VariantSelector.cs
+++ b/Assets/Shopify/UIToolkit/Components/VariantSelector.cs
@@ -1,0 +1,101 @@
+namespace Shopify.UIToolkit {
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Collections.Specialized;
+    using System.Linq;
+    using Shopify.Unity;
+
+    using UnityEngine;
+
+    /// <summary>
+    /// Converts the list of product variant and their respective selectable options that looks like:
+    /// [{
+    ///    "selectedOptions": [{
+    ///       "name": "optionName1",
+    ///       "value": "value1"
+    ///     }, {
+    ///       "name": "optionName2",
+    ///       "value": "value1"
+    ///     }]
+    /// },
+    /// {
+    ///    "selectedOptions": [{
+    ///       "name": "optionName1",
+    ///       "value": "value2"
+    ///     }, {
+    ///       "name": "optionName2",
+    ///       "value": "value2"
+    ///     }]
+    /// }]
+    /// into a map that looks like:
+    /// {
+    ///   "optionName1": ["value1", "value2", ...],
+    ///   "optionName2": ["value1", "value2", ...],
+    /// }
+    ///
+    /// </summary>
+    public class VariantSelector {
+        private const string SingleVariantTitle = "Default Title";
+
+        private List<ProductVariant> _variants;
+
+        public VariantSelector(List<ProductVariant> variants) {
+            _variants = variants;
+        }
+
+        public Dictionary<string, List<string>> AllOptions() {
+            return ValidOptionsForSelections(new Dictionary<string, string>());
+        }
+
+        public Dictionary<string, List<string>> ValidOptionsForSelections(Dictionary<string, string> selections) {
+            if (IsSingleOrEmpty()) {
+                return null;
+            }
+
+            var allOptions = new Dictionary<string, List<string>>();
+            foreach (var variant in _variants) {
+                var options = variant.selectedOptions();
+
+                var containsSelections = selections.Aggregate(true, (accum, entry) => {
+                    return accum && options.Find(o => {
+                        return o.name().Equals(entry.Key) && o.value().Equals(entry.Value);
+                    }) != null;
+                });
+
+                // Skip this variant if it's not valid with the selections.
+                if (containsSelections) {
+                    options.Aggregate(allOptions, (accum, option) => {
+                        List<string> values;
+                        if (!accum.ContainsKey(option.name())) {
+                            accum.Add(option.name(), new List<string>());
+                        }
+                        values = accum[option.name()];
+
+                        // Avoid duplicates.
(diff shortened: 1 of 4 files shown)


Assets/Shopify/UIToolkit/Components/VariantOptionLookup.cs
```diff
@@ -0,0 +1,132 @@
...
+            AddInitialOptions(variants);
+        }
+
+        /// <summary>
+        /// Adds the given list of variants into our lookup table and adds the initial 
+        /// selection options to the initial list.
+        /// </summary>
+        /// <param name="moreVariants">An additional set of product variants.</param>
+        public void AddPageOfVariants(List<ProductVariant> moreVariants) {
+            foreach (var variant in moreVariants) {
```
richardmonette: is `.AddRange` an option here?

Assets/Shopify/UIToolkit/Components/VariantOptionLookup.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// <param name="moreVariants">An additional set of product variants.</param>
+        public void AddPageOfVariants(List<ProductVariant> moreVariants) {
+            _variants.AddRange(moreVariants);
+            AddInitialOptions(moreVariants);
+        }
+
+        /// <summary>
+        /// Returns the next set of selectable options given a selected option. For example,
+        /// if a user selects a Color option and the product supports selecting both a color 
+        /// and a size, this API will return a list of all available sizes to choose from.
```
JakeCataford: what if there are more than 2 selected options?

richardmonette: I was wondering about this too. There seems to be some assumption that this is called each time a selector is used, and the list of further available options is updated in the subsequent options selection fields. That seems to break down though if the options are selected out of order.

sleroux: Just to close the loop on this, I'm going to rewrite this to take in a list of already-selected options and return a filtered list of all remaining available options. For example, when passing in nothing, it will return all of the available selections for each option. After selecting regardless of order it will produce the available selections for the remaining options. For example,

sleroux:
@JakeCataford / @richardmonette I simplified the logic a lot while making it much more flexible at the same time. Instead of having to select options in order, you can select whatever you want and the `VariantSelector` will generate a new set of options that are valid for that selection. This will make interacting with the dropdown UI much easier. Could you guys do a re-review? 🙏

Assets/Shopify/UIToolkit/Components/VariantSelector.cs
```diff
@@ -0,0 +1,101 @@
...
+                return allOptions;
+        }
+
+        private bool IsSingleOrEmpty() {
+            if (_variants.Count == 0)  {
+                return true;
+            }
+
+            // This is what the Storefront API returns back to us when there are no variants...
+            if (_variants.Count == 1 && _variants[0].title() == "Default Title") {
```
mikkoh: Should we put `"Default Title"` in a constant just to draw it out if this changes in the future?

Assets/Shopify/UIToolkit/Components/VariantSelector.cs
```diff
@@ -0,0 +1,101 @@
...
+                    var options = variant.selectedOptions();
+
+                    var containsSelections = selections.Aggregate(true, (accum, entry) => {
+                        return accum && options.Find(o => {
+                            return o.name().Equals(entry.Key) && o.value().Equals(entry.Value);
+                        }) != null;
+                    });
+
+                    // Skip this variant if it's not valid with the selections.
+                    if (!containsSelections) continue;
```
mikkoh: Can we get rid of the `continue` and just change to `if (containsSelections)`

Assets/Shopify/UIToolkit/Components/VariantSelector.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        public VariantSelector(List<ProductVariant> variants) {
+            _variants = variants;
+        }
+
+        public Dictionary<string, List<string>> AllOptions() {
+            return ValidOptionsForSelections(new Dictionary<string, string>());
+        }
+
+        public Dictionary<string, List<string>> ValidOptionsForSelections(Dictionary<string, string> selections) {
```
mikkoh: The contents of this whole function are double indented.

Assets/Shopify/UIToolkit/Components/VariantSelector.cs
```diff
@@ -0,0 +1,101 @@
...
+                            return o.name().Equals(entry.Key) && o.value().Equals(entry.Value);
+                        }) != null;
+                    });
+
+                    // Skip this variant if it's not valid with the selections.
+                    if (!containsSelections) continue;
+
+                    options.Aggregate(allOptions, (accum, option) => {
+                        List<string> values;
+                        if (accum.ContainsKey(option.name())) {
```
mikkoh: This can just be `if (!accum.ContainsKey(option.name())) {` then outside of the if statement do `values = accum[option.name()];`

Assets/Shopify/UIToolkit/Components/VariantSelector.cs
sleroux: Thats.. weird. I'll fix!

sleroux:
@mikkoh Pushed review fixes.

> What is this class used for exactly?
 
The storefront API returns a product variant's list of options in a pretty loose format

```json
{
   "name": "Size",
   "value": "Medium"
},
{
   "name": "Color",
   "value": "Blue"
}
```

This is fine for the example app since we display variants by a concatenating the values into a string:

`Medium / Blue`,
`Small / Blue`

For the toolkit, we want to be able to select per an option but this data structure makes it difficult to display and alter that. `VariantSelector` allows you to 'select' a single option type (ie. Color or Size) independently of other options and provide set of valid options for the remaining selections.

Assets/Shopify/UIToolkit/Components/VariantSelector.cs
```diff
@@ -0,0 +1,101 @@
+namespace Shopify.UIToolkit {
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Collections.Specialized;
```
mikkoh: I'm not sure any classes are being used from here.

PR #420, opened 2017-12-13
Publish unitypackage from the editor

JakeCataford:
Includes configurations so that we only package directories included in
the whitelist, as well as the ability to package files from the editor.


scripts/publish.sh
```diff
@@ -86,8 +86,9 @@ printf "Exporting the unitypackage at: %s\n" "$UNITY_PACKAGE"
     -logFile "$UNITY_LOG_PATH" \
     -importPackage "ThirdParty/play-services-resolver-1.2.54.0.unitypackage" \
     -projectPath "$PROJECT_ROOT" \
-    -exportPackage Assets/Shopify Assets/PlayServicesResolver "$UNITY_PACKAGE" \
-    -quit
+    -executeMethod "Publishing.PublishFromCommandLine" \
```
JakeCataford: This is some cool secret sauce... we could probably use this to simplify a lot of workflows in the app

PR #422, opened 2017-12-13
[UI Toolkit] Should Be Able To See Details For a Particular Product

sleroux:
- [x] Description, price and variant selector
- [x] Action to add item to cart
- [x] Browsing Product Images
- [x] No images state
- [ ] Show loading state while controller fetches product
- [ ] Add integration test cases


sleroux:
@JakeCataford I've pushed up what I've done so far for the dropdown stuff. Along the way I tidied up some of the alignments/layout stuff in the Generic theme to make it a but more robust. You can try out what I've done by loading the Assets/Working.scene. The dropdown is functioning but is just using static values for the time being.

Assets/Shopify/UIToolkit/Components/Dropdown/OptionDropdown.cs
```diff
@@ -0,0 +1,111 @@
...
+        public void Build() {
+            TitleLabel.GetComponent<Text>().text = Title;
+
+            // Configure the dropdown to contain options.
+            for (var i = 0; i < Options.Length; i++) {
+                var option = Options[i];
+                AddOptionNamed(option, i == SelectedIndex);
+            } 
+
+            // Resize the scroll view content view to fit the options snug.
```
JakeCataford: theres also content size fitter which would do this

Assets/Shopify/UIToolkit/Components/Dropdown/OptionDropdown.cs
```diff
@@ -0,0 +1,111 @@
...
+            var previousSelection = ScrollViewContent.transform.GetChild(SelectedIndex);
+
+            // Don't bother reselecting a the previous selection!
+            if (newSelection == previousSelection.gameObject) {
+                return;
+            }
+
+            // Deselect previous selection and new selection containers.
+            foreach (Transform child in previousSelection.transform) {
+                GameObject.Destroy(child.gameObject);
```
JakeCataford: you sure you want to destroy here? The GC will have to cleanup... we can also recycle these views but if you opt for that i think doing it in another pr would be best

sleroux: Ideally we'd do recycling but for now I figured we'd just throw it away. I don't think we'll have too many dropdown items anyways but I can profile it just in case.

A good follow up PR would be to add in list recycling that we can also use for product listings.

@richardmonette Do we already have code that does this as part of the example app?

richardmonette: In the example app right now we combine all the options (there is just one selector with all the combinations). It could be upgraded to this though.

sleroux:
@richardmonette / @JakeCataford I've updated this branch to have a working selection picker for the variants that loosely match with Evan's designs here: https://shopify.invisionapp.com/share/ZSEK33M83#/screens/264963812. I originally wanted to go the route of having each option be selected independent from each other but the current designs don't do a good job of representing that since the labels for each option no longer show once you've selected an option. For now I've gone with having the whole variant be an option in the dropdown. We can come back later and add fancier selections if we need to.

I've also pulled in Jake's product picker branch so it's easier to play around with seeing a product render in the generic single product theme.

Assets/Shopify/UIToolkit/Components/SelectionPicker.cs
```diff
@@ -0,0 +1,120 @@
...
+        public string Title;
+
+        [HideInInspector]
+        public UnityEvent onSelectOption = new UnityEvent();
+
+        // Game object references to the prefab'ed UI.
+        public GameObject TitleLabel;
+        public GameObject ScrollViewContent;
+        public GameObject SelectedOptionPrefab;
+        public GameObject UnselectedOptionPrefab;
```
JakeCataford: These can be references directly to the components

ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 2017.1.0f3
+m_EditorVersion: 2017.1.1f1
```
JakeCataford: dont think you wanna change this

Assets/Shopify/UIToolkit/Components/SelectionPicker.cs
```diff
@@ -0,0 +1,120 @@
...
+    /// </summary>
+    public class SelectionPicker : MonoBehaviour {
+        [HideInInspector] 
+        public ISelectionPickerItem[] Items;
+
+        [HideInInspector]
+        public string Title;
+
+        [HideInInspector]
+        public UnityEvent onSelectOption = new UnityEvent();
```
JakeCataford: `OnSelectOption`?

Assets/Shopify/UIToolkit/Components/SelectionPicker.cs
```diff
@@ -0,0 +1,120 @@
...
+    /// Interface for items that are used in the selection picker.
+    /// </summary>
+    public interface ISelectionPickerItem {
+        string Title { get; }
+    }
+
+    /// <summary>
+    /// A custom selection picker implementation.
+    /// </summary>
+    public class SelectionPicker : MonoBehaviour {
```
JakeCataford: Can this be accomplished by leveraging https://docs.unity3d.com/Manual/script-Dropdown.html?

Assets/Shopify/UIToolkit/Editor/ShopCredentialsVerifier.cs
```diff
@@ -34,6 +34,7 @@ public class ShopCredentialsView : IShopCredentialsView {
             DrawMessageBox();
             DrawActionButton();
             EditorGUILayout.Separator();
+            serializedObject.ApplyModifiedProperties();
```
JakeCataford: I was thinking that this is handled by the editor itself...

sleroux:
I've renamed this PR to match with the issues on the UI Toolkit issue board https://github.com/Shopify/gaming-projects/issues/9.

This PR includes most of the functionality we want for that issue with the exception of some better loading states and controller callback cleanup.

Some notes:

* Removed the `WIP` scene file as it wasn't supposed to be in `master` anyways. It was just a temporary scene I was working in for the multi product shop.
* Removes the custom dropdown code in favor of just using the Unity dropdown component and it works pretty well (with less code!).
* Addresses @JakeCataford's comments from the previous review and tidies up the properties.

After the controller refactor and loading UI lands Ill create a new PR to add the states in and some tests.

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -55,7 +55,7 @@ public class RemoteImageLoader : MonoBehaviour {
                 if (texture == null) {
                     Debug.LogWarning("Failed to generate texture from image located at " + imageURL);
                     if (failure != null) {
-                        failure(error);
+                        failure("Texture not found");
```
sleroux: Error is null here since we don't actually get an error for this so use our own string.

JakeCataford:
Might wanna rebase this sooner rather than later?

JakeCataford:
🎩  looks good, but why does the button say "add to cart" in the single product prefab?

sleroux:
> 🎩 looks good, but why does the button say "add to cart" in the single product prefab?

I guess it depends on what we want the button to do. We had it as adding the product to a cart so having it say `Buy Now` seems weird. Do we want to add to cart and check out immediately instead?

JakeCataford:
@sleroux yea thats what we should do

Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericSingleProductShop.cs
```diff
@@ -1,44 +1,194 @@
...
+            _controller = GetComponent<SingleProductShopController>();
+            _controller.Shop = this;
+            _controller.OnShow();
+        }
+
+        #endregion
+
+        #region IShop Interface
+
+        void IShop.OnLoadingStarted() {}
```
richardmonette: Should these not just be left empty so the exceptions happen?

sleroux: I've put the exceptions back in

JakeCataford:
@sleroux good to go on this one?

PR #425, opened 2017-12-14
[UI Toolkit] Refactors SingleProduct editor into BaseEditor

JakeCataford:
This commit refactors the SingleProductThemeControllerEditor into a
BaseThemeControllerEditor. It also modifies the single product tc editor
to provide the product picker when the credentials of the base class are
valid and writes new tests to cover the feature.


Assets/Shopify/UIToolkit/Editor/ProductPicker.cs
```diff
@@ -20,10 +20,10 @@ public interface IProductPickerController {
     }
 
     public class ProductPicker : IProductPickerController {
+        public ShopifyClient Client;
```
JakeCataford: Made this public so we can change it if it updates

Assets/Shopify/UIToolkit/Editor/ShopCredentialsVerifier.cs
```diff
@@ -37,8 +37,8 @@ public class ShopCredentialsView : IShopCredentialsView {
         }
 
         private void DrawCredentialsForm(SerializedObject serializedObject) {
-            EditorGUILayout.PropertyField(serializedObject.FindProperty("ShopDomain"));
-            EditorGUILayout.PropertyField(serializedObject.FindProperty("AccessToken"));
+            EditorGUILayout.PropertyField(serializedObject.FindProperty("_shopDomain"));
+            EditorGUILayout.PropertyField(serializedObject.FindProperty("_accessToken"));
```
JakeCataford: Im a magician and now I can access private vars in the serialized object lol

Assets/Shopify/UIToolkit/Editor/ShopCredentialsVerifier.cs
```diff
@@ -73,6 +73,9 @@ public class ShopCredentialsView : IShopCredentialsView {
     /// and drawing a UI to support the feature.
     /// </summary>
     public class ShopCredentialsVerifier {
+        public delegate void OnCredentialsStateChangedHandler();
+        public event OnCredentialsStateChangedHandler OnCredentialsStateChanged;
```
JakeCataford: This event is used to update the theme controllers client when the credentials change

Assets/Shopify/UIToolkit/Test/Unit/Editor/TestBaseThemeControllerEditor.cs
```diff
@@ -0,0 +1,56 @@
...
+    using Shopify.Unity.SDK;
+
+    [TestFixture]
+    public class TestBaseThemeControllerEditor {
+        private BaseThemeControllerEditor _editor;
+        private SingleProductThemeController _controller;
+
+        [SetUp]
+        public void Setup() {
+            _controller = GlobalGameObject.AddComponent<SingleProductThemeController>();
```
JakeCataford: using the single product theme controller cause abstract class stuff

Assets/Shopify/UIToolkit/Test/Unit/Editor/TestSingleProductThemeControllerEditor.cs
```diff
@@ -1,11 +1,10 @@
-namespace Shopify.UIToolkit.Test.Unit {
-    using UnityEngine.TestTools;
-    using UnityEditor;
+namespace Shopify.UIToolkit.Editor {
```
JakeCataford: hrm lol

Assets/Shopify/UIToolkit/Theme Controllers/Editor/SingleProductThemeControllerEditor.cs
```diff
@@ -3,54 +3,51 @@
...
-                return;
-            }
+        protected override void OnClientChanged() {
+            _productPicker.Client = Client;
         }
 
-        private void BindThemeIfPresent() {
-            if (Target.Theme != null) return;
-            Target.Theme = Target.GetComponent<ISingleProductTheme>();
+        void ISingleProductThemeControllerEditorView.DrawProductPicker() {
```
mikkoh: Why do we use `ISingleProductThemeControllerEditorView.DrawProductPicker` here.

JakeCataford: its called explicit interface implementation and it means that this method will only be available when the class is cast to `ISingleProductThemeControllerEditorView`, its just a safeguard that keeps us from using the view methods without casting the class to a view, which means that refactoring the view in the future will be easier (such as extracting it to a new class or performing dependency injection, etc).

Assets/Shopify/UIToolkit/Editor/ProductPicker.cs
mikkoh: Hmmm... Is this weird to be able to modify the client?

Assets/Shopify/UIToolkit/Theme Controllers/Editor/SingleProductThemeControllerEditor.cs
mikkoh: Yeah I was just mainly wondering about why you wanted to do this.

JakeCataford: the view concept in general? This is primarily so that we can narrow the untested surface of the controllers to only include the actual methods that touch the editor GUI (because we don't know how to test the editor). It allows us to mock the view dependency away and assert that particular things are being "told to draw" given certain states in the editor.

mikkoh: Nah I was wondering mainly about explicitly defining the interface implementation since there are no other interfaces associated with this class.

Issue #414, opened 2017-12-06
Asset store cleanups


richardmonette:
* [x] Assets/Shopify/UIToolkit/Test/Unit/Editor/ - is being bundled into the .unitypackage
* [x] https://github.com/shopify/unity-buy-sdk#using-the-unity-buy-sdk-in-unity deep link is broken in Github https://github.com/Shopify/unity-buy-sdk/pull/416
* [x] https://help.shopify.com/api/sdks/custom-storefront/unity-buy-sdk we state "    fetch information about a single product or a collection of products in Shopify" gives the feeling that you can't query all products https://github.com/Shopify/documentation/pull/13269
* [x] Maybe "Example Code Snippets" should be higher in the README.md file #432
* [ ] Our examples should include `using Shopify.Unity;`
* [x] ShopifyError does not exist at Shopify.Unity https://github.com/Shopify/unity-buy-sdk/pull/426
* [x] Example comment states "// The following queries the shop for all products" when it actually just queries the first page https://github.com/Shopify/unity-buy-sdk/pull/423
* [x] Example comment for collections states "// Queries all collections on shop" when it just queries the first page https://github.com/Shopify/unity-buy-sdk/pull/423
* [x] In EXAMPLES.md "You can also make more complex requests using custom queries (see below)." (see below) should link to the section https://github.com/Shopify/unity-buy-sdk/pull/417
* [x] In EXAMPLES.md Build a Cart does not use new paginated callback
* [x] Build a cart has invalid code where it tries to add a product as a line item instead of of a product variant ***_Note: I think this is correct, since it's passing options_*** https://github.com/Shopify/unity-buy-sdk/pull/421
* [x] Nothing is really output for the user for Build a Cart. Makes you wonder if it's actually working #427
* [x] For Build a cart we call "customAttributes" "options" https://github.com/Shopify/unity-buy-sdk/pull/423
* [x] Need to add DocBlock comments on CartLineItem #429
* [x] Example "Cart line items based on selected options" does not implement pagination using after #430
* [x] For Cart line items based on selected options. When copying pasting example we get an error stating "NoMatchingVariantException: Could not `AddOrUpdate` line item as no matching variant could be found for selected options". This is because there is no white option anymore https://github.com/Shopify/unity-buy-sdk/pull/421 #430
* [x] Supported Build Targets does not work in README.md
* [x] iOSPostProcessor didn't run and versions were set wrong
[...]

PR #428, opened 2017-12-18
add string version of AddOrUpdate

richardmonette:
Adds an overload of `AddOrUpdate` which takes a `string`:`variantId`. 

In order to ensure that the ID is **a)** valid and **b)** we get the correct price, the overload works by querying the ID and, if a successful result is found, the main `AddOrUpdate` is called.

related: https://github.com/Shopify/unity-buy-sdk/issues/414


mikkoh:
I can see there being use for this but the more we've thought about it this PR is a bit sketchy.

For instance we'd ideally like to check whether the variant is still available before adding it as a cart line item. But this introduces a lot of magic.

richardmonette:
Agreed, it makes more sense to have done all this logic **_before_** trying to add to the cart. Closing.

PR #433, opened 2017-12-19
fix pagination in examples

richardmonette:
Went through all the EXAMPLES.md, except for the Native Pay section, and found a few small issues. 

First change is because the compiler picks the wrong overload when the page size is not given.

Second change is just a missing after.

Checked on: 
screen shot 2017-12-19 at 11 46 40 am

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -50,6 +50,8 @@ using Shopify.Unity;
 using Shopify.Unity.SDK;
 
 void Start () {
+    const int productsPerPage = 10;
+
     string accessToken = "b8d417759a62f7b342f3735dbe86b322";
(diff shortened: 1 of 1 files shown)


EXAMPLES.md
```diff
@@ -50,6 +50,8 @@ using Shopify.Unity;
 using Shopify.Unity.SDK;
 
 void Start () {
+    const int productsPerPage = 10;
```
mikkoh: I feel like it's better just to tell people to use the default value since it gives us control to be able to have a "safe" value. Also the example is simpler without this.

richardmonette: Updated!

PR #435, opened 2017-12-20
[UI Toolkit] Adds image selection/loading/caching to generic single product detail controller

sleroux:
Adds loading of images to the generic single product detail prefab.

### DIFF
--- a/Assets/Shopify/UIToolkit/Components/ProductImageHolder.cs
+++ b/Assets/Shopify/UIToolkit/Components/ProductImageHolder.cs
@@ -1,0 +1,20 @@
+namespace Shopify.UIToolkit {
+    using UnityEngine;
+    using UnityEngine.UI;
+    using Image = UnityEngine.UI.Image;
+
+    public class ProductImageHolder : MonoBehaviour {
+        public Image BackgroundImage;
+        public Button ForegroundButton;
+        public Image BrokenImageIcon;
+
+        public void LoadImage(string src) {
+            ForegroundButton.gameObject.GetComponent<RemoteImageLoader>().LoadImage(
+                src, 
+                success: () => {
+                },
+                failure: null
+            );
+        }
+    }
+}
--- a/Assets/Shopify/UIToolkit/Components/ProductImageHolder.cs
+++ b/Assets/Shopify/UIToolkit/Components/ProductImageHolder.cs
@@ -9,12 +9,7 @@ namespace Shopify.UIToolkit {
         public Image BrokenImageIcon;
 
         public void LoadImage(string src) {
-            ForegroundButton.gameObject.GetComponent<RemoteImageLoader>().LoadImage(
-                src, 
-                success: () => {
-                },
-                failure: null
-            );
+            ForegroundButton.gameObject.GetComponent<RemoteImageLoader>().LoadImage(src);
         }
     }
 }
--- a/Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
+++ b/Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
@@ -46,20 +46,26 @@ namespace Shopify.UIToolkit {
             LoadImageURL(imageURL, (texture, error) => {
                 if (error != null) {
                     Debug.LogWarning("Failed to download image at " + imageURL + " Reason: " + error);
-                    failure(error);
+                    if (failure != null) {
+                        failure(error);
+                    }
                     return;
                 }
 
                 if (texture == null) {
                     Debug.LogWarning("Failed to generate texture from image located at " + imageURL);
-                    failure("Texture not found");
+                    if (failure != null) {
+                        failure("Texture not found");
+                    }
                     return;
                 }
 
                 _image.sprite = SpriteFromTexture(texture);
                 _image.preserveAspect = true;
 
-                success();
+                if (success != null) {
+                    success();
+                }
             });
         }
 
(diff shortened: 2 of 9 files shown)


Assets/Shopify/UIToolkit/Components/ProductImageHolder.cs
```diff
@@ -0,0 +1,15 @@
+namespace Shopify.UIToolkit {
+    using UnityEngine;
+    using UnityEngine.UI;
+    using Image = UnityEngine.UI.Image;
+
+    public class ProductImageHolder : MonoBehaviour {
```
sleroux: This is pretty much copied from the Example's `ImageHolder` class. We should probably consolidate in the future?

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -46,20 +46,26 @@ public class RemoteImageLoader : MonoBehaviour {
...
+                        failure("Texture not found");
+                    }
                     return;
                 }
 
                 _image.sprite = SpriteFromTexture(texture);
                 _image.preserveAspect = true;
 
-                success();
+                if (success != null) {
```
sleroux: First time actually using the image loader and was running into NPE whenever calling success here with null callbacks.

Assets/Shopify/UIToolkit/Themes/Generic/GenericSingleProductTheme.cs
```diff
@@ -18,6 +18,9 @@ public class GenericSingleProductTheme : MonoBehaviour, ISingleProductTheme {
         public GameObject VariantDropdownMenu;
         public GameObject VariantPicker;
         public GameObject VariantPickerTitle;
+        public GameObject ActiveImage;
+        public GameObject ImageContainer;
+        public GameObject ImageHolderTemplate;
```
JakeCataford: Why are these all game object? isnt it better to have them explicitly reffered to as an `Image` so we dont have to rely on a null-unsafe `GetComponent` call?

Assets/Shopify/UIToolkit/Themes/Generic/GenericSingleProductTheme.cs
```diff
@@ -66,9 +69,10 @@ public class GenericSingleProductTheme : MonoBehaviour, ISingleProductTheme {
         void ISingleProductTheme.OnShouldShowProduct(Product product, ProductVariant[] variants) {
             Title.GetComponent<Text>().text = product.title();
             Description.GetComponent<Text>().text = product.description();
```
JakeCataford: Did I write this this way? These should just be `Text` iirc but i might be a big dummy

Assets/Shopify/UIToolkit/Themes/Generic/GenericSingleProductTheme.cs
```diff
@@ -81,6 +85,29 @@ public class GenericSingleProductTheme : MonoBehaviour, ISingleProductTheme {
...
         }
 
+        public void SelectImage(Button button) {
+            ActiveImage.GetComponent<UnityEngine.UI.Image>().sprite = button.image.sprite;
+        }
+
+        private void SetupPreviewImages(Product product) {
+            var images = (List<Shopify.Unity.Image>)product.images();
+            if (images.Count == 0) {
+                // TODO: Hide image area?
```
JakeCataford: we should display a placeholder in this case imo

Assets/Shopify/UIToolkit/Themes/Generic/GenericSingleProductTheme.cs
```diff
@@ -81,6 +85,29 @@ public class GenericSingleProductTheme : MonoBehaviour, ISingleProductTheme {
...
+                return;
+            }
+
+            ActiveImage.GetComponent<RemoteImageLoader>().LoadImage(images[0].src(), () => {
+                ActiveImage.GetComponent<UnityEngine.UI.Image>().enabled = true;
+            }, null);
+
+            foreach (var image in images) {
```
JakeCataford: we should only use a subset of images incase someone has like 50, lets just take the `.First(3)`?

Assets/Shopify/UIToolkit/Themes/Generic/GenericSingleProductTheme.cs
sleroux: Could we also disable the container that holds all the images and have the product text just expand full width?

Assets/Shopify/UIToolkit/Themes/Generic/GenericSingleProductTheme.cs
sleroux: Alternatively we could add in a paging carousel but since this is just a generic view I'm good for limiting it to 3 for now.

Assets/Shopify/UIToolkit/Themes/Generic/GenericSingleProductTheme.cs
sleroux: I only thought you could drag GameObjects are script references but boy was I wrong! This is cool I didn't know you can do this.

Assets/Shopify/UIToolkit/Themes/Generic/GenericSingleProductTheme.cs
JakeCataford: might be too wide... future problems lol

PR #438, opened 2017-12-21
[UI Toolkit] Adds saving of domain/access token to EditorPrefs

sleroux:
https://github.com/Shopify/gaming-projects/issues/14

### DIFF
--- a/Assets/Shopify/UIToolkit/Theme Controllers/ThemeControllerBase.cs
+++ b/Assets/Shopify/UIToolkit/Theme Controllers/ThemeControllerBase.cs
@@ -1,5 +1,6 @@
 ﻿namespace Shopify.UIToolkit {
     using UnityEngine;
+    using UnityEditor;
     using Shopify.Unity;
     using Shopify.Unity.SDK;
     using System.Linq;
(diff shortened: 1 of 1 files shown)


sleroux:
@JakeCataford Since the domain/access token are serialized properties I'm not even sure if we need this? Was there something else you were thinking of when it comes to remembering the user's domain/tokens?

JakeCataford:
The idea is that if you create a second shop within your game, or create a new game, we default to the shopify store you have connected to before, because it's likely the one you will want to use again

Assets/Shopify/UIToolkit/Theme Controllers/ThemeControllerBase.cs
```diff
@@ -1,5 +1,6 @@
 ﻿namespace Shopify.UIToolkit {
     using UnityEngine;
+    using UnityEditor;
```
JakeCataford: this won't build (to a player), as the editor assembly is not available during the runtime. You should put this functionality into the BaseThemeControllerEditor and inject the correct credentials from there as EditorPrefs is only supposed to be used from the editor context.

Assets/Shopify/UIToolkit/Theme Controllers/ThemeControllerBase.cs
```diff
@@ -13,25 +14,47 @@ public abstract class ThemeControllerBase : MonoBehaviour, IShopCredentials {
...
+            get { 
+                if (_shopDomain == null) {
+                    _shopDomain = EditorPrefs.GetString(ShopDomainEditorPrefsKey);
+                }
+                return _shopDomain; 
+            }
             set {
                 _shopDomain = value;
                 InvalidateClient();
+                EditorPrefs.SetString(ShopDomainEditorPrefsKey, value);
```
JakeCataford: I think we should only persist these if the credentials are valid, moving this to the editor gives you a real good spot to hook into there for this case

Assets/Shopify/UIToolkit/Theme Controllers/ThemeControllerBase.cs
```diff
@@ -13,25 +14,47 @@ public abstract class ThemeControllerBase : MonoBehaviour, IShopCredentials {
...
             set {
                 _accessToken = value;
                 InvalidateClient();
+                EditorPrefs.SetString(AccessTokenEditorPrefsKey, value);
             }
         }
         [SerializeField] private string _accessToken;
+        private string AccessTokenEditorPrefsKey {
+            get {
+                return GetInstanceID() + ":accessToken";
```
JakeCataford: As per my comment from before, the goal is to save credentials across instances rather than for an individual instance (you are right that the serialization takes care of it per instance). So the key should be something like `shopify_buy_last_valid_access_token` or something (note the namespacing to avoid conflicts), especially as EditorPrefs are shared by all of the different scripts you might have installed.

sleroux:
I'm going to close this PR for now and revisit after the refactoring.

PR #445, opened 2018-01-03
[UI Toolkit] Adds show/dismiss functionality to generic sp theme

JakeCataford:
Simple show dismiss animation + animator slot for single product generic theme.

This show dismiss function works by setting a "shown" param on the animator, allowing developers to customize the animations however they want as long as they respect that "shown" param.

show_dismiss


JakeCataford:
@mikkoh The idea here was to separate the "commands" `show()` and `hide()` from the "events" onShow and onHide. While it adds a layer of indirection right now, it allows us to do generalized things like, for example, serializing and saving a cart on Hide in the base theme controller, while the theme controller subclasses can specialize for their own use case. 

If you want we can make show and hide abstract right now because we aren't leveraging this ^ and it makes it more complicated/indirect.

JakeCataford:
Closing this and will move the show/hide logic to the theme itself

PR #447, opened 2018-01-03
Remove some useless files

JakeCataford:
Both scenes are no longer required to run ITests and the examples metafile is annoying because it appears after a publish so im just deleting it.

Sanity tests are also useless because our test suite has tests in it...


richardmonette:
@JakeCataford Can you rebase this? I think that will fix the CI issues.

PR #448, opened 2018-01-04
Implemented loading events for SingleProductThemeController

mikkoh:
This PR simply implements `OnLoadingStarted` and `OnLoadingFinished`.

I do have one question should loading images be handled by themes? I think so but would love to hear thoughts.

### DIFF
--- a/Assets/Shopify/UIToolkit/ShopControllers/SingleProductShopController.cs
+++ b/Assets/Shopify/UIToolkit/ShopControllers/SingleProductShopController.cs
@@ -25,6 +25,7 @@
         }
 
         public override void OnShow() {
+            Theme.OnLoadingStarted();
             Client.products(OnProductsLoaded, ProductGID);
         }
 
@@ -38,6 +39,8 @@
                 Shop.OnError(new ShopifyError(ShopifyError.ErrorType.UserError, "Product not found"));
                 return;
             }
+
+            Theme.OnLoadingFinished();
 
             var product = products[0];
             var variants = product.variants().edges().Select((x) => x.node()).ToArray();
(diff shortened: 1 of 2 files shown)


JakeCataford:
Yea, I think the themes should be responsible for loading their own images, mostly because its a concern of "how the data is viewed" vs "how it is retreived" (ie: I could decide not to show images at all)

Assets/Shopify/UIToolkit/Theme Controllers/SingleProductThemeController.cs
```diff
@@ -25,6 +25,7 @@ public class SingleProductThemeController : ThemeControllerBase {
         }
 
         public override void OnShow() {
+            Theme.OnLoadingStarted();
```
JakeCataford: wondering if we should hook these events into all network requests that the controller makes, including when its creating a checkout for example? WDYT?

mikkoh: I feel like from a UI perspective you may need to know what type of load was started and finished. Possible implementations:

* Pass an enum to the handler function
* Have separate handler functions to implement: `OnLoadingProductStart`, `OnLoadingCheckoutStart`, etc.

As I'm working on this I do wonder how often developers would just not implement certain functions.

JakeCataford: I like this approach, its what was outlined in the tech design:

```
OnLoadingStarted(LoadingContext context) -> void

Callback that is called when the controller begins loading something.  This callback exists so that the developer can show spinners or other loading views when pieces of the theme are loading. The context contains information about what the controller started loading, specifically an enum containing the different kinds of loading that an controller can perform. The possible values are ProductList, ProductDetails, Collections and Cart.
```

JakeCataford:
Gonna close this one and tackle it with the UI instead

PR #451, opened 2018-01-08
[UI Toolkit] Scaffold Shop Theme Files

JakeCataford:
Creates a base to work on for the shop theme controller and shop themes.


richardmonette:
I'm curious how this would relate to an associated prefab. Is one of these classes supposed to be assigned to a container game object?

JakeCataford:
Ill create a prefab to go with this pr

PR #452, opened 2018-01-08
[UI Toolkit] Error Message popup view

JakeCataford:
An error message popup component that displayed error messages one by
one. The messages are first checked to see if they are already in the
queue, and if not, they are queued to be displayed.

To tophat it, you can run the integration tests and watch it play in the scene view

### DIFF
--- a/Assets/Shopify/UIToolkit/Shops/Generic/Assets/ErrorPopup.meta
+++ b/Assets/Shopify/UIToolkit/Shops/Generic/Assets/ErrorPopup.meta
@@ -1,0 +1,9 @@
+fileFormatVersion: 2
+guid: 9e2070f5c6b7d43318c1515c39ef1f76
+folderAsset: yes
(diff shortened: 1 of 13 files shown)


richardmonette:
Can we please not use spaces in filenames or paths? `...t/Themes/Generic/Assets/Error Popup/Error Popup Animator Controller.controller`

JakeCataford:
@richardmonette Spaces in filenames are very common in unity projects, unity itself names files with spaces in the filenames when you create new assets in your library, for example. I get that its harder to work with in the command line and I'm motivated by that, but I wont change it in this pr because the current state of the repo is to have the spaces. If we pr in the future to remove the spaces and change the status quo of the repo, then I will continue to follow that convention

JakeCataford:
@richardmonette updated and removed the space in the path name, left it for the filenames because thats the convention, however

PR #453, opened 2018-01-09
[UI Toolkit] Created UI for a generic Cart to be used by the UI Toolkit

sleroux:
Adds a prefabbed UI screen for a store cart as per Evans designs. Currently isn't hooked up to any scripts and contains a template game object for creating cart items in the list. I modeled the UI based on the revised layout changes Jake made in his fix up PR.


richardmonette:
Can you please post a screenshot of how this looks?

Beyond the scope of this PR, but what is the plan for moving between views, i.e. going from product view to cart view?

sleroux:
screen shot 2018-01-09 at 1 12 19 pm

> Beyond the scope of this PR, but what is the plan for moving between views, i.e. going from product view to cart view?

I'm not entirely sure. The designs we have don't really include any aspect of flow. I'm also not sure if our intention is to build a store experience or just a couple of these pre-canned prefabs for developers to mix and match. @JakeCataford thoughts?

[1] https://shopify.invisionapp.com/share/ZSEK33M83#/screens/264963815

JakeCataford:
yea I was going to start work on the container/framework for moving between views, if we keep the cart/product details etc working in prefabs so we can just drop them in at the end/when we are working on it.

PR #457, opened 2018-01-09
Added some meta files for PlayServicesResolver

mikkoh:
This PR adds some Unity `.meta` files that are currently missing.


JakeCataford:
these werent missing they were modified, but it seems like the changes are benign? im ok with it as meta file changes are usually harmless

PR #459, opened 2018-01-10
[UI Toolkit] Remove View Oriented callbacks from shop theme

JakeCataford:
We are moving towards not handling view concerns from our theme
controllers, so the interface from theme controller to theme only
requires the data that drives the theme. This reduces the interface
surface of the shop theme significantly.

### DIFF
--- a/Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericMultiProductShop.cs
+++ b/Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericMultiProductShop.cs
@@ -5,14 +5,6 @@
     using UnityEngine;
 
     public class GenericMultiProductShop : MonoBehaviour, IMultiProductShop {
-        void IMultiProductShop.OnCartItemsChanged(CheckoutLineItem[] lineItems) {
-            throw new System.NotImplementedException();
-        }
-
-        void IMultiProductShop.OnCollectionFilterChanged(Collection collection) {
-            throw new System.NotImplementedException();
-        }
-
         void IShop.OnCartQuantityChanged(int newQuantity) {
             throw new System.NotImplementedException();
         }
@@ -26,10 +18,6 @@
         }
 
         void IShop.OnLoadingStarted() {
-            throw new System.NotImplementedException();
-        }
-
-        void IMultiProductShop.OnProductListChanged(Product[] products) {
             throw new System.NotImplementedException();
         }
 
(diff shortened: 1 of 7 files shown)


Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericMultiProductShop.cs
```diff
@@ -49,19 +37,11 @@ public class GenericMultiProductShop : MonoBehaviour, IMultiProductShop {
...
-            throw new System.NotImplementedException();
-        }
-
-        void IMultiProductShop.OnShouldShowProductDetails(Product product, ProductVariant[] variants) {
+        void IMultiProductShop.OnCartItemsChanged(CheckoutLineItem[] lineItems) {
             throw new System.NotImplementedException();
         }
 
-        void IMultiProductShop.ShouldShowProductList(Product[] products) {
+        void IMultiProductShop.OnProductPageLoaded(Product[] products, int pageNumber) {
```
richardmonette: per our slack conversation, this should be updated to a after+count framework

sleroux: I was just working with this method as part of my multi product UI stuff and I ended up changing this on my branch to be:

`OnProductPageLoaded(Product[] products, int pageNu, bool hasMorePages)`

I don't think the view needs to know the string ID of after - just if there is any more pages or not. Is there any reason we could see having the ID in the view?

JakeCataford: oh good point @sleroux about knowing if theres more pages...

The convo me and @richardmonette was more about how its more flexible that you ask the controller say `gimme 10 products starting at the 11th product` then `gimme the 3rd page`, where the page would have to be some arbitrary or constant size. so a `(int after, int count)` is what we decided on.

What if we had a method on the shop controller that was something like `HasMorePages()`? Might be easier to query it when you want then be expected to keep the state around yourself from the last product load?

sleroux: In order to ask the controller that we want specific pages, we would need to store and under stand the data from the view layer which starts blurring the responsibilities of the view/controller. For what we are showing we don't really need that level of granularity. It's always hard to fully split the data from the view when they are co-dependent like this so I'm not really sure what the answer is. 

As a thought, let's say I was developing some custom UI for a shop that required me to store and do things with the data we get back from the client. Would I create a subclass of `MultiProductShopController` and add my logic there or put it in my `IMultiShop` classes?

richardmonette: IMultiShop I think

If we have after, count that should be general enough / it's the entirety of what the sdk api and the graphql api expose for this

PR #460, opened 2018-01-10
View management

JakeCataford:
view switcher

Allows pushing new views and maintaining a stack of views so that
navigating "back" always does the expected action. This will be used to
manage cart/product list/product details views in the Unity SDK factory
shops, and is designed to be easy to understand so that a 3rd party dev
can use it to build their own custom shops.

### DIFF
--- a/Assets/Shopify/UIToolkit/Components/ViewSwitcher/ViewSwitcher.cs
+++ b/Assets/Shopify/UIToolkit/Components/ViewSwitcher/ViewSwitcher.cs
@@ -3,10 +3,28 @@
     using System.Collections.Generic;
     using UnityEngine;
 
+    /// <summary>
+    /// A component that allows for switching between multiple UI "views",
+    /// where a view is different UI that needs to be switched between modally.
+    /// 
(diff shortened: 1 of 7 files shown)


Assets/Shopify/UIToolkit/Components/ViewSwitcher/ViewSwitcher.cs
```diff
@@ -0,0 +1,129 @@
...
+    /// <summary>
+    /// A component that allows for switching between multiple UI "views",
+    /// where a view is different UI that needs to be switched between modally.
+    /// 
+    /// In the Unity Buy SDK, we use this for the factory shops in order to switch
+    /// between the cart/product list/product details views in the builtin multi
+    /// product shops.
+    /// </summary>
+    public class ViewSwitcher : MonoBehaviour {
+        public const float AnimationDuration = 0.3f;
```
sleroux: Is the intention to give this to developers to use as well? If so I would make this non-const so devs can change it in the editor

PR #461, opened 2018-01-11
UI CartLineItems Propagate Events

mikkoh:
This PR refactors event handling on cart line items. One controversial thing is that when `Reset` is called an event will be fired per item being removed from the line items list.

### DIFF
--- a/Assets/Shopify/UIToolkit/ShopControllers/Classes/CartController.cs
+++ b/Assets/Shopify/UIToolkit/ShopControllers/Classes/CartController.cs
@@ -1,4 +1,5 @@
 namespace Shopify.UIToolkit {
+    using System;
     using UnityEngine.Events;
     using Shopify.Unity;
     using Shopify.Unity.SDK;
@@ -20,10 +21,6 @@ namespace Shopify.UIToolkit {
         public PurchaseFailedEvent OnPurhchaseFailed = new PurchaseFailedEvent();
 
         public CartController(Cart cart) {
-            SetCart(cart);
-        }
-
-        public void SetCart(Cart cart) {
             Cart = cart;
             Cart.LineItems.OnChange += OnLineItemsChange;
         }
--- a/Assets/Shopify/UIToolkit/ShopControllers/Classes/CartController.cs
+++ b/Assets/Shopify/UIToolkit/ShopControllers/Classes/CartController.cs
@@ -25,6 +25,7 @@ namespace Shopify.UIToolkit {
 
         public void SetCart(Cart cart) {
             Cart = cart;
+            Cart.LineItems.OnChange += OnLineItemsChange;
         }
 
         public CartLineItems LineItems {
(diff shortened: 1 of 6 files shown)


JakeCataford:
@mikkoh looks like you have some already merged commits on this branch... can you try rebase again?

mikkoh:
Hmmm... I did.

mikkoh:
@JakeCataford fixed.

richardmonette:
Conflict is just because the file in question was moved I think.

Assets/Shopify/Unity/SDK/CartLineItems.cs
```diff
@@ -418,6 +425,19 @@ public class CartLineItems {
             return Delete(variantId);
         }
 
+        /// <summary>
+        /// Deletes all line items
+        /// </summary>
+        public void Reset() {
+            if (OnChange != null) {
```
richardmonette: Should this really wrap the whole thing? I would have thought this would always run the removal, and then the condition would be around OnChange call?

sleroux: Agreed - we should extract out the actual removal from the callback triggering :+1:

Assets/Shopify/Unity/SDK/CartLineItems.cs
```diff
@@ -161,6 +161,14 @@ public class CartLineItem {
     /// Is used to add, update, or delete line items in a <see ref="Cart">Cart </see>.
     /// </summary>
     public class CartLineItems {
+        public enum LineItemChangeType {
+            add,
```
sleroux: Nit: Looks like we've either used CamelCasing or CAP_CASING for enums. Should standardize around CamelCasing?

Assets/Shopify/UIToolkit/ShopControllers/Classes/CartController.cs
```diff
@@ -25,6 +25,7 @@ public class PurchaseFailedEvent : UnityEvent<ShopifyError> {}
 
         public void SetCart(Cart cart) {
             Cart = cart;
+            Cart.LineItems.OnChange += OnLineItemsChange;
```
richardmonette: Would it be worth being a bit more defensive here, and making sure that the event is only added once? Might help to prevent any bugs cropping up in the future.

Assets/Shopify/Unity/CartState.cs
```diff
@@ -43,11 +43,12 @@ public partial class CartState {
 
         public CartState(ShopifyClient client) {
             Client = client;
-            _LineItems = new CartLineItems(OnDeleteLineItem);
+            _LineItems = new CartLineItems();
+            _LineItems.OnChange += OnLineItemChange;
```
richardmonette: Same thing here about this potential getting re-registered.

Assets/Shopify/Unity/SDK/CartLineItems.cs
mikkoh: Sorry do you mean Pascal Case? I'm cool with Pascal.

Assets/Shopify/UIToolkit/ShopControllers/Classes/CartController.cs
mikkoh: I'm gonna make it so that `Cart` can only be set through the constructor.

Assets/Shopify/Unity/CartState.cs
mikkoh: You can't really re-register here since the `CartLineItems` instance is created in the constructor for `CartState`. But you are right that we potentially need to have a destructor/cleanup function.

PR #464, opened 2018-01-11
[UI Toolkit] Force repaint in shop controller editor when state changes

JakeCataford:
Forces a repaint when the credentials verifier or the product picker
finished loading and updates the state of the editor. This fixes a bug
where some movement/UI event is required to show the state changes.

### DIFF
--- a/Assets/Shopify/UIToolkit/Editor/ProductPicker.cs
+++ b/Assets/Shopify/UIToolkit/Editor/ProductPicker.cs
@@ -20,6 +20,9 @@
     }
 
     public class ProductPicker : IProductPickerController {
+        public delegate void ProductListUpdatedHandler();
+        public event ProductListUpdatedHandler OnProductListUpdated;
+
         public ShopifyClient Client;
         private IProductPickerView _view;
         private static Dictionary<string, string> _cachedProductNameToGIDMap;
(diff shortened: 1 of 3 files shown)


Assets/Shopify/UIToolkit/Shop Controllers/Editor/SingleProductShopControllerEditor.cs
```diff
@@ -32,7 +32,11 @@ public class SingleProductShopControllerEditor : ShopControllerBaseEditor, ISing
 
         private ProductPicker _productPicker {
             get {
-                _cachedPicker = _cachedPicker ?? new ProductPicker(Client);
+                if (_cachedPicker == null) {
+                    _cachedPicker = new ProductPicker(Client);
+                    _cachedPicker.OnProductListUpdated += ProductListDidUpdate;
```
richardmonette: Can you just add `Repaint` directly, without an extra method in between? Or maybe just a lambda?

JakeCataford: Sure, I thought it was a bit clearer from like a high level that things happen when the product list updates in that callback method or whatever?

PR #468, opened 2018-01-12
Created UI for multi list product shop along with some scripts to loading the first page of products

sleroux:
screen shot 2018-01-12 at 11 59 50 am

Small PR that builds out the basic multi product store UI along with some code to load in the first page of products.

Key things that are missing:

* Ability to select a product/dive into a product detail
* Paging of products


Assets/Shopify/UIToolkit/ShopControllers/MultiProductShopController.cs
```diff
@@ -1,10 +1,40 @@
...
+            } 
+        }
+
+        public override void OnHide() {
+            throw new System.NotImplementedException();
+        }
+
+        public override void OnShow() {
+            Client.products(OnProductsLoaded);
+        }
```
JakeCataford: worth noting that we are going to remove the hide/show func from the shop controller and leave it as a view concern... we will fix this forward thought

PR #469, opened 2018-01-15
[UI Toolkit] Multi Product Shop Base

JakeCataford:
Creates subviews for every part of the multi product shop, brings in the
implemented product list and creates prefabs for each subview.


JakeCataford:
going to change 2 things before merge -> injecting the Shop as a dep to the subviews and instantiating the subviews from prefabs

JakeCataford:
I added the injection and prefab instantiation we talked about... gonna merge when green and we can fix forward if theres bad things

richardmonette:
@JakeCataford did you include the fix for the duplicated CartController? https://github.com/Shopify/unity-buy-sdk/pull/470 Otherwise this won't make it through CI

PR #471, opened 2018-01-15
IShopController

mikkoh:
This effects #448 

This PR does the following:

* Adds in `IShopController`
* Removes `OnShow` `OnHide` in favour of `Load` and `UnLoad`

In a later PR I'd like to see if there's some way to clean up/isolate the credentials management.


Assets/Shopify/UIToolkit/Editor/ShopCredentialsVerifier.cs
```diff
@@ -76,15 +76,15 @@ public class ShopCredentialsVerifier {
         public delegate void OnCredentialsStateChangedHandler();
         public event OnCredentialsStateChangedHandler OnCredentialsStateChanged;
 
-        public IShopCredentials Credentials { private set; get; }
+        public IShopController Credentials { private set; get; }
```
JakeCataford: Why hijack `IShopCredentials`? I'd rather limit the interfaces to specific use cases and use multiple inheritance? Also, I liked @sleroux's idea of creating a IShopCredentials class instead of having that interface, and the object is serialized into the shop controller. With that in mind this makes sense, because IShopController can have a singe `ShopCredentials GetCredentials();` method

sleroux: What I find odd about this declaration is that we're saying that `Credentials` is a `IShopController`  which doesn't really seem right. I would either rename this to signify a trait (`IHasCredentials`) or just change it so it's a plain old data struct.

mikkoh: So my thought was to clean this up in the next PR.

I think the `IShopController` should contain an `IShopCredentials` instance. But for now to not break tests and make this PR large I did this.

JakeCataford:
im gonna rebase and blast this one

PR #472, opened 2018-01-16
Cart View Beginnings + CartItem/CartItemView bindings

sleroux:
This PR starts building out the pieces we need to drive the cart view UI. It includes:

* Beginnings of a `GenericCartView` class that connects with the cart UI and adds items to the list.
* `CartItemView`/`CartItemViewModel` for binding the cart data models into each list item view.
* `CartItem` abstraction for bundling the line item/product variant information together inside the `CartController` so we can use it up the call chain for the UI.

I've also added a `OnCartItemsChange` event to `CartController` that will push out all of the cart items whenever an update is made but I still need to figure out a good way of connecting this to the view. Currently all of the CartController events are being listened to the `ShopControllerBase` class. I would be nice to bind this event directly to the view but not sure how that would work with out `IShop` abstraction. 

Still needs tests but figured I would throw it up for comments first before I move forward and gather your thoughts 😎


Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericCartView.cs
```diff
@@ -0,0 +1,53 @@
+namespace Shopify.UIToolkit {
+    using UnityEngine;
+    using UnityEngine.UI;
+    using Shopify.Unity;
+    using Shopify.Unity.SDK;
+    using System.Collections.Generic;
+    using System;
+
+    public class GenericCartView : MonoBehaviour {
```
JakeCataford: I think `CartView` and `GenericViewBase or something` exist already, mind merging this with those classes?

Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericCartView.cs
```diff
@@ -0,0 +1,53 @@
...
+        public IShop Shop;
```
JakeCataford: I dont think the Generic Cart View should talk to the shop via IShop, I think it should hold a more specific reference `GenericMultiProductShop` or alternatively talk through a new `IGenericShop` interface or something

Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericCartView.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        #endregion
+
+        #region Events
+        void OnQuantityChanged(int quantity) {
+        }
+
+        void OnCartItemsChanged(List<CartItem> cartItems) {
+            UpdateCartList(cartItems);
+        }
```
JakeCataford: I dig this

Assets/Shopify/UIToolkit/Shops/Generic/Classes/CartItemView.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// A View Model for converting the CartItem data into a format for the CartItemView component.
+    /// </summary>
+    public struct CartItemViewModel {
+        public readonly string Title;
+        public readonly string VariantDescription;
+        public readonly string PriceText;
+        public readonly string QuantityText;
+        public readonly string ImageURL;
+
+        public CartItemViewModel(CartItem cartItem) {
```
JakeCataford: I honestly had no idea structs could have constructors hahaha

Assets/Shopify/UIToolkit/Shops/Generic/Classes/CartItemView.cs
```diff
@@ -0,0 +1,67 @@
...
+            var product = variant.product();
+
+            Title = product.title();
+            VariantDescription = variant.title();
+            PriceText = string.Format("${0}", variant.price());
+            QuantityText = cartItem.Quantity.ToString();
+
+            string imageURL;
+            try {
+                imageURL = variant.image().src();
```
JakeCataford: are we supposed to ignore this src attribute now? whats the deal with the deprecation

Assets/Shopify/UIToolkit/Shops/Debug/DebugSingleProductShop.cs
```diff
@@ -44,6 +45,10 @@ public class DebugSingleProductShop : MonoBehaviour, ISingleProductShop {
             LogEvent("OnCartQuantityChanged", newQuantity);
         }
 
+        public void OnCartItemsChanged(List<CartItem> cartItems) {
+            LogEvent("OnCartItemsChanged", cartItems);
+        }
```
JakeCataford: This event should exist in the MultiProduct shop but interface I don't think it should exist in the single product shop interface, as the purpose of the single product shop is to immediately buy a single product once.

Assets/Shopify/UIToolkit/Shops/Interfaces/IShop.cs
```diff
@@ -58,5 +58,12 @@ public interface IShop {
         /// </summary>
         /// <param name="newQuantity">The new number of items in the cart</param>
         void OnCartQuantityChanged(int newQuantity);
+
+        /// <summary>
+        /// Called when anything changes with the items that are in the cart. For example,
+        /// adding/removing/updating an item in your cart will trigger this.
+        /// </summary>
+        /// <param name="cartItems">List of cart items in the cart</param>
+        void OnCartItemsChanged(List<CartItem> cartItems);
```
JakeCataford: Wanna move this + OnCartQuantity changed to the `IMultiProductShop` interface? WDYT? does single product shop need a cart?

Assets/Shopify/UIToolkit/Shops/Debug/DebugSingleProductShop.cs
sleroux: I guess along those same lines we should also remove `OnCartQuantityChanged`?

Assets/Shopify/UIToolkit/Shops/Interfaces/IShop.cs
sleroux: If we're saying a single product shop is like a 'buy this thing now' kind of store then we can remove. If we want to be able to add multiple of a single item then we'll probably need a cart. I say we should just remove it for now and go with the first option.

Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericCartView.cs
sleroux: Agreed - I made the update so the shop invokes methods on the `GenericCartView` instance it has.

Assets/Shopify/UIToolkit/Shops/Generic/Classes/CartItemView.cs
sleroux: Not sure - I'll investigate after this PR

Assets/Shopify/UIToolkit/ShopControllers/Classes/CartController.cs
```diff
@@ -42,7 +74,12 @@ public class PurchaseFailedEvent : UnityEvent<ShopifyError> {}
             var newQuantity = existingItem == null ? 1 : existingItem.Quantity + 1;
             Cart.LineItems.AddOrUpdate(variant, newQuantity);
 
+            if (!_idsToVariants.ContainsKey(variant.id())) {
```
richardmonette: This appears twice, maybe make a method?

richardmonette: Oh, three times I think actually, might be a method.

Assets/Shopify/UIToolkit/Shops/Debug/DebugSingleProductShop.cs
```diff
@@ -1,5 +1,6 @@
 ﻿namespace Shopify.UIToolkit.Shops {
     using System.Linq;
+    using System.Collections.Generic;
```
richardmonette: Needed?

Assets/Shopify/UIToolkit/Shops/Generic/Classes/CartItemView.cs
richardmonette: Should use transformedSrc now IIRC

Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericCartView.cs
```diff
@@ -0,0 +1,52 @@
+namespace Shopify.UIToolkit.Shops.Generic {
+    using UnityEngine;
+    using UnityEngine.UI;
+    using Shopify.Unity;
+    using Shopify.Unity.SDK;
+    using System.Collections.Generic;
+    using System;
+
+    public class GenericCartView : GenericMultiProductShopView {
```
richardmonette: Should this just be named CartView? (The other views don't seem to have this prefix..?)

Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericSingleProductShop.cs
```diff
@@ -1,6 +1,7 @@
 ﻿namespace Shopify.UIToolkit.Shops {
     using Shopify.Unity;
     using Shopify.Unity.SDK;
+    using System.Collections.Generic;
```
richardmonette: Need this?

Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericCartView.cs
sleroux: I'm not sure because the interface we implement is named `GenericMultiProductShopView`. I feel like if the interface is prefixed with `Generic` then it make sense for this to be `Generic` as well?

richardmonette: Fair enough, but I think that would imply we rename all the other views as well, perhaps in another PR?

sleroux: Ah I see what you mean - like CartItemView and stuff. Yea I wonder if we want to either prefix all this with Generic? Or if having them in the folder is enough for people to know?

richardmonette: imho adding generic in front doesn't add anything. I suspect people will just copy these classes, or the prefabs wholesale, so these are probably more likely to _be_ the the implementation in the end.

sleroux:
I've pushed up fixes for the comments and renamed GenericCartView->CartView for now. We can do a follow up PR for removing the Generic prefix from the other classes if we want.

PR #475, opened 2018-01-23
[UI Toolkit] Product Details in multi product shop

JakeCataford:
Refactors @sleroux's work on the single product shop into an object that can be reused in the single product shop and the multi product shop's view layers for displaying product details.

The first commit is the feature itself, and could be merged as is. The second commit is the minimum amount of work I had to do to demo the feature.

pd

### DIFF
--- a/Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericSingleProductShop.cs
+++ b/Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericSingleProductShop.cs
@@ -8,38 +8,14 @@
     using System.Linq;
     using System.Collections.Generic;
     using System.Globalization;
+    using Shopify.UIToolkit.Shops.Generic;
 
     [RequireComponent(typeof(SingleProductShopController))]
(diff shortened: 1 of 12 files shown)


Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericSingleProductShop.cs
```diff
@@ -86,123 +62,19 @@ public class GenericSingleProductShop : MonoBehaviour, ISingleProductShop {
...
         void ISingleProductShop.OnProductLoaded(Product product, ProductVariant[] variants) {
-            _productVariants = variants;
-
-            Title.text = product.title();
-            Description.text = product.description();
-            SetupVariantOptions(variants);
-
-            UpdateDetailsUsingVariant(variants[0]);
-
-            var images = product.images().edges().Select((x) => x.node()).ToArray();
```
JakeCataford: It would be really cool to have an extension method that can get `Variants()` directly from a `Product()`, that way we arent breaking the graphql model, but also allowing for that more convenient approach.

PR #476, opened 2018-01-23
add load more to multi product

richardmonette:
Splitting this off from my main infinite pagination PR. 

Adds a method to the `MultiProductShopController` which exposes triggering fetching more products. Handles dealing with storing the cursor and eventually setting `allProductsLoaded` to prevent more network calls once the entire product list is loaded. 

The `ProductListView` inherits from `GenericMultiProductShopView` which has a reference to the  `IGenericMultiProductShop` which, in turn, has a reference to `MultiProductShopController`, so adding a method in order to delegate.

### DIFF
--- a/Assets/Shopify/UIToolkit/Components/ProductCache.cs
+++ b/Assets/Shopify/UIToolkit/Components/ProductCache.cs
@@ -1,0 +1,46 @@
+namespace Shopify.UIToolkit {
+    using System;
+    using System.Collections.Generic;
+    using Shopify.Unity;
+
+    public class ProductCache {
+        private string _cursor = null;
+        private bool _complete = false;
+
+        private List<Product> _products;
+
+        public ProductCache() {
+            _products = new List<Product> ();
+        }
+
+        public string Cursor {
+          get {
+            return _cursor;
+          }
+        }
+
+        public bool Complete {
+          get {
+            return _complete;
+          }
+        }
+
+        public int Count {
+          get {
+            return _products.Count;
+          }
+        }
+
+        public void Add(Product[] products, string after) {
+            if (string.IsNullOrEmpty(after)) {
+                _complete = true;
+            }
+
+            _cursor = after;
+
+            foreach (var product in products) {
+                _products.Add(product);
+            }
+        }
+    }
+}
--- a/Assets/Shopify/UIToolkit/ShopControllers/MultiProductShopController.cs
+++ b/Assets/Shopify/UIToolkit/ShopControllers/MultiProductShopController.cs
@@ -5,14 +5,17 @@
     using UnityEngine;
 
     public class MultiProductShopController : ShopControllerBase {
-        public new IMultiProductShop Shop { 
-            get { 
-                return base.Shop as IMultiProductShop; 
+        private string cursor = null;
+        private bool allProductsLoaded = false;
+
+        public new IMultiProductShop Shop {
+            get {
+                return base.Shop as IMultiProductShop;
             }
 
             set {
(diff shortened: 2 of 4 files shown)


Assets/Shopify/UIToolkit/ShopControllers/MultiProductShopController.cs
```diff
@@ -5,22 +5,33 @@
...
+            }
         }
 
         public override void OnHide() {
             throw new System.NotImplementedException();
         }
 
         public override void OnShow() {
-            Client.products(OnProductsLoaded);
+            LoadMoreProducts ();
```
sleroux: Nit: Whitespace between `LoadMoreProducts` and `()`

Assets/Shopify/UIToolkit/ShopControllers/MultiProductShopController.cs
```diff
@@ -5,22 +5,33 @@
...
+        }
+
+        public void LoadMoreProducts() {
+            if (allProductsLoaded) {
+                return;
+            }
+
+            Client.products(OnProductsLoaded, after: cursor);
```
JakeCataford: hmm, wondering if the theme should be sending the cursor? or the number of products to load? In our case we dont really care but imagine an app that has a more classic pagination style, they would want to specify what products to load right?

richardmonette:
> It also feels a bit odd to separate the cursor and allProductsLoaded state from the place we cache the products but I guess that's a side effect of how this is all setup with the generic views.

Yeah, this is because the controller appears to have the responsibility for actually talking to `Client` https://github.com/Shopify/unity-buy-sdk/pull/476/files#diff-267aa3d90e46fc978476b96dae937d72R34 Thinking about it some more, we could have another class in the Shop, called ProductCache or something, which would be able to hold those attributes. However, then the controller would kind of need to reach into the Shop.

Lemme take a look at pulling a bit more of the caching into this PR to show and explore that idea.

richardmonette:
Created `ProductCache` to hold the products, moved most of the logic into the `Shop` (and out of the Controller.)

Assets/Shopify/UIToolkit/Shops/Generic/Classes/GenericMultiProductShop.cs
```diff
@@ -152,5 +161,9 @@ public class GenericMultiProductShop : MonoBehaviour, IMultiProductShop, IGeneri
             CloseButton.gameObject.SetActive(!canNavigateBack);
             BackButton.gameObject.SetActive(canNavigateBack);
         }
+
+        public void LoadMoreProducts() {
+            _controller.LoadMoreProducts (_productCache.Cursor);
```
JakeCataford: I'd like to rework this in another pr maybe... feels like it should just be what you said earlier as `LoadProducts(from: cursor, count: 5)` or whatever

Assets/Shopify/UIToolkit/Components/ProductCache.cs
```diff
@@ -0,0 +1,46 @@
...
+        private string _cursor = null;
+        private bool _complete = false;
+
+        private List<Product> _products;
+
+        public ProductCache() {
+            _products = new List<Product> ();
+        }
+
+        public string Cursor {
```
JakeCataford: you can make these properties more concise by using `public string Cursor { get; private set; }`

richardmonette: When declared like this, I take it `private string _cursor = null;` would not be required?

PR #477, opened 2018-01-23
Adds increasing/decreasing of quantities for cart items and a basic empty state

sleroux:
Connects the increase/decrease quantity buttons from the CartItemViews to the Shop in order to tell the CartController when to change the quantity. When the quantity is changed, the OnCartItemsChanged event is triggers causing the CartView to update.

@JakeCataford I feel like this stuff would fit right into the work your doing in hooking everything up.

### DIFF
--- a/Assets/Shopify/UIToolkit/ShopControllers/Classes/CartController.cs
+++ b/Assets/Shopify/UIToolkit/ShopControllers/Classes/CartController.cs
@@ -84,7 +84,7 @@ namespace Shopify.UIToolkit {
         /// </summary>
         /// <param name="variant">The variant to modify</param>
         /// <param name="quantity">The desired quantity</param>
-        public void UpdateVariant(ProductVariant variant, int quantity) {
+        public void UpdateVariant(ProductVariant variant, long quantity) {
             if (quantity <= 0) {
                 Cart.LineItems.Delete(variant);
                 RemoveVariantByIDFromCache(variant.id());
--- a/Assets/Shopify/UIToolkit/Shops/Debug/DebugSingleProductShop.cs
+++ b/Assets/Shopify/UIToolkit/Shops/Debug/DebugSingleProductShop.cs
@@ -1,6 +1,7 @@
 ﻿namespace Shopify.UIToolkit.Shops {
     using System.Linq;
     using System.Text;
+    using System.Collections.Generic;
     using Shopify.Unity;
     using Shopify.Unity.SDK;
     using UnityEngine;
@@ -40,8 +41,16 @@
             LogEvent("OnProductLoaded", product, variants);
         }
 
-        public void OnCartQuantityChanged(int newQuantity) {
-            LogEvent("OnCartQuantityChanged", newQuantity);
+        void IShop.OnCartQuantityChanged(int totalNumberOfCartItems) {
+            LogEvent("OnCartQuantityChanged", totalNumberOfCartItems);
+        }
+
+        void IShop.OnCartItemsChanged(List<CartItem> cartItems) {
+            LogEvent("OnCartItemsChanged");
+        }
+
+        void IShop.UpdateCartQuantityForVariant(ProductVariant variant, long quantity) {
+            LogEvent("UpdateCartQuantityForVariant");
         }
 
         private void LogEvent(string eventName, params object[] args) {
--- a/Assets/Shopify/UIToolkit/Shops/Generic/Classes/CartItemView.cs
+++ b/Assets/Shopify/UIToolkit/Shops/Generic/Classes/CartItemView.cs
@@ -1,67 +1,71 @@
 namespace Shopify.UIToolkit {
     using Shopify.Unity;
     using Shopify.Unity.SDK;
+    using Shopify.UIToolkit.Shops.Generic;
     using System.Collections.Generic;
     using System.Linq;
     using System;
     using UnityEngine;
+    using UnityEngine.Events;
     using UnityEngine.UI;
-
-    /// <summary>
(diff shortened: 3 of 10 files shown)


Assets/Shopify/UIToolkit/ShopControllers/Classes/CartController.cs
```diff
@@ -84,7 +84,7 @@ public class CartItemsChangeEvent : UnityEvent<List<CartItem>> {}
         /// </summary>
         /// <param name="variant">The variant to modify</param>
         /// <param name="quantity">The desired quantity</param>
-        public void UpdateVariant(ProductVariant variant, int quantity) {
+        public void UpdateVariant(ProductVariant variant, long quantity) {
```
sleroux: The Shopify SDK uses `long`s for representing quantities so instead of flooring it down I just updated the method to use `int`.

Assets/Shopify/UIToolkit/Shops/Generic/Classes/CartItemView.cs
```diff
@@ -1,67 +1,71 @@
...
     using System.Linq;
     using System;
     using UnityEngine;
+    using UnityEngine.Events;
     using UnityEngine.UI;
 
-    /// <summary>
-    /// A View Model for converting the CartItem data into a format for the CartItemView component.
-    /// </summary>
-    public struct CartItemViewModel {
```
sleroux: This view model was a bit overkill for what we needed. It also got in the way of passing in the actual model to the line item view so we can know which product variant to increase/decrease quantity for.

Assets/Shopify/UIToolkit/Shops/Interfaces/IShop.cs
```diff
@@ -53,10 +53,22 @@ public interface IShop {
...
+        /// </summary>
+        /// <param name="lineItems">The new line items in the cart</param>
+        void OnCartItemsChanged(List<CartItem> lineItems);
+
+        /// <summary>
+        /// Call to update the quantity of the given variant in the cart.
+        /// </summary>
+        /// <param name="variant">Variant object that is currently in the user's Cart.</param>
+        /// <param name="quantity">The quantity value to change the cart item to.</param>
+        void UpdateCartQuantityForVariant(ProductVariant variant, long quantity);
```
sleroux: There was some discussion in a previous PR about keeping the Cart callbacks on the `IMultiProductShop` interface instead of `IShop` but there are a few issues I ran into in doing so:

1. The `CartController` lives in the `ShopControllerBase` which means the cart is accessible to all shops. If we wanted to change this we'd probably want to move it to a `MultiShopControllerBase` or something.

2. Even though a single product shop might not technically need a cart _view_, the concept of adding to a cart behind the scenes is still valid. There's also the case of adding/buying multiple items from a single product to consider.

Assets/Shopify/UIToolkit/ShopControllers/Classes/CartController.cs
JakeCataford: lol sounds good but who would ever order > `int.MaxValue` products in a cart hahaha

Assets/Shopify/UIToolkit/Shops/Generic/Classes/CartItemView.cs
JakeCataford: could always introduce a presenter if you need a layer to transform data in

Assets/Shopify/UIToolkit/Shops/Debug/DebugSingleProductShop.cs
```diff
@@ -40,8 +41,16 @@ public class DebugSingleProductShop : MonoBehaviour, ISingleProductShop {
             LogEvent("OnProductLoaded", product, variants);
         }
 
-        public void OnCartQuantityChanged(int newQuantity) {
-            LogEvent("OnCartQuantityChanged", newQuantity);
+        void IShop.OnCartQuantityChanged(int totalNumberOfCartItems) {
```
JakeCataford: nit: should this be long as well?

Assets/Shopify/UIToolkit/Shops/Generic/Classes/CartItemView.cs
```diff
@@ -1,67 +1,71 @@
...
+        public GenericMultiProductShop Shop;
+
+        [Header("View Properties")]
         public RemoteImageLoader ImageLoader;
         public Text TitleLabel;
         public Text PriceLabel;
         public Text VariantLabel;
         public Text QuantityLabel;
+        public Button DecreaseQuantityButton;
+        public Button IncreaseQuantityButton;
```
JakeCataford: I dont think you use these reffs?

Assets/Shopify/UIToolkit/ShopControllers/Classes/CartController.cs
sleroux: Maybe someone _really_ likes doritoes who knows? I think it's just a GraphQL/API thing.

Assets/Shopify/UIToolkit/Shops/Debug/DebugSingleProductShop.cs
sleroux: I don't think so since this isn't a value driven by the GraphQL API. `int` should be alright here.

Assets/Shopify/UIToolkit/Shops/Generic/Classes/CartItemView.cs
sleroux: Leftover code!

PR #478, opened 2018-01-23
add ui toolkit testing scene

JakeCataford:
Adds a simple testing scene with buttons for spawning the single and multi product shop prefabs.


richardmonette:
I am seeing the same bug, could reasonably be fixed forward, or in this PR.

JakeCataford:
yea seems unrelated to this at least

PR #480, opened 2018-01-24
Connected checkout methods to Cart Panel and added Apple Merchant ID to base controller

sleroux:
Small PR that connects the buttons to the `StartPurchase` method in the CartController. I've also abstracted the NativePayKey into a property that lives on the ShopController and can be edited from the editor.

Instead of exposing a 'Native Pay Key' which seems pretty ambigious I've exposed a 'Apple Merchant ID' field for Apple Pay instead. When Android Pay is supported we can add another optional field named Android Key or something. This way it's a lot more clear.

### DIFF
--- a/Assets/Shopify/UIToolkit/ShopControllers/BaseShopController.cs
+++ b/Assets/Shopify/UIToolkit/ShopControllers/BaseShopController.cs
@@ -28,7 +28,6 @@ namespace Shopify.UIToolkit {
                 InvalidateClient();
             }
         }
-        [Header("Store Properties")]
         [SerializeField] private string _shopDomain;
 
         /// <summary>
@@ -43,14 +42,6 @@ namespace Shopify.UIToolkit {
         }
         [SerializeField] private string _accessToken;
 
-        /// <summary>
-        /// The Apple Merchant ID for processing Apple Pay payments.
-        /// </summary>
-        public string AppleMerchantID {
-            set { Cart.AppleMerchantID = value; } 
-            get { return Cart.AppleMerchantID; }
-        }
-        [Header("Payment Properties")]
         [SerializeField] private string _appleMerchantID;
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


JakeCataford:
Quick idea: what if we only show this when the publishing platform is iOS?

sleroux:
Sounds good 👍

Assets/Shopify/UIToolkit/ShopControllers/BaseShopController.cs
```diff
@@ -42,6 +43,16 @@ public abstract class ShopControllerBase : MonoBehaviour, IShopCredentials {
...
 
+        /// <summary>
+        /// The Apple Merchant ID for processing Apple Pay payments.
+        /// </summary>
+        public string AppleMerchantID {
+            set { Cart.AppleMerchantID = value; } 
+            get { return Cart.AppleMerchantID; }
+        }
+        [Header("Payment Properties")]
+        [SerializeField] private string _appleMerchantID;
```
JakeCataford: the header attributes here serve no purpose because the custom editor overrides it

Assets/Shopify/UIToolkit/ShopControllers/BaseShopController.cs
```diff
@@ -42,6 +43,16 @@ public abstract class ShopControllerBase : MonoBehaviour, IShopCredentials {
         }
         [SerializeField] private string _accessToken;
 
+        /// <summary>
+        /// The Apple Merchant ID for processing Apple Pay payments.
+        /// </summary>
+        public string AppleMerchantID {
+            set { Cart.AppleMerchantID = value; } 
+            get { return Cart.AppleMerchantID; }
+        }
```
JakeCataford: Instead of proxying here, can we invert the dependency? (cart is initialized with the apple merchant id from this class?)

JakeCataford: Or if the class is serialized, you can use `FindProperty("Cart.AppleMerchantID")` in the editor to expose it without having to edit this class at all. The Cart object is serialized so therefore it will persist any changes made to it.

PR #481, opened 2018-01-24
add infinite pagination to product list

richardmonette:
infinite scroll

Needs more work refining the UX, but this adds in the infinite pagination!

### DIFF
--- a/Assets/Shopify/UIToolkit/Shops/Generic/Classes/ProductListView.cs
+++ b/Assets/Shopify/UIToolkit/Shops/Generic/Classes/ProductListView.cs
@@ -18,7 +18,6 @@
 		private bool loadingMore;
 		private bool firstLoad = true;
 
-        private List<Product> _products = new List<Product> ();
         private List<MultiProductListItem> _scrollPoolElements = new List<MultiProductListItem> ();
 
         public ContentRegion contentRegion;
@@ -28,8 +27,6 @@
 
 			RectTransform rtt = (RectTransform)ListItemTemplate.transform;
 			elementWidth = rtt.rect.width;
-
-
 
 			for (int i = 0; i < potentiallyVisibleElements; i++) {
 				var element = Instantiate (ListItemTemplate, content.transform);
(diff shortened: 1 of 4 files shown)


sleroux:
I just tried to run this locally using the Testing scene but I'm getting the following errors:

screen shot 2018-01-26 at 3 16 50 pm

richardmonette:
@sleroux The error is because the ContentRegion isn't set on the ProductListView prefab, not sure why the Unity editor isn't seeing that change : /

richardmonette:
As discussed IRL, should be 🆗 after the latest commit!

Assets/Shopify/UIToolkit/Shops/Generic/Classes/ProductListView.cs
```diff
@@ -9,17 +9,91 @@
...
-                });
+            if (firstLoad) {
+                for (int i = 0; i < potentiallyVisibleElements; i++) {
+                    _scrollPoolElements[i].SetProduct (products [i]);
+                }
+
+                firstLoad = false;
             }
+
+            contentRegion.endOffset = -elementWidth * (shop.ProductCache.Count - 5);
```
sleroux: Should this 5 be the `potentiallyVisibleElements` value? If so we should use the variable instead of a constant here.

richardmonette:
screen shot 2018-01-29 at 2 44 51 pm

Fixed:

- Vertical layout
- Prevents list repeating

sleroux:
When I try to 🎩 I'm getting the same errors as I first got with this PR where the links between the Content Region/ListView are broken on the prefab and the ContentRegion script isn't attached 😕

richardmonette:
Looks like something is mixed up with the commit history and that fix is missing, will investigate

richardmonette:
@sleroux deleted and re-did the assignment which allowed me to make an explicit commit again for the ContentRegion assignment issues

Also, made an issues for tracking the remaining todo/bugs for product list view https://github.com/Shopify/unity-buy-sdk/issues/490

PR #484, opened 2018-01-24
Fixes resizing issues with Native Pay Button UI and iOS compile issues

sleroux:
* Fixes a couple of bugs resulting in NPEs when loading a NativePayButtonUI. If any aspects of the size were 0, it would crash.
* NativePayButtonUI now properly redraws the image when resized.
* Fixed an annoying bug when exporting a non-test iOS build for testing the SDK with.

### DIFF
--- a/Assets/Shopify/Unity/UI/NativePayButtonUI.cs
+++ b/Assets/Shopify/Unity/UI/NativePayButtonUI.cs
@@ -64,8 +64,6 @@ namespace Shopify.Unity.UI {
         private void GenerateNativePayImage() {
         #if UNITY_IOS
             GenerateApplePayImage();
-        #elif UNITY_ANDROID
-            GenerateAndroidPayImage();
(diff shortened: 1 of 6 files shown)


Assets/Shopify/Unity/UI/NativePayButtonUI.cs
```diff
@@ -99,10 +111,10 @@ public class NativePayButtonUI : MonoBehaviour {
 #endif
 
 #if UNITY_ANDROID
-        private void GenerateAndroidPayImage() {
+        private void GenerateAndroidPayImage(RectTransform rectTransform) {
```
richardmonette: Needed?

sleroux: Nope. I'm going to remove this stuff for now until we get around to it.

PR #485, opened 2018-01-24
fast cache RemoteImageLoader

richardmonette:
Instead of waiting for 304, we can update the target image right away which, in the common case will result in quicker UI. In the case we need to re-cache, will work same as before.

### DIFF
--- a/Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
+++ b/Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
@@ -18,16 +18,16 @@ namespace Shopify.UIToolkit {
 
         private Image _image;
 
-        private delegate void RemoteImageCompletionDelegate(Texture2D texture, string error);
+        private delegate void RemoteImageLoadedDelegate(Texture2D texture, string error);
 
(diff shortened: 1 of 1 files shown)


Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -80,33 +80,38 @@ public class RemoteImageLoader : MonoBehaviour {
...
-                StartCoroutine(CachedLoadImageURLRoutine(url, completion));
+                StartCoroutine(CachedLoadImageURLRoutine(url, onImage));
             } else {
-                StartCoroutine(LoadImageURLRoutine(url, completion));
+                StartCoroutine(LoadImageURLRoutine(url, onImage));
             }
         }
 
-        private IEnumerator LoadImageURLRoutine(string url, RemoteImageCompletionDelegate completion) {
+        private IEnumerator LoadImageURLRoutine(string url, RemoteImageCompletionDelegate onImage) {
```
sleroux: Nit: I would also rename the delegate to like `OnRemoteImageDelegate` or something

PR #486, opened 2018-01-26
Credentials as an object

JakeCataford:
This makes IShopCredentials into `ShopCredentials` and makes the Shop Credendials verifier accept it as a dependency.

Ill outline some of the quirks inline


Assets/Shopify/UIToolkit/Editor/ShopCredentialsVerifier.cs
```diff
@@ -34,12 +34,11 @@ public class ShopCredentialsView : IShopCredentialsView {
...
             DrawActionButton();
             EditorGUILayout.Separator();
-            serializedObject.ApplyModifiedProperties();
         }
 
         private void DrawCredentialsForm(SerializedObject serializedObject) {
-            EditorGUILayout.PropertyField(serializedObject.FindProperty("_shopDomain"));
-            EditorGUILayout.PropertyField(serializedObject.FindProperty("_accessToken"));
+            EditorGUILayout.PropertyField(serializedObject.FindProperty("Credentials._Domain"));
+            EditorGUILayout.PropertyField(serializedObject.FindProperty("Credentials._AccessToken"));
```
JakeCataford: You can Keypath into other objects which is cool????????? Have to use the field name instead of the property here though because object serialization.

Assets/Shopify/UIToolkit/Internal/ShopCredentials.cs
```diff
@@ -0,0 +1,51 @@
+﻿using UnityEngine;
+
+namespace Shopify.UIToolkit {
+    [System.Serializable]
+    public class ShopCredentials {
+        public delegate void UpdateHandler(ShopCredentials self);
+        public event UpdateHandler OnUpdate;
```
JakeCataford: This is an observer pattern so that we can observe the changes to this class and invalidate the client... i dont love it :/

Assets/Shopify/UIToolkit/Test/Unit/Editor/TestShopControllerBase.cs
```diff
@@ -46,34 +46,10 @@ public class MockShopController : ShopControllerBase {
...
-            var client = _shopController.Client;
-
-            _shopController.ShopDomain = "other.myshopify.com";
-
-            var clientOnNextCall = _shopController.Client;
-
-            Assert.AreNotSame(client, clientOnNextCall);
-            Assert.AreEqual("other.myshopify.com", clientOnNextCall.Domain);
-        }
-
```
JakeCataford: this doesn't happen anymore.... because we just invalidate the client if anything changes on the serialized object.

Assets/Shopify/UIToolkit/Editor/ShopCredentialsVerifier.cs
sleroux: Cool! I had no idea. So you're able to pull in a serialized private data member out of an object?

Assets/Shopify/UIToolkit/Editor/ShopCredentialsVerifier.cs
```diff
@@ -34,12 +34,11 @@ public class ShopCredentialsView : IShopCredentialsView {
             DrawMessageBox();
             DrawActionButton();
             EditorGUILayout.Separator();
-            serializedObject.ApplyModifiedProperties();
```
sleroux: Is this intentional?

JakeCataford: yea because it should be called by the "user" of this class instead of here.... this thing just makes changes

Assets/Shopify/UIToolkit/Editor/ShopCredentialsVerifier.cs
JakeCataford: Yea i think it has to do with the fact that everything is just reduced to key value pairs so the "private/public" accessors stop meaning anything once you are in serializedland

JakeCataford:
@sleroux so I've removed the observer pattern and extracted the view logic out of the verifier... now the verifiers ONLY responsibility is to know if the credentials are correct, and the base shop controllers editor is responsible for drawing all the things.

Assets/Shopify/UIToolkit/ShopControllers/Editor/ShopControllerBaseEditor.cs
```diff
@@ -87,5 +79,53 @@ public abstract class ShopControllerBaseEditor : Editor, IShopControllerBaseEdit
             EditorGUILayout.PropertyField(serializedObject.FindProperty("_appleMerchantID"));
             #endif
         }
+
+        void IShopControllerBaseEditorView.DrawShopCredentialsVerifier() {
```
sleroux: Much better as a method than a class 👍

Assets/Shopify/UIToolkit/ShopControllers/Editor/ShopControllerBaseEditor.cs
```diff
@@ -87,5 +79,53 @@ public abstract class ShopControllerBaseEditor : Editor, IShopControllerBaseEdit
...
+            EditorGUILayout.Separator();
+
+            var disableCredentialsForm = _verifier.HasVerifiedCredentials() || _verifier.RequestInProgress;
+
+            EditorGUI.BeginDisabledGroup(disableCredentialsForm);
+
+            EditorGUI.BeginChangeCheck();
```
sleroux: This is interesting. You mentioned in slack about how the editor stuff is Immediate mode rendering and this API makes it pretty clear that it is. Good to now for future editor scripts!

JakeCataford:
@sleroux fixed it... that was way harder than expected

Assets/Shopify/UIToolkit/ShopControllers/BaseShopController.cs
```diff
@@ -94,22 +71,18 @@ public abstract class ShopControllerBase : MonoBehaviour, IShopCredentials {
...
-            _cachedClient = new ShopifyClient(LoaderProvider.GetLoader(AccessToken, ShopDomain));
+            _cachedClient = new ShopifyClient(LoaderProvider.GetLoader(Credentials.AccessToken, Credentials.Domain));
             _cachedCart = new CartController(_cachedClient.Cart(), _appleMerchantID);
             
             _cachedCart.OnPurchaseStarted.AddListener(Shop.OnPurchaseStarted);
-            _cachedCart.OnPurhcaseCancelled.AddListener(Shop.OnPurchaseCancelled);
+            _cachedCart.OnPurchaseCancelled.AddListener(Shop.OnPurchaseCancelled);
             _cachedCart.OnPurchaseComplete.AddListener(Shop.OnPurchaseCompleted);
-            _cachedCart.OnPurhchaseFailed.AddListener(Shop.OnPurchaseFailed);
-            _cachedCart.OnCartItemsChange.AddListener(Shop.OnCartItemsChanged);
```
sleroux: I think you might have accidentally removed the `OnCartItemsChange` listener

PR #491, opened 2018-01-30
Updated Cart Panel to enable Apple Pay checkouts

sleroux:
Adds `CanCheckoutWithNativePay` call to the Shop so the CartView can query when to show the Apple Pay button. As I was testing this patch I noticed the minimum spacing flag on the `NativePayButtonUI` script was causing problems and had an bug in it's method signature so I've decided to remove it since it's kind of pointless anyways. It wasn't clear why it existed and make it so the size of the button the developer sees in Unity matches the one rendered on iOS.

### DIFF
--- a/Assets/Shopify/Plugins/iOS/Shopify/ApplePayButtonGenerator.mm
+++ b/Assets/Shopify/Plugins/iOS/Shopify/ApplePayButtonGenerator.mm
@@ -54,21 +54,11 @@ extern "C" {
     
     // Generates a base64 encoded string containing the rendered image output of a PKPaymentButton from iOS.
     const char* _GenerateApplePayButtonImage(const char *type, const char* style, float width, 
-                                             float height, bool includeMinMargin,
-                                             bool includeMinimumSpacingForIOS) {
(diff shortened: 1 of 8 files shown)


Assets/Shopify/Plugins/iOS/Shopify/ApplePayButtonGenerator.mm
```diff
@@ -54,29 +54,18 @@ PKPaymentButtonType SBPayButtonTypeFromString(const char *type) {
     
     // Generates a base64 encoded string containing the rendered image output of a PKPaymentButton from iOS.
     const char* _GenerateApplePayButtonImage(const char *type, const char* style, float width, 
-                                             float height, bool includeMinMargin,
-                                             bool includeMinimumSpacingForIOS) {
```
sleroux: The actual method doubled up on the `includeMinMargin` and `includeMinimumSpacingForiOS` parameters but the compiler never caught this! This resulted in undefined values being passed in which was causing a lot of headaches debugging this.

PR #492, opened 2018-01-30
Fixes bug where checkout wasn't happening on single product shop renamed button to Buy

sleroux:
Small logic bug fix and renames the button on the single product shop page to just 'Buy'


Assets/Shopify/UIToolkit/ShopControllers/Classes/CartController.cs
```diff
@@ -159,14 +159,7 @@ private struct ProductAndVariantPair {
         /// </summary>
         /// <param name="mode">How do you want to make the purchase? Native, Web or Auto</param>
         public void StartPurchase(CheckoutMode mode) {
-
-            // Validate that the developer has given us the proper native payment key for the platform.
             var nativePayKey = NativePayKeyForCurrentPlatform();
-            if ((mode == CheckoutMode.Native || mode == CheckoutMode.Auto) && nativePayKey == null) {
```
sleroux: This wasn't working as intended when using `Auto` checkout mode.

PR #495, opened 2018-02-01
[UI Toolkit] Loading overlay for single product prefab

JakeCataford:
loading
Doesnt include testing scene changes, you will have to revert the prefab
after opening. Also it renames "onShow" and "onHide" to "load" and
"unload" for the shop controllers.


sleroux:
Since I have the fix on my local branch I can just push up that commit 👍

PR #496, opened 2018-02-01
Part 2 - Replaces ContentRegion script with infinite scrolling ScrollRect

sleroux:
Instead of using the ContentRegion script which captures the mouse down/up events manually, I've used a regular ScrollRect component to give us that bouncy scrolling feel. The content area still only ever uses a max of 5 elements while scrolling but always expands the size of the content to fit.

Edit: Added preview below 

list-stuff


sleroux:
Currently the images go a bit wonky when scrolling because of the async nature of the image loading. I'm going to fix this in a follow up PR. Another nice to have would be to add a 'loading' item view that indicates that there is more in the list. It'd be cool if we could use that spinning icon you made for the single product page as the loading cell. This is a nice to have as I gave it a crack during this PR and found it to be a pain to implement.

sleroux:
Nevermind I've added a fix for the image thing as part of this branch as well as a fix for the wrong product being opened. The image loading still isn't perfect and I'm not a fan of using boolean guard.

Assets/Shopify/UIToolkit/Components/RemoteImageLoader.cs
```diff
@@ -16,12 +16,16 @@ public class RemoteImageLoader : MonoBehaviour {
...
 
+        public bool LoadingInProgress;
+
         private Image _image;
 
         private delegate void RemoteImageLoadedDelegate(Texture2D texture, string error);
 
         private WebImageCache _imageCache = WebImageCache.SharedCache;
 
+        private bool cancelNextLoad;
```
JakeCataford: not sure what this is for?

Assets/Shopify/UIToolkit/Shops/Generic/Classes/MultiProductListItem.cs
```diff
@@ -27,13 +27,15 @@ public class MultiProductListItem : MonoBehaviour {
 
             var images = (List<Shopify.Unity.Image>)product.images();
             if (images.Count > 0) {
+                if (ImageLoader.LoadingInProgress) {
+                    ImageLoader.CancelPreviousLoad();
```
JakeCataford: oh i see... so if no new images are returned then we need to stop some loading loop?

PR #501, opened 2018-02-13
Fixes broken logic for showing variant title on cart item view

sleroux:


JakeCataford:
Should it use `String.IsNullOrEmpty`?

sleroux:
Didn't know that was a thing! I'll update it

JakeCataford:
did the tests pass on that? I think it might be lowercase `string` lol

JakeCataford:
nvm im wrong its upcase

PR #502, opened 2018-02-28
[Hotfix] Destroy message receiver after the checkout was cancelled so we don't get any callbacks

sleroux:
Adds the same destroy logic we had after a success callback for when cancel gets fired so we never get cancelled fired twice.

Note: After this lands into master I'll do the hotfix release from the v0.4.1 tag


sleroux:
Was a bit of a pain to get working because there were a lot of small changes that we made since v0.4.1 that were required for it to run on device and stuff. For the v0.4.2 hot fix release I'll need to cherry-pick those commits as well.

PR #511, opened 2018-04-24
Disable initialization of unsupported Android Pay checkout

sleroux:


sleroux:
@flavioarfaria Could you rubberstamp this 2 line change? Looks like there were issues  with NoClassDef found errors when initializing the native checkout pay stuff for Android. Since it's not in use at the moment I just commented it out for now.

sleroux:
Ah that makes sense. I'll keep a note of that when we add Android Pay support back in. Thanks Flavio!

Issue #512, opened 2018-05-17
image("medium").transformedSrc() always returns 1000x1000 image


Mitnainartinarian:
I'm trying to get a scaled down image of a product variant, but I'm always getting the same 1000x1000 image even if I specify something smaller. I had it working a while ago in an older version of the SDK (before image.src was deprecated). Is this the correct way to specify the size?

`variant.image("medium").transformedSrc()`

I use this to retrieve the products:

`ShopifyBuy.Client().products((fetchedProducts, err) => {...}, fetchIDs);`

I get the variant from the list returned by product.variants(). The code for transformedSrc is based on the comment for Shopify.Unity.ShopifyClient (on line 139 in ShopifyBuy.cs in v0.5.1 of the Unity SDK). I tried moving the "medium" alias to transformedSrc instead, but it didn't work either.

mikkoh:
@sleroux can we close this?

sleroux:
I've added this fix to the latest v0.5.2 release:

@Mitnainartinarian, to get the image for the given size, the API has slightly changed from:

`variant.image("medium").transformedSrc()`

to

`variant.image().transformedSrc("medium")`

PR #515, opened 2018-06-18
Use image aliases for transformedSrc properties instead of image queries

sleroux:
For loading different image resolutions, we were sending aliased `image` fields to the Storefront API and relying on the Image's `src` property for the sized URL. Using `src` directly has now been deprecated and `transformedSrc` should be used instead.

`transformedSrc` now takes in an alias to be used so instead of having a query that looks like this:

```graphql
    image___pico: image(maxWidth: 16, maxHeight: 16) {
        altText
        transformedSrc
    }
    image___icon: image(maxWidth: 32, maxHeight: 32) {
         altText
         transformedSrc
    }
```

Querying for different image sizes looks like this:

```graphql
    image {
        altText
        transformedSrc
        transformedSrc___pico: transformedSrc(maxWidth: 16, maxHeight: 16)
        transformedSrc___icon: transformedSrc(maxWidth: 32, maxHeight: 32)
    }
```

This means the client side API has changed from `product.image("pico").src` to `product.image().transformedSrc("pico")`

### DIFF
--- a/Assets/IntegrationTests/Utils.cs
+++ b/Assets/IntegrationTests/Utils.cs
@@ -5,7 +5,7 @@ using UnityEngine;
 namespace Shopify.Unity.Tests {
     public class Utils {
         public const float MaxQueryDuration = 3f;
-        public const string MaxQueryMessage = "Query failed to run in 1 seconds";
+        public const string MaxQueryMessage = "Query failed to run in 3 seconds";
 
         public static StoppableWaitForTime GetWaitQuery() {
             return new StoppableWaitForTime (MaxQueryDuration);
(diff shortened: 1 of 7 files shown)


sleroux:
Should resolve https://github.com/Shopify/unity-buy-sdk/issues/512

Assets/IntegrationTests/Utils.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Shopify.Unity.Tests {
     public class Utils {
-        public const float MaxQueryDuration = 1f;
+        public const float MaxQueryDuration = 3f;
         public const string MaxQueryMessage = "Query failed to run in 1 seconds";
```
mikkoh: Nit just need to change this to be 3s

Issue #537, opened 2018-10-12
Add UNION type GraphQL generation support


sleroux:
The storefront API recently added a new UNION type (https://help.shopify.com/en/api/custom-storefronts/storefront-api/reference/union/pricingvalue) but our GraphQL generation scripts don't support UNIONs. This results in an error when trying to generate the GraphQL C# classes.

PR #538, opened 2018-10-12
Added GraphQL UNION type generation support

sleroux:
Fixes https://github.com/Shopify/unity-buy-sdk/issues/537

The Storefront API introduced `PricingValue` which is the first UNION GraphQL type to be used in the API. We didnt have a way of handling UNION types so this PR adds that functionality to the generator. The approach here is to do a similiar thing we do for interfaces since a union is just a GraphQL interface that has no shared members. The main code difference is that since unions dont have any fields, the field generation code is omitted.

### DIFF
--- a/scripts/generator/graphql_generator/csharp.rb
+++ b/scripts/generator/graphql_generator/csharp.rb
@@ -106,7 +106,7 @@ module GraphQLGenerator
       # output type definitions
       schema.types.reject{ |type| type.builtin? || type.scalar? }.each do |type|
         # output
-        if type.object? || type.interface?
+        if type.object? || type.interface? || type.union?
           File.write("#{path_graphql}/#{type.name}Query.cs", reformat(TYPE_ERB.result(binding)))
           File.write("#{path}/#{type.name}.cs", reformat(TYPE_RESPONSE_ERB.result(binding)))
         elsif type.input_object? || type.kind == 'ENUM'
@@ -134,7 +134,7 @@ module GraphQLGenerator
         "List<#{graph_type_to_csharp_type(type.of_type)}>"
       when 'ENUM'
         is_non_null ? type.name : "#{type.name}?"
-      when 'INPUT_OBJECT', 'OBJECT', 'INTERFACE'
+      when 'INPUT_OBJECT', 'OBJECT', 'INTERFACE', 'UNION'
         type.classify_name
       else
         raise NotImplementedError, "Unhandled #{type.kind} input type"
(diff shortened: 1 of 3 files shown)


sleroux:
Note: There is still a compile error with the CurrencyCode? -> CurrencyCode conversion. I'll address that in the next PR.

scripts/generator/graphql_generator/csharp.rb
```diff
@@ -216,6 +216,12 @@ def response_init_interface(field)
       "Unknown#{type.classify_name}.Create((Dictionary<string,object>) dataJSON[key])"
     end
 
+    def response_init_union(field)
```
sleroux: Same implementation as `response_init_interface` but I duplicated it here to make it more clear.

scripts/generator/graphql_generator/csharp/type.cs.erb
```diff
@@ -44,53 +44,57 @@ namespace Shopify.Unity.GraphQL {
...
-                    <%# now we we want to handle generating arguments %>
-                    <% if field.args.any? %>
-                        if (alias != null) {
-                            ValidationUtils.ValidateAlias(alias);
 
-                            Query.Append("<%= field.name + ALIAS_SEPARATOR %>");
-                            Query.Append(alias);
-                            Query.Append(":");
-                        }
+            <% unless type.union? %>
```
sleroux: Unions don't have fields so omit generating any of these.

scripts/generator/graphql_generator/csharp/type_response.cs.erb
```diff
@@ -54,6 +82,8 @@ namespace Shopify.Unity {
                 }
             }
     <% end %>
+
+    <% unless type.union? %>
```
sleroux: Same comment as above - skip these functions since we don't have any fields.

scripts/generator/graphql_generator/csharp/type_response.cs.erb
```diff
@@ -25,6 +25,34 @@ namespace Shopify.Unity {
                 return connection.Nodes;
             }
         <% end %>
+    <% elsif type.union? %>
+        <%= docs_interface(type) %>
```
mikkoh: Should there be a `docs_union` function?

mikkoh:
Last I checked we weren't using Unions in the Storefront API. Would be nice on the C# side to have tests to ensure all this works but yeah the implementation is probably very similar to interfaces which seems like what you've done here.

sleroux:
They recently added a new PricingValue type which is a UNION which is why this is breaking now:

https://help.shopify.com/en/api/custom-storefronts/storefront-api/reference/union

Issue #527, opened 2018-08-10
Update Unity tests with `transformedSrc` changes to Image 


sleroux:
Whenever we moved from aliasing on Image to aliasing on `transformedSrc`, the tests were not updated. Some of the tests are now failing because of it.

PR #541, opened 2018-10-15
Fix issues with Play Mode tests

sleroux:
Related issue: https://github.com/Shopify/unity-buy-sdk/issues/527

Some of play mode tests were failing because of the following:

* Incorrect usage of `image(alias)`: When we changed to use `transformedSrc` the test wasn't updated to use the `image().transformedSrc(alias)` API.
* Timeouts on collection/product queries: The tests were a bit flakey and were running over the time limit by 0.5s so I bumped the timeout a bit to make these more reliable.
* GlobalGameObject already exists: Any tests that make use of the UnityLoader will touch the GlobalGameObject causing it to exist for a test case that checks it's existence. Annoying side effect of having a global. I just added a setup phase to clear out the GlobalGameObject in case it exists.
* Broken WebImageCache test: I've removed this rest all together since it doesn't test anything that unit test doesn't already test and relies on broken image links.
* Error Prefab not found: Simple path fix.

While I was investigating the timeouts on the queries, I noticed some minor issues with the query string we were building when querying for `transformedSrc`. We were sending multiple `image` queries each with a single `transformedSrc` for an alias instead of sending one `image` query with multiple `transformedSrc`. For example,

Before

```
image {
   altText
   transformedSrc
}
image {
   altText
   transformedSrc___pico: transformedSrc(maxWidth: 16, maxHeight: 16)
} 
image {
   altText
   transformedSrc___icon: transformedSrc(maxWidth: 32, maxHeight: 32)
}
```

After

```
 image {
   altText
   transformedSrc
   transformedSrc___pico: transformedSrc(maxWidth: 16, maxHeight: 16)
   transformedSrc___icon: transformedSrc(maxWidth: 32, maxHeight: 32)
   transformedSrc___thumb: transformedSrc(maxWidth: 50, maxHeight: 50)
}
```

I've updated this in the second commit.


Assets/IntegrationTests/TestGlobalGameObject.cs
```diff
@@ -13,6 +13,13 @@ namespace Shopify.Unity.Tests
     using System.Collections.Generic;
 
     public class TestGlobalGameObject : MonoBehaviour {
+        [SetUp]
+        public void SetUp() {
+            // GlobalGameObject might have been created from previous test runs.
+            var globalGameObject = GameObject.Find(GlobalGameObject.Name);
```
richardmonette: Duplicate of `TearDown`? Perhaps introduce a private method with the implementation?

PR #543, opened 2018-10-16
Migrate Unity/C# CI to Travis CI

sleroux:
Fixes part of https://github.com/Shopify/unity-buy-sdk/issues/508

This moves over the Unity/C# tests to run in Travis CI instead of Circle CI. On average, an integration takes 6 minutes with nothing cached a 4m30s with a cached copy of the Unity package. I've also been experimenting with including the iOS tests in CI and found that it makes the integrations take ~25mins with no cache/18 mins with cache. I'm not sure how I feel about having every integration take that long though so I've excluded it for now.

I've also configured Travis to only run one job at a time to prevent multiple activations from happening.


sleroux:
I've ended up just moving the iOS tests back to Travis for now and pushing it to this PR.

scripts/test_iOS.sh
```diff
@@ -31,7 +31,7 @@ xcodebuild test \
     -project "$IOS_PROJECT_PATH" \
     -sdk iphonesimulator \
     -scheme Unity-iPhone \
-    -destination 'platform=iOS Simulator,OS=11.0.1,name=iPhone 6'
+    -destination 'platform=iOS Simulator,OS=12.0,name=iPhone 8'
```
mikkoh: How do we remember to update this when a new OS comes out?

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+env:
+  UNITY_DOWNLOAD_DIR: $HOME/unity
+  UNITY_PKG_LOCATION: $UNITY_DOWNLOAD_DIR/Unity.pkg
+  UNITY_PKG_URL: https://download.unity3d.com/download_unity/5d30cf096e79/MacEditorInstaller/Unity-2017.1.1f1.pkg
+  IOS_PKG_LOCATION: $UNITY_DOWNLOAD_DIR/Unity-iOS.pkg
+  IOS_PKG_URL: http://netstorage.unity3d.com/unity/5d30cf096e79/MacEditorTargetInstaller/UnitySetup-iOS-Support-for-Editor.pkg
+
+notifications:
+  slack:
+    secure: tcssuzUngHLy4xd8ue9LCaF8prIJTCqLWmDDhuIOcRDIyavBe7Td5hDgTK2Dsfzuv5h1y8fCHm/J9qH3EytM94t2aZVIWY03QHv6MCRRk8Wm4TePPtBQaReYWC5rnzFN3tFtGLzxgja99FLbsebNQmb79yUdlelfLfLcDpVos4LrfVadhHaGfOoyS8Jz3DEH7d5AeatmZIvvbiFqJOeSbhYZmiRXRmPNSWJEPthsZQg2eat8mmtJnxyCLB5B+kEZinSOgz8/ufnkAJoUAbH/bGIZ8WYpNIMhLDg96JdZ8bi2MYq6T6eSPTN44Fg2FQcC7gszPOCdS1PdeeGH4VtvcTj/RvhfUDBH9SWbh7yEpz6h83nzX4+B8Snri33zL43S8JGwThTBXJyY0Lq9NtMdKf3MRKrO+0ehJQSgL1lfiH51ILWpXuqPczeB3IK+8lsOTxbP/q9nFFi3QfBdnaZc2LGRGAkL7x1VJIezDaI/C4aIryOwoJjmBvIx+kNrBdiL6zLja/1BHOHGE/XSPnT/urusT/1QFV4xQXMscxqptzb5YxFlF+qVnIPTY2ESO7hCajQWIZxhAQukLwbFMyt1Znv97CrWoQyfeWU8pMBfaeOEFYRq0rjnsMl1VDt9uALmqUa2NFC4t9WDo1W5Ptih9pZlIgSXmfP5yN0aYexzu5M=
```
mikkoh: Nice.

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+
+install: ./scripts/install_unity.sh
```
mikkoh: Does Unity get cached or does it do an install each time?

scripts/test_iOS.sh
sleroux: We can make a check list somewhere whenever a new Xcode is released in the fall of each year.

.travis.yml
sleroux: Before I added in the iOS stuff I had it cached but with iOS included I started to run into write permission errors when writing to the cache. I'm not sure why but I had a feeling it might be because our cache was 3-4GBs in size. We don't get _that_ much of a savings from it being cached though since downloading the cache takes a while anyways.

Issue #530, opened 2018-10-01
Update to the latest version of PlayServicesResolver


sleroux:
We're currently at 1.2.64 and the latest release is 1.2.91. There were some issues with package resolution in our current version so updating might help.

Issue #516, opened 2018-06-19
Building for Android using the Gradle build system fails


sleroux:
When using Unity 2018.1.0f2 and importing the Unity Buy SDK by itself into a new project, Android builds are failing when using the Gradle build system with the following error:

```
CommandInvokationFailure: Gradle build failed. 
/Library/Java/JavaVirtualMachines/jdk1.8.0_131.jdk/Contents/Home/bin/java -classpath "/Applications/Unity/Hub/Editor/2018.1.0f2/PlaybackEngines/AndroidPlayer/Tools/gradle/lib/gradle-launcher-4.0.1.jar" org.gradle.launcher.GradleMain "-Dorg.gradle.jvmargs=-Xmx2048m" "assembleRelease"

stderr[
/Users/sleroux/Desktop/TestAndroid/Temp/gradleOut/build/intermediates/res/merged/release/values/values.xml:26: AAPT: Error retrieving parent for item: No resource found that matches the given name 'Theme.AppCompat.NoActionBar'.
    
/Users/sleroux/Desktop/TestAndroid/Temp/gradleOut/build/intermediates/res/merged/release/values/values.xml:28:21-32: AAPT: No resource found that matches the given name: attr 'colorAccent'.
    
/Users/sleroux/Desktop/TestAndroid/Temp/gradleOut/build/intermediates/res/merged/release/values/values.xml:29:21-42: AAPT: No resource found that matches the given name: attr 'colorControlHighlight'.
    
/Users/sleroux/Desktop/TestAndroid/Temp/gradleOut/build/intermediates/res/merged/release/values/values.xml:27:21-39: AAPT: No resource found that matches the given name: attr 'colorControlNormal'.
    
/Users/sleroux/Desktop/TestAndroid/Temp/gradleOut/build/intermediates/res/merged/release/values/values.xml:30:21-33: AAPT: No resource found that matches the given name: attr 'colorPrimary'.
    
/Users/sleroux/Desktop/TestAndroid/Temp/gradleOut/build/intermediates/res/merged/release/values/values.xml:32: AAPT: Error retrieving parent for item: No resource found that matches the given name 'Theme.Design.Light.BottomSheetDialog'.
    
/Users/sleroux/Desktop/TestAndroid/Temp/gradleOut/build/intermediates/res/merged/release/values/values.xml:34:21-32: AAPT: No resource found that matches the given name: attr 'colorAccent'.
    
[...]

PR #546, opened 2018-10-22
Updated PlayServicesResolver to 1.2.92

sleroux:
Resolves https://github.com/Shopify/unity-buy-sdk/issues/530 and https://github.com/Shopify/unity-buy-sdk/issues/516

The issue was mostly to keep our dependencies up to date but it turns out that by updating to the latest PlayServicesResolver we fix the Gradle build issue on Unity 2018 as well.

The Gradle build issue was being caused by a bug in 2018 which doesn't replace the `${applicationId}` field in the generated AndroidManifest.xml file. PlayServicesResolver has a workaround built in that updates the AndroidManifest.xml to be correct for 2018. Two birds on stone 🙂


richardmonette:
In https://github.com/Shopify/unity-buy-sdk/pull/545 we are removing AndroidPay, is there something else we need the PlayServicesResolver for?

sleroux:
The Web Pay module also requires sub dependencies which get pulled in using the PlayServicesResolver.

PR #564, opened 2019-01-15
Generate v1.0.0 unitypackage

rebeccajfriedman:
Generates the v1.0.0 unitypackage file.

Notes
- It looks like the Asset changes are actually just undoing asset changes that were merged here: https://github.com/Shopify/unity-buy-sdk/pull/560.
- scripts/version is already updated to v1.0.0 (PR: https://github.com/Shopify/unity-buy-sdk/pull/563)


mikkoh:
🚀

PR #565, opened 2019-04-17
Migrate iOS plugin to Swift 4

sleroux:
Migrates the iOS plugin to support Swift 4 and resolves any errors that were caused from the switch. The two key errors that were happening were:

1. Illegal overriding of a Swift extension method in Deserializable. Fixed this by added in `@objc` prefix to make it into an ObjC extension instead.
2. Failable initializer not captured in `PaymentAuthorizationController` protocol. The `PaymentAuthorizationViewController` framework class has a failable initializer and since it implements this protocol the protocol method definition should match.

### DIFF
--- a/Assets/Shopify/Plugins/iOS/Shopify/PaymentAuthorizationControlling/PaymentAuthorizationControlling.swift
+++ b/Assets/Shopify/Plugins/iOS/Shopify/PaymentAuthorizationControlling/PaymentAuthorizationControlling.swift
@@ -31,7 +31,7 @@ import PassKit
     
     var authorizationDelegate: NSObjectProtocol? { get set }
     
-    init(paymentRequest request: PKPaymentRequest)
+    init?(paymentRequest request: PKPaymentRequest)
     func present(completion: ((Bool) -> Void)?)
     func dismiss(completion: (() -> Void)?)
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 os: osx
-osx_image: xcode10
+osx_image: xcode10.2
 
 env:
   UNITY_DOWNLOAD_DIR: $HOME/unity
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 os: osx
-osx_image: xcode10.2
+osx_image: xcode10
 
 env:
   UNITY_DOWNLOAD_DIR: $HOME/unity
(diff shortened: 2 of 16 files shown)


sleroux:
@rebeccajfriedman I did some manual testing and ran the test suite and everything seems to be fine. I'm pretty confident we can just land this because I'm not sure if there is anyone around to review it 🤷‍♂️

rebeccajfriedman:
LGTM! @dbart01 Any concerns?

dbart01:
Any particular reason why Swift 4 and not 5?

Assets/Shopify/Plugins/iOS/Shopify/PaymentAuthorizationControlling/PaymentAuthorizationControlling.swift
```diff
@@ -31,7 +31,7 @@ import PassKit
     
     var authorizationDelegate: NSObjectProtocol? { get set }
     
-    init(paymentRequest request: PKPaymentRequest)
+    init?(paymentRequest request: PKPaymentRequest)
```
dbart01: Why is this necessary? Making this a failable initializer effectively cascades that dependency down to `PaymentSession` where it is force-unwrapped.

sleroux:
> Any particular reason why Swift 4 and not 5? 

I was concerned initially about moving straight to 5 before I did these fixes but now that I resolved the issues with 4, 5 seems to just work as well :+1: I'll update it to target that instead.

Assets/Shopify/Plugins/iOS/Shopify/PaymentAuthorizationControlling/PaymentAuthorizationControlling.swift
sleroux: Looks like this protocol is used both on `PKPaymentAuthorizationController` and `PKPaymentAuthorizationViewController`. The first has a regular initializer while the view controller class has a failable one and Swift 4 throws an error saying that the protocol doesn't support overriding a failable initialier with a normal one. I guess making this failable is more true of what could happen if `PaymentSession` is nil. In a previous commit I did the force unwrap (bad) but I've since updated it to properly return a `false` indicating that that the user has no supported payment options.

dbart01: Sounds reasonable 👍

sleroux:
After upgrading to Swift 5 and Xcode 10.2 there were some other errors that sprung up. For now we'll go with Swift 4 to unblock compiling this with the new Xcode and follow up with the Swift 5 fix in the future.

PR #581, opened 2019-10-04
Update to a newer storefront API version

osamie:
**Summary of Changes**
- Point to a newer version of the storefront API graphql endpoint. 

Both unit and integration tests pass locally. 
Tested using sample app locally.


jmignac:
Should we pull the schema changes on this PR too?

jmignac:
> Should we pull the schema changes on this PR too?

Discussed over chat: we don't need to do that unity sdk. We use introspection to determine the graphql schema during the graphql client generation phase.

osamie:
> Should we pull the schema changes on this PR too?

Chatted with Jesse. We don't need to do that unity sdk. We use introspection to determine the graphql schema during the graphql client generation phase

PR #583, opened 2019-10-07
Fix printf error in test_iOS script

osamie:
**Errors when running test_iOS.sh and test_unity.sh** 
```
./scripts/publish.sh: line 11: printf: --: invalid option
printf: usage: printf [-v var] format [arguments]
cat: /Users/zuwaomigie/src/github.com/Shopify/unity-buy-sdk/export.log: No such file or directory
```

**Fix**
Add formatter in printf string

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -8,7 +8,7 @@ check() {
         printf "Export %s finished properly\n." "$@"
     else
         printf "Export %s failed. Exited with %s\n" "$@" "$?"
-        printf "------------------\n\n"
+        printf "%s------------------\n\n"
         cat "$UNITY_LOG_PATH"
 
         exit 1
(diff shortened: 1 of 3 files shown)


scripts/publish.sh
```diff
@@ -8,7 +8,7 @@ check() {
         printf "Export %s finished properly\n." "$@"
     else
         printf "Export %s failed. Exited with %s\n" "$@" "$?"
-        printf "------------------\n\n"
+        printf "%s------------------\n\n"
```
rebeccajfriedman: Shouldn't this be `printf '%s\n\n' "------------------"`?

osamie: @rebeccajfriedman True, fixed
> Shouldn't this be printf '%s\n\n' "------------------"?

PR #591, opened 2020-01-20
Remove openssl from dev.yml

jmignac:
Followed instructions on `dev up` failure.


DrTeePot:
How is HTTPS implemented? Does the web server just handle it? Why was this included to begin with?

PR #624, opened 2023-05-16
Checkout + Build Fixes

kent-sole:
-- Note: I see that the PR Guidelines request that I target the 'beta' branch, but there is no 'beta' branch.

-- Tester.cs
- Adds UNITY_IOS preprocessor tags around code that will cause non iOS builds to fail because of unavailable DllImport target

-- CartLineItem.cs
- Sets cart to an unsaved state when you add/remove items to force checkouts to always update the remote cart
- To reproduce this bug: (1) Add an item to a cart, (2) go to checkout, (3) close the checkout, (4) add a second item to the cart, (5) go to checkout again. The cart will only contain the first item because adding/removing items does not dirty the cart.

### DIFF
--- a/Assets/NativeTesting/Native/Tester.cs
+++ b/Assets/NativeTesting/Native/Tester.cs
@@ -6,7 +6,8 @@
     using Shopify.Unity.SDK.iOS;
 
     public class Tester : MonoBehaviour {
-
+        
+#if UNITY_IOS
         [DllImport ("__Internal")]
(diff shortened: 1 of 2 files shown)


kent-sole:
I have signed the CLA!

kent-sole:
In my initial commit I only added the line to dirty the cart in AddOrUpdate(), but I forgot to add it to Delete(). I force pushed the change for a clean commit.

